PR #32, opened 2017-02-27
initial attempt at adding typings, fixes #30

### LINKED ISSUE #30: Add typescript bindings

tycho01:


package.json
```diff
@@ -70,6 +70,8 @@
     "phantom": "=4.0.0",
     "rimraf": "=2.5.4",
     "testem": "=1.15.0",
+    "typescript": "^2.2.0",
```
char0n: Can we use `=` prefix to have absolute control over dev environment ?

char0n:
Before merging the pull request, can we resolve following issues ?

1.) some type of test coverage
2.) Documentation in README.md  (typescript support, installing and using the typings)
3.) is it possible to have https://github.com/types/npm-ramda-adjunct ?

tycho01:
I'm able to create a repo at @types, though if you want push access there we might need to ask @blakeembrey. That said, keeping the typings in here could potentially simplify things for users (less packages to install).

char0n:
Agree on point 3.) then, let's keep it here. What about point 1 and 2 ?

tycho01:
If typings are in this repo, they'd get used automatically by those who need it (`index.d.ts` listed in `package.json`). I agree tests should be added, but my todo-list was kinda big already.

char0n:
Thanks for your work, I'll merge it tonight. It will be part of next release.

char0n:
Great job, merging

tycho01:
On a minor note, you'll notice the positive versions are more granular than the negative functions, managing to distinguish `true` vs. `false`. Not sure yet how to get the latter on par so far.

char0n:
What do you mean by that ? Try to collaborate more pls.

tycho01:
Return type for `isNotString` would be `boolean`, while the return type for
`isString` will end up as `true` or `false` (see TS [guards](
https://www.typescriptlang.org/docs/handbook/advanced-types.html)).

char0n:
I see, type guards. Thanks for the link.

PR #132, opened 2017-09-13
Typescript `boolean` is not `Boolean`

aliem:
It's probably a typo in the definition, probably the inconsistency has been introduced recently in the typescript side.

In the current stable Typescript the javascript `boolean` type is defined with lower case letters, the current definition is apparently referencing a global interface: https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#322-the-boolean-type


char0n:
@aliem nice one, thank you for your contribution.

Issue #137, opened 2017-09-23
Change all Object types to object in index.d.ts


BjornMelgaard:
because `type Object only allow you to assign any value to them` ([from](https://www.typescriptlang.org/docs/handbook/basic-types.html))

https://blog.mariusschulz.com/2017/02/24/typescript-2-2-the-object-type#object-vs-object-vs

char0n:
Seems like we have more problems like that in the typings. Sometimes the types are capitalized, sometimes not.

`Number` vs `number`
`String` vs `string`

BjornMelgaard:
It better to change them all to primitives

BjornMelgaard:
I'll do this after mergePaths

char0n:
Nice job. We'll go for the release this evening.

PR #139, opened 2017-09-23
Refactor/better ts typings

BjornMelgaard:
fix https://github.com/char0n/ramda-adjunct/issues/137 + more

commits are self-explanatory

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -98,7 +98,7 @@ declare namespace RamdaAdjunct {
         /**
          * Checks if input value is `Generator Function`.
          */
-        isGeneratorFunction(val: any): boolean;
+        isGeneratorFunction(val: any): val is Function;
 
         /**
          * Checks if input value is complement of `Generator Function`.
@@ -108,7 +108,7 @@ declare namespace RamdaAdjunct {
         /**
          * Checks if input value is `Async Function`.
          */
-        isAsyncFunction(val: any): boolean;
+        isAsyncFunction(val: any): val is Function;
 
         /**
          * Checks if input value is complement of `Async Function`.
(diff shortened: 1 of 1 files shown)


src/index.d.ts
```diff
@@ -210,12 +212,12 @@ declare namespace RamdaAdjunct {
         /**
          * Checks if value is a positive `Number` primitive or object.
          */
-        isPositive(val: any): boolean
+        isPositive(val: any): val is number
```
char0n: Are we sure here ? `isPositive` returns false for the NaN, -Infinity, +Infinity. Nevertheless they are still valid `number` values. The same applies for `isNegative`, `isInteger`, `isFloat`

src/index.d.ts
```diff
@@ -468,8 +469,8 @@ declare namespace RamdaAdjunct {
         /**
          * Returns whether or not an object has an own property with the specified name at a given path.
          */
-        hasPath(path: Array<String|number>, obj: any): boolean;
-        hasPath(path: Array<String|number>): (obj: Object) => boolean;
+        hasPath(path: Array<String|number>, obj: object): boolean;
```
char0n: String -> string

src/index.d.ts
BjornMelgaard: I dont know how to narrow number type to omit named values

stackoverflow too
https://stackoverflow.com/questions/46389560/how-to-narrow-number-type-to-omit-nan-infinity-infinity

BjornMelgaard: even haskell cant do this I think :smile: (but idris can)

BjornMelgaard:
@char0n  done

Issue #99, opened 2017-07-08
Assimilate ramda-extension


char0n:
Repo: https://www.npmjs.com/package/ramda-extension

Seems like ramda-adjunct already implements most of what ramda-extension contains. We already have separate issues for some utils contained within ramda-extension. There are some e.g. reduce callable that are worth exploring.

char0n:
We discontinue this effort. ramda-extension seems to be still evolving.

PR #144, opened 2017-09-30
feature (trace): minimal working example

BjornMelgaard:


char0n:
@BjornMelgaard, please next time create a new issue describing a feature before creating a PR so that we can collaborate on it. 

I am sorry, but I am not going to merge this PR from the following reasons:

 - idea of `trace` has been already considered and rejected. It is part of #99 in it's [simple form](https://github.com/tommmyy/ramda-extension/tree/master/src)
 - it is too specific even for ramda-adjunct
 - it doesn't implement [Dependancy Inversion Principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) and it's not composable with any external logging mechanism
 - it directly contains [side effects](https://en.wikipedia.org/wiki/Side_effect_%28computer_science%29) which should not be part of any functional library
 - it can be easily be done using `tap` operator composition

```javascript
 const logging = R.curryN(2, console.dir);

 const performComplexCalc = R.compose(
   R.tap(logging('after')),
   R.subtract(__, 2),
   R.divide(__, 4),
   RA.trace('middle'),
   R.multiply(10),
   R.add(10),
   R.tap(logging('before'))
);
```

BjornMelgaard:
Fine, took it out [here](https://github.com/BjornMelgaard/ramda-universal-trace)

char0n:
@BjornMelgaard sorry ;[

BjornMelgaard:
@char0n  no worries)

tommmyy:
@BjornMelgaard If you don´t agree with @char0n , you can still use `trace`, from [ramda-extension](https://github.com/tommmyy/ramda-extension) . Our library is modular and we are focused on pretty codebase.

BjornMelgaard:
@tommmyy tnx, noted

guillaumearm:
There exist a [ramda-debug](https://github.com/sebinsua/ramda-debug) npm package.
I never used it, but it seems very useful.

PR #168, opened 2017-10-19
feat: add isPair, isNotPair

### LINKED ISSUE #90: isPair
@char0n open to a PR for `isPair` and `isNotPair`? Something along the lines of:
```javascript
const isPair = both(isArray, pipe(length, equals(2)));
const isNotPair = complement(isPair);
```

rockymadden:
Closes #90

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -265,6 +265,16 @@ declare namespace RamdaAdjunct {
         isEven(val: any): number;
 
         /**
+         * Checks if input value is a pair.
+         */
+        isPair(val: any): boolean;
+
+        /**
+         * Checks if input value is complement of a pair.
+         */
+        isNotPair(val: any): boolean;
+
+        /**
          * A function that returns `undefined`.
          */
         stubUndefined(): undefined;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -267,7 +267,7 @@ declare namespace RamdaAdjunct {
         /**
          * Checks if input value is a pair.
          */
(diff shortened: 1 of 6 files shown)


src/index.d.ts
```diff
@@ -265,6 +265,16 @@ declare namespace RamdaAdjunct {
         isEven(val: any): number;
 
         /**
+         * Checks if input value is a pair.
+         */
+        isPair(val: any): boolean;
```
char0n: Maybe we could go for a 
```javascript
isPair(val: any): val is Array
```

The same for isNotPair

test/isNotPair.js
```diff
@@ -0,0 +1,31 @@
+import RA from '../src/index';
+import eq from './shared/eq';
+
```
char0n: one more empty line pls

test/isPair.js
```diff
@@ -0,0 +1,31 @@
+import RA from '../src/index';
+import eq from './shared/eq';
+
```
char0n: one more empty line pls

test/isPair.js
```diff
@@ -0,0 +1,31 @@
...
+describe('isPair', function() {
```
char0n: Can we reword the `it` sections so that they are more consistent with ramda tests (all sections should start with test) ?

Example: https://github.com/char0n/ramda-adjunct/blob/40d74e6a2f1e11769e065ad012a4150989e458d8/test/isInteger.js#L8

char0n:
Good job!

Issue #166, opened 2017-10-17
spread


char0n:
Ramda cookbook reference: https://github.com/ramda/ramda/wiki/Cookbook#spread

char0n:
@BjornMelgaard final addition to our merging algorithms. Are you interested ? 

PS: I am currently working on isomorphic lenses [here](https://github.com/char0n/ramda-adjunct/pull/165). Will appreciate code review after it's fully implemented.

BjornMelgaard:
@char0n yeah, would like to help, I'll make it in 7 days

BjornMelgaard:
@char0n so...what about change spread to flattenProp and flattenPath?? (lol)

example

```
const flattenProp = R.converge(R.merge, [R.dissoc, R.propOr({})]);
// usage:
// flattenProp("b", { a: 1, b: { c: 3, d: 4 } }); // -> { a: 1, c: 3, d: 4 }


const flattenPath = ...
// usage:
// flattenPath(["b1", "b2"] { a: 1, b1: { b2: { c: 3, d: 4 } } }); // -> { a: 1, c: 3, d: 4 }
```

char0n:
@BjornMelgaard will think about it. Will have to research is `flatten` is a proper name of the mechanism and `spread` is only an implementation detail or vice versa.

BjornMelgaard:
@char0n https://github.com/acdlite/recompose/blob/master/docs/API.md#flattenprop

so flatten from recompose is not the same of spread here: flatten leaves original object, while spread - not
I would say that recompose behavior is more useful

char0n:
As I understand it, we have two distinctive behaviors. `spreadPath` and `flattenPath` (`spreadProp` and `flattenProp` can be derived later from `*Path` functions). They both do different things which seems equivalently practical. 

When we have both (and already having the set you implemented earlies), I guest we can completely minic ES6 spread operator in functional way.

BjornMelgaard:
then.... I'll made `spreadPath` and `flattenPath` (and others) on weekends, awesome

char0n:
Pls send 2 independent PR when doing this.

BjornMelgaard:
noted

char0n:
Task for me: reflect the tests from `spreadProp` into `spreadPath`

char0n:
I was thinking we can now compose `spreadProp` and `spreadPath` directly from `flattenProp` and `flattenPath` ? What do you think ?

char0n:
[...]

PR #183, opened 2017-10-23
feature(#166): add spreadProp & spreadPath

BjornMelgaard:


char0n:
Will check during today

src/spreadPath.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Merges object under path with root object, removing path to object.
+ *
+ * @func spreadPath
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.19.0|v1.19.0}
+ * @category Object
+ * @sig [k] -> {a: *} -> {b: *}
+ * @param {!Object} path The path to object to merge with
+ * @param {!Object} obj The root object
```
char0n: Let's borrow the lingo from [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_operator)

`The provided object`

src/spreadPath.js
```diff
@@ -0,0 +1,29 @@
+import { curry, pathOr, merge, dissoc } from 'ramda';
+
+/**
+ * Merges object under path with root object, removing path to object.
```
char0n: `Spreads object under property path onto provided object.`

IMHO we don't  need to tell that it removed the object from the path. Spread is by definition a new composition of the object.

src/spreadPath.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @func spreadPath
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.19.0|v1.19.0}
+ * @category Object
+ * @sig [k] -> {a: *} -> {b: *}
```
char0n: Isn't this signature closer to actually what happens ?
```javascript
 * @sig
 *   [Idx] -> {k: v} -> {k: v}
 *   Idx = String | Int
```

test/spreadPath.js
```diff
@@ -0,0 +1,45 @@
+import RA from '../src/index';
+import eq from './shared/eq';
+
+
+describe('spreadPath', function() {
```
char0n: We should finally utilize the power of mocha more. I'll demonstrate what I have in mind.
Also don't try to create variables when they are not needed.

```javascript
import RA from '../src/index';
import eq from './shared/eq';


describe('spreadPath', function() {
  let obj;

  beforeEach(function () {
    obj =  {
      a: 1,
      b1: { b2: { c: 3, d: 4 } },
    };
  });

  it('tests currying', function () {
    const path = ['b1', 'b2'];
    const expected = { a: 1, c: 3, d: 4  };

    eq(RA.spreadPath(path, obj), expected);
    eq(RA.spreadPath(path)(obj), expected);
  });

  context('when path leads  to non object', function () {
    specify('should return object without path', function () {
      const obj = {
        a: 1,
        b1: { b2: 999 },
      };
      eq(RA.spreadPath(['b1', 'b2'], obj), { a: 1, b1: {} });    
    });
  });
  
  context("when path doesn't exist", function () {
    specify('should return object with identical structure as provided object', function () {
      eq(RA.spreadPath(['does', 'not', 'exist'], obj), obj);
    });
  });  
});

File: test/spreadProp.js [hunk 5]
```diff
@@ -0,0 +1,34 @@
+import RA from '../src/index';
+import eq from './shared/eq';
+
+
+describe('spreadProp', function() {
```
char0n: Redo these tests according what I explained on `spreadPath` tests and I think we should have more tests here

 - [x] currying
 - [ ] non existing prop
 - [x] prop that is non object
 - [ ] test for overriding existing props on provided object
 - [ ] test for standard usecase

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
+import { of, curry } from 'ramda';
+
+import spreadPath from './spreadPath';
+
+/**
+ * Merges object under prop with root object, removing prop to object.
```
char0n: Spreads object under property onto provided object.

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @func spreadProp
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.19.0|v1.19.0}
+ * @category Object
+ * @sig @sig k -> {a: *} -> {b: *}
```
char0n:
```javascript
 * @sig
 *   Idx -> {k: v} -> {k: v}
 *   Idx = String | Int
```

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {!Object} prop The name of property that holds object to merge with
```
char0n: The property path to spread

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ * @param {!Object} obj The root object
```
char0n: The provided object

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ * @return {!Object} Result of merging root object with object under prop
```
char0n: The result of the spread

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ * @see {@link RA.spreadPath|spreadPath}
+ * @example
+ *
+ * R.spreadProp("b", { a: 1, b: { c: 3, d: 4 } }); // => { a: 1, c: 3, d: 4 }
```
char0n: use single quotes: 'b'

src/spreadProp.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+
```
char0n: drop this empty line

test/spreadProp.js
```diff
@@ -0,0 +1,34 @@
...
+  it('tests currying', function () {
+    const prop = 'b';
```
char0n: line 7 to 14 are not needed here IMHO. We are testing currying here only. The body of this test could be following
```javascript
it('tests curring', function () {
  eq(RA.spreadProp('prop', {}), {}); 
  eq(RA.spreadProp('prop')({}), {});
});

File: src/spreadPath.js [hunk 14]
```diff
@@ -0,0 +1,26 @@
...
+ * Spreads object under property path onto provided object.
+ *
+ * @func spreadPath
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.19.0|v1.19.0}
+ * @category Object
+ * @sig
+ *   [Idx] -> {k: v} -> {k: v}
+ *   Idx = String | Int
+ * @param {!Object} path The property path to spread
```
char0n: ```@param {!Array.<string|number>}``` Not sure about proper syntax for union types in jsdoc

src/spreadPath.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param {!Object} obj The provided object
+ * @return {!Object} The result of the spread
+ * @see {@link RA.spreadProp|spreadProp}
+ * @example
+ *
+ * R.spreadPath(
+ *   ['b1', 'b2'],
+ *   { a: 1, b1: { b2: { c: 3, d: 4 } } }
+ * ); // => { a: 1, c: 3, d: 4, b1 {} }
```
char0n: missing colon `{ a: 1, c: 3, d: 4, b1: {} }`

src/spreadProp.js
```diff
@@ -0,0 +1,25 @@
...
+ * Spreads object under property onto provided object.
+ *
+ * @func spreadProp
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.19.0|v1.19.0}
+ * @category Object
+ * @sig
+ *   Idx -> {k: v} -> {k: v}
+ *   Idx = String | Int
+ * @param {!Object} prop The property to spread
```
char0n: ```@param {string|number}```

BjornMelgaard:
done

PR #218, opened 2017-11-25
feature(197): add stubArray

BjornMelgaard:


src/stubArray.js
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * A function that returns new empty array on every call.
+ *
+ * @func stubArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.1.0|v2.1.0}
+ * @category Function
+ * @sig ... -> []
+ * @return {[]}
```
char0n: We should infere the type not the value
```javascript
@return {Array} New empty array
```

src/index.d.ts
```diff
@@ -284,6 +284,11 @@ declare namespace RamdaAdjunct {
          */
         stubNull():  null;
 
+        /**
+         * A function that returns new empty array on every call.
+         */
+        stubArray():  {};
```
char0n: Shouldn't we infere type here, not the value ?
```javascript
stubArray(): Array;
```

src/stubArray.js
```diff
@@ -0,0 +1,19 @@
...
+ * @example
+ *
+ * RA.stubArray(); //=> []
+ * RA.stubArray(1, 2, 3); //=> []
+ */
+const stubArray = always([]);
```
char0n: we cannot use always here because we will always be returning the same instance. We should always return new pristine instance.
```javascript
const stubArray = () => [];
```

test/stubArray.js
```diff
@@ -0,0 +1,11 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('stubArray', function() {
+  it('tests `function` that returns `[]`', function() {
```
char0n: whould rephrase for 
```
it('tests `function` that returns empty array',...
```

char0n: Also let's add this test
```javascript
context('when called', function() {
  specify('should always return new instance', function() {
 ....
  });
});

File: test/stubArray.js [hunk 5]
```diff
@@ -0,0 +1,11 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
```
char0n: extra new line pls

src/index.d.ts
BjornMelgaard: oops, must be `[]`, literal types

char0n: Will you change for literal type `[]` ? After that we can merge.

BjornMelgaard: Oops))))

char0n:
Tests for typescript signature are failing. I guess array literal type is not supported ?

BjornMelgaard:
@char0n typescript error, [should be fixed not](https://github.com/Microsoft/TypeScript/issues/16636), but whatever

char0n:
Strange, we're using typescript 2.6.1

PR #220, opened 2017-11-25
feature(198): add stubString

BjornMelgaard:


src/index.d.ts
```diff
@@ -284,6 +284,11 @@ declare namespace RamdaAdjunct {
          */
         stubNull():  null;
 
+        /**
+         * A function that returns new empty string.
```
char0n: A function that returns empty string.

src/index.d.ts
```diff
@@ -284,6 +284,11 @@ declare namespace RamdaAdjunct {
...
+         */
+        stubString(): '';
```
char0n: Again, infering type not the value ?

src/stubString.js
```diff
@@ -0,0 +1,19 @@
+import { always } from 'ramda';
+
+/**
+ * A function that returns new empty string.
```
char0n: A function that returns empty string.

src/stubString.js
```diff
@@ -0,0 +1,19 @@
...
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.1.0|v2.1.0}
+ * @category Function
+ * @sig ... -> ''
+ * @return {''}
+ * @example
+ *
+ * RA.stubString(); //=> ''
+ * RA.stubString(1, 2, 3); //=> ''
+ */
+const stubString = always('');
```
char0n: This is OK because in JS two empty strings are always equal

test/stubString.js
```diff
@@ -0,0 +1,11 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('stubString', function() {
+  it("tests `function` that returns `'''`", function() {
+    eq(RA.stubString(), '');
+    eq(RA.stubString([1]), '');
+    eq(RA.stubString(new Array()), '');
+    eq(RA.stubString(1, 2, 3), '');
+  });
```
char0n: context('when called', function() {
  specify('should always return empty string', function() {
 ....test here for equality of results of 2 consequent invocations
  });
});

src/index.d.ts
BjornMelgaard: Is this a problem, typescript handles this fine?

char0n: If it handles it, it's fine. Did not know about literal types in typescript.

src/stubString.js
```diff
@@ -0,0 +1,19 @@
+import { always } from 'ramda';
+
+/**
+ * A function that returns empty string.
+ *
+ * @func stubString
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.1.0|v2.1.0}
+ * @category Function
+ * @sig ... -> ''
```
char0n: Hindley-Milner signatures don't use literal types. They explusively use types. It's deal with that

BjornMelgaard: @char0n you merged it with wrong signature `@sig ... -> ''`?

char0n: yeah, fixed it here: https://github.com/char0n/ramda-adjunct/commit/555ae2b30c29ae3cdb98b4dfe6b2c33ab4d4617f

PR #221, opened 2017-11-25
refactor(isArray): use Array.isArray if available

### LINKED ISSUE #211: isArray
Use Array.isArray under the hood, otherwise fallback to ramda.

char0n:
Closes #211

### DIFF
--- a/src/isArray.js
+++ b/src/isArray.js
@@ -1,4 +1,6 @@
 import _isArray from 'ramda/src/internal/_isArray';
+import { or } from 'ramda';
+
 
 /**
  * Checks if input value is `Array`.
@@ -17,6 +19,6 @@ import _isArray from 'ramda/src/internal/_isArray';
  * RA.isArray(null); //=> false
  * RA.isArray({}); //=> false
  */
-const isArray = _isArray;
+const isArray = or(Array.isArray, _isArray);
 
 export default isArray;


char0n:
@BjornMelgaard CR pls

char0n:
We now have native way in ES6 to check if something is array with `Array.isArray`. This PR uses this ES6 feature if available, if not it falls-back to ramda array checking.

BjornMelgaard:
I wonder how it works with [`or`](https://github.com/ramda/ramda/blob/v0.25.0/source/or.js)? Isn't it must use [`either`](https://github.com/ramda/ramda/blob/v0.25.0/source/either.js)?

char0n:
Nope, `either` is higher order function, `or` is boolean switch. So the code can be rewritten to:
```javascript
const isArray = Array.isArray || _isArray;
```
or 
```javascript
const isArray = (val) => {
  if (isFunction(Array.isArray)) { 
    return Array.isArray(val);
  }
  return _isArray(val);
}
```

BjornMelgaard:
@char0n this what I mean too
Isn't `Array.isArray || _isArray;` just checks presence of `Array.isArray` or `_isArray`

char0n:
So is it clear now ? Shall we go for the merge ?

BjornMelgaard:
@char0n yes, its clear now

```js
const isArray = Array.isArray || _isArray;
```

is more like

```js
let isArray
if(Array.isArray) {
  isArray = Array.isArray
else {
  isArray = _isArray
}
```

sorry, can merge now

PR #223, opened 2017-11-26
Annotating function calls as pure for better dead code elimination

Andarist:
Automatically inserted `#__PURE__` comments help uglifyJS to perform dead code elimination, because of heavy usage of higher order function they are not being able to get tree-shaken / dead-code eliminated properly by most of the tools (`rollup` might be able to tree-shake some of the calls due to recent improvements).

The added plugin is helping [`ramda`](https://github.com/ramda/ramda/blob/e58a9201838d9957b1a8e32bbccc52f6b5b07696/package.json#L67)'s users to reduce their bundle sizes.


char0n:
Thanks @Andarist. Can you point me to some resources regarding using `sideEffects` property in package.json ?

Andarist:
It's not really "live" yet, but the feature got merged in into webpack some time ago and will be part of the webpack@4 when it gets released. I've put it in the PR, because I think it's really a small, non-harming change, which will give benefits in the future.

https://github.com/webpack/webpack/pull/5815/files#diff-af5e3e160c1b7d4900eda0b2ed6f75edR5

Andarist:
Also keep in mind that my plugin is somewhat dangerous, although ramda-like libraries are actually the safest target for it ;) If you see any problems with it - let me know so I can fine tune the algorithm.

char0n:
I've just read this [article](https://iamakulov.com/notes/polished-webpack/). I understand what's going on, and I think we should be safe using your transformer.

Andarist:
That article in fact gave me an idea for creating the plugin :) and nowadays `polished` is using it to insert those comments too.

I hope one day they won't really be needed as they are pretty obscure, but as for today - well, whatever works.

Issue #35, opened 2017-03-02
Add promise <-> monad bridge functions


char0n:
~~resolveP~~, ~~allP~~, tapP, ~~thenP~~, catchP, thenCatchP, ~~rejectP~~, raceP, ~~reduceP~~, ~~isThenable~~, ~~isPromise~~, tryP

rockymadden:
Something like https://github.com/kirillrogovoy/promised-pipe might also be useful.

char0n:
@rockymadden isn't [pipeP](http://ramdajs.com/docs/#pipeP) what you're looking for ?

rockymadden:
If all functions are return promises, yes. The promised-pipe brings the ability to intermix, at any point, sync and async functions.

char0n:
pipeP does the same. The only requirement is that first function in the pipe must be promise generator (returns promise). 

```javascript
const { add, subtract }  = require('ramda');

const resolveP = Promise.resolve.bind(Promise);
const asyncAdd = val => resolveP(val).then(add(1));

pipeP(
  resolveP,
  add(1),
  subtract(1),
  asyncAdd,
  add(2),
  subtract(1)
)(1);
```

rockymadden:
Nice, hmm yeah ignore me then. Thanks!

char0n:
NP

rockymadden:
Thoughts on your `resolveP` function or (`R.bind(Promise.resolve, Promise)`) being made into a standard function in adjunct (e.g. `resolveP` or `ofP`)? Seems to be generally usable to kick things off as a Promise.

char0n:
Yep, the plan is to implement it. Checkout out the first comment of this issue. It is the first function to be implemented for this issue ;]

char0n:
You can provide a pull request for it if you wish and become a contributor. ramda-adjunct ecosystem is ready to receive PR's and I always welcome any help. What do you think ?

rockymadden:
Yep, love it. I've had a subset of this library internally for a while. Was reading through the Ramda wiki and saw `ramda-adjunct`, which was exactly what I myself was doing.

char0n:
Yep, I guess everybody is doing it. This library has a potential saving a lot of development time to a lot of people. Check this [article](https://www.linkedin.com/pulse/ramda-adjunct-vladim%C3%ADr-gorej) to find out more.

[...]

PR #229, opened 2017-12-01
feat: add isPromise

char0n:
Ref #35

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -619,6 +619,13 @@ declare namespace RamdaAdjunct {
          */
         isThenable(val: any): boolean;
 
+        /**
+         * Checks if input value is a native `Promise`.
+         * The Promise object represents the eventual completion (or failure)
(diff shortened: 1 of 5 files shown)


char0n:
Not sure. Should only work with native Promises. `isThenable` should work with other than native and promiseA+ compliant ones.

BjornMelgaard:
why it shouldn't work? https://runkit.com/embed/0oiabt527e0b 
bluebirds `promise.toString() === "[object Promise]"`

char0n:
I didn't know they are using toStringTag. There are myriads of promise implementations and on some it may work, on some it may not. That is why I created also `isThenable` that works on all PromiseA+ implementations.

But yeah, if Bluebird is using [Symbol.toStringTag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) then we will  treat it like a native promise.

Issue #66, opened 2017-05-06
Functional Javascript by Michael Fogus


char0n:
~~isSet~~, ~~isMap~~, isIndexed, ~~isTruthy~~, ~~isFalsy~~, ~~isEven~~, ~~isOdd~~, repeatedly, ~~fnull~~, ~~dispatch~~, hasKeys, trampoline, skipatke, pipeline, condition <-> validator pattern

char0n:
https://developer.mozilla.org/pl/docs/Glossary/Falsy

char0n:
@fogus just wanted to let you know, I read your book and it was really great. I compiled the list of some of the functions you described I would like to implement them into ramda-adjunct. There are couple of more that I would to add to the list, but have them stored in my notes on the iPad. Would you mind if I do so ?

Thank you very much for the book

fogus:
I do not mind at all.  Update me with a link when you get started.

char0n:
@fogus will sure do. Will be progressively implementing utils from the list spanning multiple releases. I can update you with every new release that will contain at least one of the utils of your design.

char0n:
@fogus `isOdd`, `isEvent` already in [v2.1.0](https://char0n.github.io/ramda-adjunct/2.1.0/). For next release v2.2.0, `isTruthy` and `isFalsy` is planned.

rockymadden:
Thrush combinator is known as `applyTo` in Ramda: https://github.com/ramda/ramda/pull/1847

char0n:
@rockymadden thrush combinator removed from list. Thanks

Undistraction:
@char0n Shall we get the remaining functions into their own issues?

char0n:
@Undistraction yep, but I need to do that (or anybody that read the @fogus book). We'll have to provide original implementation as a description for these separate issues and explanation from the book for possible usecases.

char0n:
I've created separate issue for [fnull](https://github.com/char0n/ramda-adjunct/issues/1235)

char0n:
I've created separate issue for [isIndexed](https://github.com/char0n/ramda-adjunct/issues/1236)

char0n:
repeatedly is basically an alias of [R.times](https://ramdajs.com/docs/#times)

char0n:
`hasKeys` will not be implemented as it seems very very specialized. But can be easily implemented using `R.allPass([R.has('key1'), R.has('key2')])`

char0n:
[...]

PR #239, opened 2017-12-09
feat: add isTruthy

char0n:
Ref #66


char0n:
@BjornMelgaard can you review pls ?

PR #245, opened 2017-12-18
feat: add defaultWhen

### LINKED ISSUE #216: defaultWhen
```javascript
// defaultToBy :: (a -> Boolean) -> b -> a -> a | b
const defaultToBy = curry((predicate, defaultVal, val) => (predicate(val) ? val : defaultVal));
```

char0n:
Closes #216

### DIFF
--- a/src/defaultWhen.js
+++ b/src/defaultWhen.js
@@ -1,0 +1,25 @@
+import { curry } from 'ramda';
+
+
+/**
+ * Returns the second argument if predicate function returns `true`,
+ * otherwise the third argument is returned.
+ *
+ * @func defaultWhen
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.2.0|v2.2.0}
+ * @category Logic
+ * @sig  (a -> Boolean) -> b -> a -> a | b
+ * @param {!function} predicate The predicate function
+ * @param {*} defaultVal The default value
+ * @param {*} val `val` will be returned instead of `defaultVal` if predicate returns false
+ * @return {*} The `val` if predicate returns `false`, otherwise the default value
+ * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
+ * @example
+ *
+ * RA.defaultWhen(RA.isNull, 1, null); // => 1
+ * RA.defaultWhen(RA.isNull, 1, 2); // => 2
+ */
+const defaultWhen = curry((predicate, defaultVal, val) => (predicate(val) ? defaultVal : val));
+
+export default defaultWhen;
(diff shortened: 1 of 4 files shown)


char0n:
@BjornMelgaard, @Undistraction can you validate this PR pls for me ? Does the logic of it makes sense ? Whould you expect to work like that ?

Thanks

Undistraction:
Can't actually run at the moment, but name, logic and intent all look good to me.

src/defaultWhen.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @func defaultWhen
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.2.0|v2.2.0}
+ * @category Logic
+ * @sig  (a -> Boolean) -> b -> a -> a | b
+ * @param {!function} predicate The predicate function
+ * @param {*} defaultVal The default value
+ * @param {*} val `val` will be returned instead of `defaultVal` if predicate returns false
+ * @return {*} The `val` if predicate returns `false`, otherwise the default value
```
BjornMelgaard:
```
* @param {b} defaultVal The default value
* @param {a} val `val` will be returned instead of `defaultVal` if predicate returns false
* @return {a|b} The `val` if predicate returns `false`, otherwise the default value

```

??

src/index.d.ts
```diff
@@ -639,6 +639,14 @@ declare namespace RamdaAdjunct {
          */
         isFalsy(val: any): boolean;
 
+        /**
+         * Returns the second argument if predicate function returns `true`,
+         * otherwise the third argument is returned.
+         */
+        defaultWhen(predicate: Function, defaultVal: any, val: any): any;
+        defaultWhen(predicate: Function, defaultVal: any): (val: any) => any;
+        defaultWhen(predicate: Function): (defaultVal: any) => (val: any) => any;
```
BjornMelgaard:
```
defaultWhen<DefVal, Val>(predicate: Function, defaultVal: DefVal, val: Val): DefVal | Val;
defaultWhen<DefVal, Val>((predicate: Function, defaultVal: DefVal): (val: Val) => DefVal | Val;
defaultWhen(predicate: Function): <DefVal, Val>((defaultVal: DefVal) => (val: Val) => DefVal | Val;

```

char0n: Will replace with those, thanks

src/defaultWhen.js
char0n: I guess we've already talked about this in past. JSDOC deals with JavaScript types. It has no concept of type variables. I guess what we could do is to use `@typedef` to define them, but we can only define new types globally and not locally, which is a problem...

That is why we introduced `@sig` jsodc tag, to overcome this limitation and describe the function interface in Hindley-Millner signatures.

BjornMelgaard: Oh, sorry, got it

char0n:
Typescript signatures replaced

PR #256, opened 2017-12-25
Typo in docs

sdaves:
Corrects a minor typo in the Identity equals documentation.


char0n:
Thanks for the fix. I was just wondering. I was just wondering are you using the Identity monad (if yes, what is your experience with it ?) or you just spotted to typo while getting thru the docs ?

Issue #173, opened 2017-10-19
Sequencing combinator (not S-combinator)


rockymadden:
> The seq combinator is used to loop over a sequence of functions. It takes two or more functions as parameters and returns a new function, which runs all of them in sequence against the same value. This is the implementation:

```javascript
const seq = function(/*funcs*/) {
    const funcs = Array.prototype.slice.call(arguments);
    return function (val) {
       funcs.forEach(function (fn) {
          fn(val);
       });
    };
};
```

> With it, you can perform a sequence of related, yet independent, operations. For instance, after finding the student object, you can use seq to both render it on the HTML page and log it to the console. All functions will run in that order against the same student object:

```javascript
const showStudent = R.compose(
  seq(
    append('#student-info'),
    consoleLog),
  csv,
  findStudent));
```

> The seq combinator doesn’t return a value; it just performs a set of actions one after the other. If you want to inject it into the middle of a composition, you can use R.tap to bridge the function with the rest.

From page 114 of [Functional Programming in JavaScript](https://www.goodreads.com/book/show/27876180-functional-programming-in-javascript)

char0n:
Ramda gives us limited S-combinator via [ap](http://ramdajs.com/docs/#ap). Do you really think that S-combinator is practical to have in the library ? I can achieve the same effect using `tap`.

```javascript
const showStudent = R.compose(
  tap(append('#student-info')),
  tap(consoleLog),
  csv,
  findStudent));
```

But I guess the selling point for it is its declarativness. It logically groups related operation into one sequence. You know what ? Let's go for it, just for having a reference implementation of S-combinator as a part of ramda-adjunct. But pls don't use Luis Atencio's implementation. Use utils that are already part of ramda or ramda-adjunct to compose the s-combinator.

rockymadden:
Yep, it's a complete convenience because, as you stated, it is possible to accomplish already.

char0n:
@rockymadden wanna go for this one in v2.1.0 release ? ;]

rockymadden:
I've been on back-to-back on call rotations and then slammed. I am interested in implementing this and other combinators for sure. I think the author is wrong, this is not the S-combinator as it is defined in combinatory logic. Sanctuary [used to have it](https://github.com/sanctuary-js/sanctuary/tree/v0.11.1#s--a-b-c---a-b---a---c). I need to dig in more why it was removed (renamed?). The S-combinator being something along the lines of:
```javascript
[...]

PR #257, opened 2017-12-26
feat: add sequencing combinator

rockymadden:
Closes https://github.com/char0n/ramda-adjunct/issues/173


src/index.d.ts
```diff
@@ -603,6 +603,15 @@ declare namespace RamdaAdjunct {
...
         rejectP<T>(value?: T): Promise<T>;
 
+        /**
+         * Accepts a list of side-effecting functions and returns a new function. When
+         * invoked, this new function is applied to some arguments, each side-effecting
+         * function is applied to those same arguments in the order in which they were
+         * declared. The return of each side-effecting function is discarded and the
+         * new function will always return `undefined`.
+         */
+        sequencing(fns: Array<Function>): Function;
```
char0n: let's primarily name this to `seq` (the function and the files) and add it alias `sequencing`.

```typescript
/**
 ... description
*/
seq(fns: Array<Function>): Function;
sequencing(fns: Array<Function>): Function; // alias
```

src/index.js
```diff
@@ -80,6 +80,7 @@ export { default as curryRightN } from './curryRightN';
 export { default as curryRight } from './curryRight';
 export { default as resolveP } from './resolveP';
 export { default as rejectP } from './rejectP';
+export { default as sequencing } from './sequencing';
```
char0n:
```javascript
export { default as seq } from './seq';
export { default as sequencing } from './seq';  // alias of seq

File: src/sequencing.js [hunk 3]
```diff
@@ -0,0 +1,29 @@
...
+
+/* eslint-disable max-len */
+/**
+ * Accepts a list of side-effecting functions and returns a new function. When
+ * invoked, this new function is applied to some arguments, each side-effecting
+ * function is applied to those same arguments in the order in which they were
+ * declared. The return of each side-effecting function is discarded and the
+ * new function will always return `undefined`.
+ *
+ * @func sequencing
```
char0n:
```javascript
@func seq

File: src/sequencing.js [hunk 4]
```diff
@@ -0,0 +1,29 @@
...
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Function
+ * @sig [((a, b, ...) -> x | undefined), ((a, b, ...) -> y | undefined), ...] -> (a -> b -> ... -> undefined)
+ * @param {Array} functions A list of side-effecting functions.
```
char0n:
```js
@param {Array} fns A list of side-effecting functions
```
Rename `functions` to `fns` to we are consistent with typescript and drop the dot at the end of the line

src/sequencing.js
```diff
@@ -0,0 +1,29 @@
...
+ * @return {Function} A new function.
```
char0n: drop the dot

src/sequencing.js
```diff
@@ -0,0 +1,29 @@
...
+ * @see {@link http://ramdajs.com/docs/#converge|converge}
+ * @example
+ *
+ * const log = message => console.log(message);
+ * const info = message => console.info(message);
+ *
+ * RA.sequencing([info, log])('foo'); //=> prints 'foo' via info then log
+ */
+/* eslint-enable max-len */
+const sequencing = converge(always(undefined));
```
char0n: We have a stub undefined function.
```js
const seq = converge(stubUndefined);

export default seq;

File: test/sequencing.js [hunk 7]
```diff
@@ -0,0 +1,81 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+
+describe('sequencing', function() {
```
char0n: Rename file to seq.js and use `describe('seq',...`

src/sequencing.js
```diff
@@ -0,0 +1,29 @@
+import { always, converge } from 'ramda';
+
+/* eslint-disable max-len */
+/**
+ * Accepts a list of side-effecting functions and returns a new function. When
```
char0n: I would drop the `side-effecting` adjective. It is not required for a function to be unpure for seq combinator to work correctly. Notice that ramda too doesn't mention it on tap combinator http://ramdajs.com/docs/#tap

src/sequencing.js
```diff
@@ -0,0 +1,29 @@
...
+ * @sig [((a, b, ...) -> x | undefined), ((a, b, ...) -> y | undefined), ...] -> (a -> b -> ... -> undefined)
+ * @param {Array} functions A list of side-effecting functions.
+ * @return {Function} A new function.
+ * @see {@link http://ramdajs.com/docs/#converge|converge}
+ * @example
+ *
+ * const log = message => console.log(message);
+ * const info = message => console.info(message);
+ *
+ * RA.sequencing([info, log])('foo'); //=> prints 'foo' via info then log
```
char0n: Example can be simplified to
```js
RA.seq([console.log, console.info])('foo');

File: src/sequencing.js (hunk 6 above)
Comment (char0n) [reply src/sequencing.js]: Anyway I am still struggling with the practicality of this combinator. Why does it always have to return undefined ? Why cannot it return at least first argument of the call ? The way it is implemented now makes it pretty hard to use it in any composition. I can use it in the following way, but in order for composition to work, the `seq` must be used as a last function in the composition.

```js
const showStudent = R.compose(
  seq(
    append('#student-info'),
    consoleLog
  ),
  csv,
  findStudent
);

File: src/index.d.ts (hunk 1 above)
Comment (rockymadden) [reply src/index.d.ts]: Will do, was unsure if some future, more general purpose, function would take the name.

File: src/index.js (hunk 2 above)
Comment (rockymadden) [reply src/index.js]: Will do.

File: src/sequencing.js (hunk 8 above)
Comment (rockymadden) [reply src/sequencing.js]: Will do, great point.

File: src/sequencing.js (hunk 3 above)
Comment (rockymadden) [reply src/sequencing.js]: Will do.

File: src/sequencing.js (hunk 4 above)
Comment (rockymadden) [reply src/sequencing.js]: Followed Ramda, but will rename.

File: src/sequencing.js (hunk 5 above)
Comment (rockymadden) [reply src/sequencing.js]: Followed Ramda, but will update.

File: src/sequencing.js (hunk 9 above)
Comment (rockymadden) [reply src/sequencing.js]: Ha, good point.

File: src/sequencing.js (hunk 6 above)
Comment (rockymadden) [reply src/sequencing.js]: Yeah, agreed.

File: test/sequencing.js (hunk 7 above)
Comment (rockymadden) [reply test/sequencing.js]: Will do.

File: src/sequencing.js [hunk 10]
```diff
@@ -0,0 +1,29 @@
...
+ * Accepts a list of side-effecting functions and returns a new function. When
+ * invoked, this new function is applied to some arguments, each side-effecting
+ * function is applied to those same arguments in the order in which they were
+ * declared. The return of each side-effecting function is discarded and the
+ * new function will always return `undefined`.
+ *
+ * @func sequencing
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Function
```
char0n: add
```js
@aliases sequencing

File: src/sequencing.js (hunk 6 above)
Comment (char0n) [reply src/sequencing.js]: Ok if agreed, I'am looking forward to the new signature then ;]

Comment (rockymadden) [reply src/sequencing.js]: I see a couple different return possibilities from the new function itself. With a rough equivalent of `fns => curryN(fns[0].length, (...args) => ...` (I'll be more defensive in actual implementation) we could:

1. Return the args, which is an array, back that the functions were applied to.
2. Return `nth(0)` of `args`.
3. Return `undefined` is args length is 0, `nth(0)` if args length is 1, and `identity` if args length is greater than 1.

Comment (char0n) [reply src/sequencing.js]: Let's go for `nth(0)` to make it consistent with `tap` (K combinator)

Comment (rockymadden) [reply src/sequencing.js]: Nice, should have something later today once I get a moment to myself. Thanks for the help!

Comment (rockymadden) [reply src/sequencing.js]: After looking at `tap`, it looks like it returns the first argument because the function must be urnary. `seq`, as it is currently defined, can accept any number of arguments and returning the first seems when there is more than one seems odd. Thoughts on if `seq` should only accept functions that are urnary OR rethinking/ensuring what are we are returning makes sense?

Comment (char0n) [reply src/sequencing.js]: Ok let me think about this before I answer. This is always the hardest part; making the util composable...

Comment (char0n) [reply src/sequencing.js]: Thought about it for some time. I'm for limiting the arity to 1, to make it predictable and easily composable. I can relate to authors of `tap` why they decided to the arity of 1. What you think ?

Comment (rockymadden) [reply src/sequencing.js]: Yeah, agreed.

**Comment by rockymadden** [comment]
@char0n Ready for any new feedback you have.

File: src/seq.js [hunk 11]
```diff
@@ -0,0 +1,27 @@
+import { curry, map, tap } from 'ramda';
+
```
char0n: add
```js
/* eslint-disable max-len */

File: src/seq.js [hunk 12]
```diff
@@ -0,0 +1,27 @@
...
+ *        thrown away
+ * @param {*} x
+ * @return {*} `x`
+ * @see {@link http://ramdajs.com/docs/#tap|tap}
+ * @see {@link http://www.cs.rpi.edu/academics/courses/spring11/proglang/handouts/lambda-calculus-chapter.pdf|sequencing combinator detailed}
+ * @example
+ *
+ * RA.seq([console.info, console.log])('foo'); //=> prints 'foo' via info then log
+ */
+
```
char0n: add 
```js
/* eslint-enable */

File: src/seq.js [hunk 13]
```diff
@@ -0,0 +1,27 @@
...
+ * Also known as the normal order sequencing combinator.
+ *
+ * Acts as a transducer if a transformer is given as second parameter.
+ *
+ * @func seq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Function
+ * @sig [(a -> *), (a -> *), ...] -> a -> a
+ * @param {Array} fns The list of functions to call in order with `x` whose return values will be
```
char0n: don't wrap the line, let it be as long as it can be. We disabled linter for max len on jsdoc

src/seq.js
```diff
@@ -0,0 +1,27 @@
...
+ *        thrown away
+ * @param {*} x
+ * @return {*} `x`
+ * @see {@link http://ramdajs.com/docs/#tap|tap}
+ * @see {@link http://www.cs.rpi.edu/academics/courses/spring11/proglang/handouts/lambda-calculus-chapter.pdf|sequencing combinator detailed}
```
char0n: Lets bee consistent with `Y` combinator link label and let's say `sequencing combinator explained`

https://github.com/char0n/ramda-adjunct/blob/9e2dcf1071bc570c17f24c8645be447ce1c1a55e/src/Y.js#L15

src/seq.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Runs the given list of functions in order with the supplied object, then returns the object.
+ * Also known as the normal order sequencing combinator.
+ *
+ * Acts as a transducer if a transformer is given as second parameter.
+ *
+ * @func seq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Function
```
char0n: Add @aliases tag
```js
@aliases sequencing
```

src/seq.js
```diff
@@ -0,0 +1,27 @@
...
+ * @sig [(a -> *), (a -> *), ...] -> a -> a
+ * @param {Array} fns The list of functions to call in order with `x` whose return values will be
+ *        thrown away
+ * @param {*} x
+ * @return {*} `x`
+ * @see {@link http://ramdajs.com/docs/#tap|tap}
```
char0n: Join multiple see declarations into one on one line. 

Example: https://github.com/char0n/ramda-adjunct/blob/8f0b75209cdbaa55f3859e0d1b9e83cac12fcedc/src/isNotAsyncFunction.js#L17

src/seq.js
```diff
@@ -0,0 +1,27 @@
...
+ * @see {@link http://www.cs.rpi.edu/academics/courses/spring11/proglang/handouts/lambda-calculus-chapter.pdf|sequencing combinator detailed}
+ * @example
+ *
+ * RA.seq([console.info, console.log])('foo'); //=> prints 'foo' via info then log
+ */
+
+const seq = curry((fns, x) => tap(_ => map(fn => fn(_))(fns))(x));
```
char0n: Please don't use underscore for naming the arguments. Name it to something more declarative.

test/seq.js
```diff
@@ -0,0 +1,37 @@
+import { append, compose, filter, flip, inc, map, range, transduce } from 'ramda';
+import { isEven, noop, seq, sequencing } from '../src/index';
```
char0n: Replace with following to be consitent with other tests; then use `RA.funcName`
```js 
import * as RA from '../src/index';
```

test/seq.js
```diff
@@ -0,0 +1,37 @@
+import { append, compose, filter, flip, inc, map, range, transduce } from 'ramda';
```
char0n: Replace with following to be consistent with other tests; then use `R.funcName`
```js
import * as R from 'ramda';
```

test/seq.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('tests transducing', function() {
+    const transducer = compose(map(inc), seq([inc]), filter(isEven));
+    const transduced = transduce(transducer, flip(append), [], range(0, 10));
+
+    eq(transduced, [2, 4, 6, 8, 10]);
+  });
+
+  it('tests exported alias', function() {
```
char0n:
```js

src/seq.js
rockymadden: This was indeed changed in commit.

Issue #234, opened 2017-12-08
Negated Logic


Undistraction:
Would you be open to a PR for the following negations of Ramda's logical functions:

```
export const nand = complement(and);
export const nor = complement(or);
export const notBoth = compose(complement, both);
export const neither = compose(complement, either);
export const notAllPass = compose(complement, allPass);
export const nonePass = compose(complement, anyPass);
```

I've found them to make my code a little more readable.

Let's deal with complements one at a time. 

- [ ] [nand](https://github.com/char0n/ramda-adjunct/issues/237)

char0n:
Let me think about them and will let you know today evening/tommorow. But most of them looks legitimate.

Undistraction:
Great. I find that with logic, this kind of extra clarity really helps me avoid brain melt. It's never good having to flip booleans in your mind to understand something. As far as naming goes, `nand` and `nor` are standard for [logic gates](https://en.wikipedia.org/wiki/Logic_gate), and the others feel like the most obvious names possible. Possibly `neither` could be `notEither` but I don't see much advantage to that.

Undistraction:
Thinking about this a little further, Ramda also currently offers `all`, `any`, `none`, but doesn't offer  `notAll`, so that would probably be worth adding as well:

```
const notAll = complement(all);
```

char0n:
Let's deal with complements one at a time. 

- [ ] [nand](https://github.com/char0n/ramda-adjunct/issues/237)

char0n:
Meanwhile lets break down the naming

LGTM:
```javascript
export const nand = complement(and);
export const nor = complement(or);
```

These needs some additional thinking. I am not sure about these. IMHO they can make code harder to read and reason about. 

```javascript
export const notBoth = compose(complement, both);
export const notAllPass = compose(complement, allPass);
export const neither = compose(complement, either);
export const nonePass = compose(complement, anyPass);
```

For me personally it is easier to read something like this
```javascript
both(isNotUndefined, isNotNull)
```
instead of
```javascript
notBoth(isUndefined, isNull)
```

We should strive for simplicity, but these functions can really IMHO make code harder to read. Our aim should to create complements for `Type` functions which are inputs for these HOC `Logic` functions and not to complement the `Logic` functions them selves.

Undistraction:
> For me personally it is easier to read something like this
> both(isNotUndefined, isNotNull)

[...]

PR #260, opened 2017-12-31
feat(api): add notBoth function

Undistraction:
Add `notBoth` function which is complement of Ramda's `both`.

- I've kept the docs close to Ramda's docs for `both`.
- I've assumed the next release will be 2.3.0.
- I've had to add Sanctuary as a dev dependency to replicate the fantasy-land tests.
- I'll add the other complements in separate PRs once this one is approved.

Ref #234

### DIFF
--- a/src/notBoth.js
+++ b/src/notBoth.js
@@ -1,5 +1,7 @@
 import { compose, complement, both } from 'ramda';
 
+
+/* eslint-disable max-len */
 /**
  * A function which calls the two provided functions and returns the complement of `&&`ing the
  * results.
@@ -14,22 +16,23 @@ import { compose, complement, both } from 'ramda';
  * @func notBoth
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
- * @category Type
+ * @category Logic
  * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
  * @param {Function} f A predicate
  * @param {Function} g Another predicate
- * @return {Function} a function that applies its arguments to `f` and `g` and returns the
- * complement of `&&`ing their outputs together.
+ * @return {Function} Returns a function that applies its arguments to `f` and `g` and returns the complement of `&&`ing their outputs together.
  * @example
  *
  * const gt10 = R.gt(R.__, 10)
  * const even = (x) => x % 2 === 0;
  * const f = RA.notBoth(gt10, even);
+ *
  * f(12); //=> false
  * f(8); //=> true
  * f(11); //=> true
  * f(9); //=> true
  */
-const _notBoth = compose(complement, both);
+/* eslint-enable max-len */
+const notBoth = compose(complement, both);
 
-export default _notBoth;
+export default notBoth;
--- a/src/notBoth.js
+++ b/src/notBoth.js
@@ -1,0 +1,35 @@
+import { compose, complement, both } from 'ramda';
+
+/**
+ * A function which calls the two provided functions and returns the complement of `&&`ing the
+ * results.
+ * It returns true if the first function is false-y and the complement of the second function
+ * otherwise. Note that this is short-circuited, meaning that the second function will not be
+ * invoked if the first returns a false-y value. In short it will return true unless both predicates
+ * return true.
+ *
+ * In addition to functions, `RA.notBoth` also accepts any fantasy-land compatible
+ * applicative functor.
+ *
+ * @func notBoth
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Type
+ * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
+ * @param {Function} f A predicate
+ * @param {Function} g Another predicate
+ * @return {Function} a function that applies its arguments to `f` and `g` and returns the
+ * complement of `&&`ing their outputs together.
+ * @example
+ *
+ * const gt10 = R.gt(R.__, 10)
+ * const even = (x) => x % 2 === 0;
+ * const f = RA.notBoth(gt10, even);
+ * f(12); //=> false
+ * f(8); //=> true
+ * f(11); //=> true
+ * f(9); //=> true
+ */
+const _notBoth = compose(complement, both);
+
+export default _notBoth;
--- a/src/notBoth.js
+++ b/src/notBoth.js
@@ -21,6 +21,7 @@ import { compose, complement, both } from 'ramda';
  * @param {Function} f A predicate
  * @param {Function} g Another predicate
  * @return {Function} Returns a function that applies its arguments to `f` and `g` and returns the complement of `&&`ing their outputs together.
+ * @see {@link http://ramdajs.com/docs/#both|both}
  * @example
  *
  * const gt10 = R.gt(R.__, 10)
--- a/test/notBoth.js
+++ b/test/notBoth.js
@@ -6,18 +6,18 @@ import eq from './shared/eq';
 
 
 const supportsFantasyLand = () => {
-  let result;
-  // Earlier versions of Ramda will throw.
   try {
-    result = RA.notBoth(Just(true), Just(true));
+    return RA.notBoth(Just(true), Just(true)).equals(Just(false));
   } catch (e) {
     return false;
   }
-  return Just(false).equals(result);
 };
 
 describe('notBoth', function() {
-  const isFantasyLandSupported = supportsFantasyLand();
+  let isFantasyLandSupported;
+  before(function() {
+    isFantasyLandSupported = supportsFantasyLand();
+  });
 
   it('combines two boolean-returning functions into one', function() {
     const even = x => x % 2 === 0;
--- a/test/notBoth.js
+++ b/test/notBoth.js
@@ -1,0 +1,47 @@
+import S from 'sanctuary';
+
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('notBoth', function() {
+  it('combines two boolean-returning functions into one', function() {
+    const even = function(x) { return x % 2 === 0 };
+    const gt10 = function(x) { return x > 10 };
+    const f = RA.notBoth(even, gt10);
+    eq(f(8), true);
+    eq(f(13), true);
+    eq(f(9), true);
+    eq(f(14), false);
+  });
+
+  it('accepts functions that take multiple parameters', function() {
+    const between = function(a, b, c) { return a < b && b < c };
+    const total20 = function(a, b, c) { return a + b + c === 20 };
+    const f = RA.notBoth(between, total20);
+    eq(f(4, 5, 11), false);
+    eq(f(12, 2, 7), true);
+    eq(f(12, 2, 6), true);
+    eq(f(5, 6, 15), true);
+  });
+
+  it('does not evaluate the second expression if the first one is false', function() {
+    let effect = 'not evaluated';
+    const F = function() { return false };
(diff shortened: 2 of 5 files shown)


src/notBoth.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Function} f A predicate
+ * @param {Function} g Another predicate
+ * @return {Function} a function that applies its arguments to `f` and `g` and returns the
+ * complement of `&&`ing their outputs together.
+ * @example
+ *
+ * const gt10 = R.gt(R.__, 10)
+ * const even = (x) => x % 2 === 0;
+ * const f = RA.notBoth(gt10, even);
+ * f(12); //=> false
```
char0n: insert one empty line here between var declarations and other statements

src/notBoth.js
```diff
@@ -0,0 +1,35 @@
...
+ * f(8); //=> true
+ * f(11); //=> true
+ * f(9); //=> true
+ */
+const _notBoth = compose(complement, both);
```
char0n: why the underscore here ?

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
+import S from 'sanctuary';
+
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
```
char0n: insert additional empty line here pls. Would be great to have this rule in eslint rules. Always 2 lines between imports and the actual code.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+describe('notBoth', function() {
+  it('combines two boolean-returning functions into one', function() {
+    const even = function(x) { return x % 2 === 0 };
```
char0n: Use arrow functions where binding is not required.
```js
const isEven = x => x % 2  === 0
```

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    const gt10 = function(x) { return x > 10 };
+    const f = RA.notBoth(even, gt10);
+    eq(f(8), true);
```
char0n: insert empty line between the variable declarations and other statements.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    eq(f(13), true);
+    eq(f(9), true);
+    eq(f(14), false);
+  });
+
+  it('accepts functions that take multiple parameters', function() {
+    const between = function(a, b, c) { return a < b && b < c };
+    const total20 = function(a, b, c) { return a + b + c === 20 };
+    const f = RA.notBoth(between, total20);
+    eq(f(4, 5, 11), false);
```
char0n: insert empty line between the variable declarations and other statements.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    eq(f(12, 2, 7), true);
+    eq(f(12, 2, 6), true);
+    eq(f(5, 6, 15), true);
+  });
+
+  it('does not evaluate the second expression if the first one is false', function() {
+    let effect = 'not evaluated';
+    const F = function() { return false };
+    const Z = function() { effect = 'Z got evaluated' };
+    RA.notBoth(F, Z)();
```
char0n: insert empty line between the variable declarations and other statements.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    eq(effect, 'not evaluated');
+  });
+
+  it('accepts fantasy-land applicative functors', function() {
+    const { Just, Nothing } = S;
```
char0n: insert empty line between the variable declarations and other statements.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    const f = RA.notBoth(even, gt10);
+    eq(f(8), true);
+    eq(f(13), true);
+    eq(f(9), true);
+    eq(f(14), false);
+  });
+
+  it('accepts functions that take multiple parameters', function() {
+    const between = function(a, b, c) { return a < b && b < c };
+    const total20 = function(a, b, c) { return a + b + c === 20 };
```
char0n: Use arrow functions where binding is not required.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    const f = RA.notBoth(between, total20);
+    eq(f(4, 5, 11), false);
+    eq(f(12, 2, 7), true);
+    eq(f(12, 2, 6), true);
+    eq(f(5, 6, 15), true);
+  });
+
+  it('does not evaluate the second expression if the first one is false', function() {
+    let effect = 'not evaluated';
+    const F = function() { return false };
```
char0n: Use arrow functions where binding is not required.

test/notBoth.js
```diff
@@ -0,0 +1,47 @@
...
+    const Z = function() { effect = 'Z got evaluated' };
```
char0n: Use sinon stubs for things like that. We already have in dev deps.

test/notBoth.js
char0n: use lowercase names for functions that will not be instantiated.

test/notBoth.js
char0n: More optimal would be to do that as import statement
```js
import { Just, Nothing } from S;
```
But anyway ramda-adjunct is already using monadic types from monet librarie, and we are aiming to have our own in the future (see [here](https://github.com/char0n/ramda-adjunct/blob/e9260f67a44954d0c3ca350483fb0b1a9bab49ff/src/fantasy-land/Identity.js#L16)). No need to use sanctuary for now.

See here how we already utilize [monet](https://github.com/char0n/ramda-adjunct/blob/baa62ea258deb95ed85b83cbc2a9c8a1722fbdc8/test/liftF.js#L7) for our tests.

char0n:
Also be aware that the tests did not pass on CircleCi.

test/notBoth.js
Undistraction: Yep. Would be good to lint for this.

src/notBoth.js
Undistraction: Was jumping between projects and carried it over.

Undistraction:
These tests are passing locally on Node 6,7 and 8. I'll try and dig into this tomorrow.

char0n:
Strange, let me know what was wrong. I've rerun the test again.

Undistraction:
If you pull it and run the tests does it fail for you?

char0n:
Yes it does fail. Seems there is problem with lower ramda versions. We are currently supporting ramda `>= 0.19.0 <= 0.25.0`. It seems there are problems with `notBoth` on lower ramda versions.

You can reproduce by running `npm run test:ramda` which tests compatibility of ramda-adjunct with various ramda versions.

Undistraction:
This makes sense as it looks like Ramda only started supporting FantasyLand as of 0.24.0 ([Upgrade guide](https://github.com/ramda/ramda/issues/2183)) so how do you want to proceed?

char0n:
I suggest to use feature detection instead of relaying in version number. 
Explanation:

```js
describe('some test suite', function () {
  let isFantasyLandSupported = false;

  beforeEach(function() {
    // ...logic for detecting if both is fantasy land compatible
    isFantasyLandSupported = true | false;
  });

  // then in fantasy land related tests...
  it('test fl', function () {
    if (!isFantasyLandSupported) { this.skip(); }

    ... the rest of the test
  });
});


```

Undistraction:
OK. It definitely makes sense to use feature detection instead, but how do you suggest we check for support? Do we want to rely on an error raised by calling `both` with an invalid value (`Just` or `Nothing`)?

char0n:
Yeah, if the result of the `both` on monads is throwing an error, then we should rely on it as a way to do feature detection.

Undistraction:
I have all tests passing locally.

I've moved the feature detection to the describe block - I don't really see the advantage of running it before each test, but if you would prefer that I'll amend.

I also hit problems with testem which ignored the `this.skip()` and ran the tests anyway, causing them to fail in versions that don't support them so I have wrapped the test in a conditional.

test/notBoth.js
```diff
@@ -5,7 +5,20 @@ import * as RA from '../src/index';
...
+  try {
+    result = RA.notBoth(Just(true), Just(true));
+  } catch (e) {
+    return false;
+  }
+  return Just(false).equals(result);
+};
+
 describe('notBoth', function() {
+  const isFantasyLandSupported = supportsFantasyLand();
```
char0n: If you don't want it to run on every test (which I agree doesn't make sense) just wrap it into [before](https://mochajs.org/#hooks) block to run it only once and semantically divides it.

```js
let  isFantasyLandSupported;

before(function () {
  isFantasyLandSupported = supportsFantasyLand();
});

File: test/notBoth.js [hunk 13]
```diff
@@ -5,7 +5,20 @@ import * as RA from '../src/index';
 import eq from './shared/eq';
 
 
+const supportsFantasyLand = () => {
```
char0n: Seems a little bit cryptic. Can we simplify to following ?

```js
const supportsFantasyLand = () => {
  try {
    return RA.notBoth(Just(true), Just(true)).equals(Just(false));
  } catch (e) {
    return false;
  }
}

File: test/notBoth.js [hunk 14]
```diff
@@ -36,15 +49,17 @@ describe('notBoth', function() {
...
-    eq(RA.notBoth(Just(true), Just(false)), Just(true));
-    eq(RA.notBoth(Just(false), Just(true)), Just(true));
-    eq(RA.notBoth(Just(false), Just(false)), Just(true));
-    eq(RA.notBoth(Just(true), Nothing()), Nothing());
-    eq(RA.notBoth(Nothing(), Just(true)), Nothing());
-    eq(RA.notBoth(Nothing(), Just(false)), Nothing());
-    eq(RA.notBoth(Just(false), Nothing()), Nothing());
-    eq(RA.notBoth(Nothing(), Nothing()), Nothing());
-  });
+  if (isFantasyLandSupported) {
```
char0n: Otherwise your solution is pristine. Good job ! ;]

src/notBoth.js
```diff
@@ -0,0 +1,38 @@
...
+ * In addition to functions, `RA.notBoth` also accepts any fantasy-land compatible
+ * applicative functor.
+ *
+ * @func notBoth
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Logic
+ * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
+ * @param {Function} f A predicate
+ * @param {Function} g Another predicate
```
char0n: My last PR comment ;] 
Thank your for you patience and sorry for so much comments. Trying to keep the codebase consistent and top quality.

Insert `@see` tag to reference to original `both` function
```js
@see {@link http://ramdajs.com/docs/#both|both}
```

test/notBoth.js
Undistraction: I tried this, but unfortunately the conditional around the suite is evaluated immediately, meaning the value of `isFantastyLandSupported` is `undefined`.

char0n: Ahh..true, then disregard my comment.

test/notBoth.js
Undistraction: I'm unsure of the change you're asking for here.

src/notBoth.js
Undistraction: Don't apologise. It's really helpful.

test/notBoth.js
Undistraction: Agreed.

test/notBoth.js
char0n: No change required, just a general comment about your solution.

Undistraction: Ok. In that case - thank you.

PR #269, opened 2018-01-07
feat: add allP

char0n:
Ref #35

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "babel-register": "=6.26.0",
     "better-npm-run": "=0.1.0",
     "chai": "=4.1.2",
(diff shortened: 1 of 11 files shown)


char0n:
@BjornMelgaard can you have a look at this ? Thanks

Undistraction:
Looks good to me. But unclear what `fail-fash` means - test/allP.js:44.

Also maybe add a test with two `RA.rejectP`s, verifying the first rejection message is used.

char0n:
@Undistraction, @BjornMelgaard thanks, all comments addressed.

Undistraction:
Nice work.

PR #273, opened 2018-01-10
feat: add neither

Undistraction:
Add `neither` which is a complement of Ramda's `either`.

Ref #234

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -711,7 +711,7 @@ declare namespace RamdaAdjunct {
         notBoth(firstPredicate: Function, secondPredicate: Function): Function;
 
          /**
-          * A function which calls the two provided functions and returns the complement of `|`ing
+          * A function which calls the two provided functions and returns the complement of `||`ing
           * the results. It returns false if the first function is truth-y and the complement of the
           * second function otherwise. Note that this is short-circuited, meaning that the second
           * function will not be invoked if the first returns a truth-y value. In short it will
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -710,6 +710,18 @@ declare namespace RamdaAdjunct {
           */
         notBoth(firstPredicate: Function, secondPredicate: Function): Function;
 
+         /**
+          * A function which calls the two provided functions and returns the complement of `|`ing
+          * the results. It returns false if the first function is truth-y and the complement of the
+          * second function otherwise. Note that this is short-circuited, meaning that the second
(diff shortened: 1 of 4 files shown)


src/neither.js
```diff
@@ -0,0 +1,39 @@
...
+ * applicative functor.
+ *
+ * @func neither
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.3.0|v2.3.0}
+ * @category Logic
+ * @sig (*... -> Boolean) -> (*... -> Boolean) -> (*... -> Boolean)
+ * @param {Function} f A predicate
+ * @param {Function} g Another predicate
+ * @return {Function} Returns a function that applies its arguments to `f` and `g` and returns the complement of `|`ing their outputs together.
```
char0n: | -> ||

src/neither.js
```diff
@@ -0,0 +1,39 @@
...
+ * @see {@link http://ramdajs.com/docs/#either|either}
```
char0n: @see {@link http://ramdajs.com/docs/#either|either}, {@link http://ramdajs.com/docs/#or|or}

test/neither.js
```diff
@@ -0,0 +1,62 @@
...
+    const total20 = (a, b, c) => a + b + c === 20;
+    const f = RA.neither(between, total20);
+
+    eq(f(4, 5, 11), false);
+    eq(f(12, 2, 6), false);
+    eq(f(5, 6, 15), false);
+    eq(f(12, 2, 7), true);
+  });
+
+  it('does not evaluate the second expression if the first one is true', function() {
```
char0n: would already use a context/specify. I know you we trying to be consistent with ramda, but this make more sense from the perspective of viewer of the tests
```js
context('when the result of first function is true', function () {
   specify('should not evaluate the second expression', function () {
     ...
   });
));

File: src/index.d.ts [hunk 4]
```diff
@@ -710,6 +710,18 @@ declare namespace RamdaAdjunct {
           */
         notBoth(firstPredicate: Function, secondPredicate: Function): Function;
 
+         /**
+          * A function which calls the two provided functions and returns the complement of `|`ing
```
char0n: | => ||

The operations are not the same

src/neither.js
char0n: This has to be addressed too pls

char0n:
Waiting for tests and merging. Nicely done.

Issue #279, opened 2018-01-15
What about a RA.isNotEmptyArray ?


guillaumearm:
Hello everyone,

Frequently, I implement myself a `isNotEmptyArray` function : 

```js
const isNotEmptyArray = R.both(RA.isArray, RA.isNotEmpty);
```

I think I'm not alone to doing that, maybe I could make a PR for adding this function ?

We should add `isEmptyArray`, `isEmptyObject` and `isNotEmptyObject` too.

char0n:
Yes, I have the same usecase in the current codebase I'm working on. But I'd go for a following naming:
`isEmptyArray`, `isNonEmptyArray`.

Why ? You can read more about why here: https://github.com/char0n/ramda-adjunct/issues/40

#40 is also dealing with `isEmptyString` and `isNonEmptyString`.

You can own both this and #40 issue and make them realize ;]

guillaumearm:
Yup, i prefer the 'isNonEmpty'  terminology.

PR coming tommorow ;-)

and I would like to say `ramda-adjunct` is an awesome library, the missing piece of `ramda`.

Thank you for your work.

char0n:
Thanks, and you'll be the next one to make it even better ;]

Please make two separate PRs for `isEmptyArray` and `isNonEmptyArray`. Check out this [PR](https://github.com/char0n/ramda-adjunct/pull/258/files) as an example of how a PR should look like.

guillaumearm:
And what do you think about `isEmptyObject` and `isNonEmptyObject` ?

char0n:
@guillaumearm let's continue object discussion here https://github.com/char0n/ramda-adjunct/issues/283

char0n:
From some reason I don't see `isNotEmptyArray` in generated docs for v2.4.0 https://char0n.github.io/ramda-adjunct/2.4.0/index.html

guillaumearm:
'cause it's `isNonEmptyArray`

char0n:
Of course ;]

PR #288, opened 2018-01-17
feat: add isEmptyArray

guillaumearm:
Ref #279

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -39,6 +39,11 @@ declare namespace RamdaAdjunct {
         isArray(val: any): val is Array<any>;
 
         /**
+         * Checks if input value is an empty `Array`.
+         */
+        isEmptyArray(val: any): val is Array<any>;
+
+        /**
          * Checks if input value is `Boolean`.
          */
         isBoolean(val: any): val is boolean;
--- a/src/isEmptyArray.js
+++ b/src/isEmptyArray.js
@@ -24,6 +24,6 @@ import isArray from './isArray';
  * RA.isEmptyArray(42); // => false
  * RA.isEmptyArray('42'); // => false
  */
-const isEmptyArray = both(isEmpty, isArray);
+const isEmptyArray = both(isArray, isEmpty);
 
 export default isEmptyArray;
--- a/src/isEmptyArray.js
+++ b/src/isEmptyArray.js
@@ -1,0 +1,29 @@
+import { both, isEmpty } from 'ramda';
+
+import isArray from './isArray';
+
+
+/**
+ * Checks if input value is an empty `Array`.
+ *
+ * @func isEmptyArray
+ * @memberOf RA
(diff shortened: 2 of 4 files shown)


guillaumearm:
@char0n : any suggestions ?

char0n:
Will get to this in a couple of hours. Thanks

src/isEmptyArray.js
```diff
@@ -0,0 +1,29 @@
...
+import isArray from './isArray';
+
+
+/**
+ * Checks if input value is an empty `Array`.
+ *
+ * @func isEmptyArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.4.0|v2.4.0}
+ * @category Type
```
char0n: Thinking if this still falls into `Type` category or is already an `Logic`. @guillaumearm what's your opinion on this ?

Undistraction: It might be using logic under the hood, but its API and usage put it in Type imho.

char0n: Agree

guillaumearm: In fact, it's a mix of both : Type and Logic

Because an empty array and a non empty array have the same type, it's a very good question to ask.
Is it ok to passing same type to a `Type` function produce a different result ?

on another side `RA.isNotEmpty` should be in `Logic` category like R.isEmpty

char0n: Thanks for creating the issue for it. I’ll handle that. Regarding the question whether this is type or logic... I would say we are still in type world. If we look at it throug the lens of category theory let us say we have two new types **EmptyArray** and **NonEmptyArray**.  Assuming this we cas say that our two new type functions are just checking out the correct types. Logic of the checking becomes only the implementation detail.

char0n:
Pls rebase this pr on top of current master also

src/isEmptyArray.js
guillaumearm: It seems correct, Agree ^^

guillaumearm:
Should I --amend my `feat` commit ?

char0n:
No need to, you have to rebase your branch `guillaumearm:feat/isEmptyArray` on top of char0n:master and resolve conflicts. I you don't know how to do that, push one additional commit that deals with the PR review comments and I'll deal with conflicts.

guillaumearm:
No no i'm ok with that ;-) thanks

src/index.d.ts
```diff
@@ -38,6 +38,11 @@ declare namespace RamdaAdjunct {
          */
         isArray(val: any): val is Array<any>;
 
+        /**
+         * Checks if input value is an empty `Array`.
+         */
+        isEmptyArray(val: any): val is Array<any>;
```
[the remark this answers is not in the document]
BjornMelgaard: same as here https://github.com/char0n/ramda-adjunct/pull/289#discussion_r162347905

PR #289, opened 2018-01-17
feat: add isNonEmptyArray

guillaumearm:
Ref #279

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -54,6 +54,11 @@ declare namespace RamdaAdjunct {
         isNotArray(val: any): boolean;
 
         /**
+         * Checks if input value is a non empty `Array`.
+         */
+        isNonEmptyArray(val: any): val is Array<any>;
+
+        /**
          * Checks if input value is complement of `Boolean`.
          */
         isNotBoolean(val: any): boolean;
--- a/src/isNonEmptyArray.js
+++ b/src/isNonEmptyArray.js
@@ -24,6 +24,6 @@ import isArray from './isArray';
  * RA.isNonEmptyArray(42); // => false
  * RA.isNonEmptyArray('42'); // => false
  */
-const isNonEmptyArray = both(isNotEmpty, isArray);
+const isNonEmptyArray = both(isArray, isNotEmpty);
 
 export default isNonEmptyArray;
--- a/src/isNonEmptyArray.js
+++ b/src/isNonEmptyArray.js
@@ -1,0 +1,29 @@
+import { both } from 'ramda';
+
+import isNotEmpty from './isNotEmpty';
+import isArray from './isArray';
+
+/**
+ * Checks if input value is not an empty `Array`.
+ *
+ * @func isNonEmptyArray
+ * @memberOf RA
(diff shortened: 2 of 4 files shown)


test/isNonEmptyArray.js
```diff
@@ -0,0 +1,23 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
```
char0n: insert one new empty line pls

src/isNonEmptyArray.js
```diff
@@ -0,0 +1,29 @@
+import { both } from 'ramda';
+
+import isNotEmpty from './isNotEmpty';
+import isArray from './isArray';
+
```
char0n: insert one new empty line pls

src/index.d.ts
```diff
@@ -53,6 +53,11 @@ declare namespace RamdaAdjunct {
          */
         isNotArray(val: any): boolean;
 
+        /**
+         * Checks if input value is a non empty `Array`.
+         */
+        isNonEmptyArray(val: any): val is Array<any>;
```
char0n: @BjornMelgaard I'm not sure about this. Can you advice ? Does this signature make sense in the context of empty array ?

src/isNonEmptyArray.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * RA.isNonEmptyArray([42]); // => true
+ * RA.isNonEmptyArray([]); // => false
+ * RA.isNonEmptyArray({}); // => false
+ * RA.isNonEmptyArray(null); // => false
+ * RA.isNonEmptyArray(undefined); // => false
+ * RA.isNonEmptyArray(42); // => false
+ * RA.isNonEmptyArray('42'); // => false
+ */
+const isNonEmptyArray = both(isNotEmpty, isArray);
```
char0n: I'd prefer one of the following compositions:
```js
const isNonEmptyArray = both(isArray, isNotEmpty);
```
or
```js
const isNonEmptyArray = both(isArray, complement(isEmptyArray)); 
```

src/index.d.ts
BjornMelgaard: @char0n I think yes, there is no way we can express that array is non-empty in types

char0n: Wouldn't be better to use `boolean` instead ?

BjornMelgaard: Look
https://www.typescriptlang.org/play/index.html#src=declare%20function%20isNonEmptyArray(val%3A%20any)%3A%20val%20is%20Array%3Cany%3E%3B%0D%0A%0D%0Afunction%20createName(name%3A%20any)%20%7B%0D%0A%20%20%20%20if%20(isNonEmptyArray(name))%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20name.join()%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20else%20%7B%0D%0A%20%20%20%20%20%20%20%20return%20name.join(%22%20%22)%3B%0D%0A%20%20%20%20%7D%0D%0A%7D

when `isNonEmptyArray` returns `false` array methods still can be applied on `name`

so `val is Array<any>` if fine )

Issue #40, opened 2017-03-08
isEmptyString & isNonEmptyString


char0n:

char0n:
@tycho01 need some collaboration over this. Does this even make sense. Thought I have one production use case in my current project.

```javascript
const description = 'user input';
if (isNotNull(description) && isNotEmptyString(description)) { return description }
```

tycho01:
Sounds like just `if (description)` covers this use-case? That said, I think this whole function-based approach primarily shines if you're doing composition/`map`. That is, without need for a function one could technically be shorter off just writing `description != ''` for this.

But yeah, why not? :)

So I've personally tended toward higher abstraction, and wanted a slightly different truthy/falsy check for arrays/objects, here's what I've been using for that.
```
export let falsy: (v: any) => boolean = R.either(R.isEmpty, R.not);
export let truthy: (v: any) => boolean = R.complement(falsy); // differs from JS's 'truthy': []/{} -> false.
```
For other cases, in the case one would prefer higher abstraction, where you want a function version of the built-in truthy/falsy check, one could also use the built-in `Boolean` (/ `R.identity`) or `R.not` for positive/negative checks respectively.

char0n:
> Sounds like just if (description) covers this use-case? That said, I think this whole function-based 
> approach primarily shines if you're doing composition/map. That is, without need for a function one 
> could technically be shorter off just writing description != '' for this.

`if (description)` would cover the usecase, but we have a airbnb codestyle where everything should be checked explicitly except for booleans. As for imperative over declarative system here, I like the first version better, I can read it faster and I know instantly what is going on there.

```javascript
if (isNotNull(description) && isNotEmptyString(description)) { return description }`
if (description !== null && description !== '') { return description }
```

Also I like you falsy, truhy approach ;]

tycho01:
Yeah. I wish the built-in checks used this logic for truthy/falsy; it could make things a bit more terse when checking objects/arrays...

char0n:
Back to the original problem...it seems that isNotEmptyString doesn't make sense alone. 

```javascript
if (isString(description) && isNotEmptyString(description)) { return description }`
```

What makes more sense is

```javascript
isNotEmptyString = allPass([isString, complement(isEmptyString)]);
```

tycho01:
`allPass` could be `both` for just two, for `complement(isEmptyString)` just `Boolean` could do, but yeah, whatevers. With composition like that you'll manage to avoid boilerplate, yeah.
[...]

PR #298, opened 2018-01-20
feat: add isEmptyString

guillaumearm:
Ref #40

### DIFF
--- a/src/isEmptyString.js
+++ b/src/isEmptyString.js
@@ -1,0 +1,29 @@
+import { equals } from 'ramda';
+
+
+/**
+ * Checks if input value is an empty `String`.
+ *
+ * @func isEmptyString
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.4.0|v2.4.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {Boolean}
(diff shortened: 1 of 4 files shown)


src/isEmptyString.js
```diff
@@ -0,0 +1,29 @@
...
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {Boolean}
+ * @see {@link RA.isNotEmptyString|isNotEmptyString}
+ * @example
+ *
+ * RA.isEmptyString(''); // => true
+ * RA.isEmptyString('42'); // => false
+ * RA.isEmptyString(new String('42')); // => false
+ * RA.isEmptyString(new String()); // => false
```
char0n: Pls replace with `new String('')`, just to be explicit

test/isEmptyString.js
```diff
@@ -0,0 +1,23 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+
+describe('isEmptyString', function () {
+  it('tests a value to be an empty `String`', function () {
+    eq(RA.isEmptyString(''), true);
+
+    eq(RA.isEmptyString('42'), false);
+    eq(RA.isEmptyString(new String()), false);
```
char0n: Pls replace with `new String('')`, just to be explicit

Issue #96, opened 2017-07-08
introduce comitizen


char0n:
Help contributors with valid conventional changelog format

- make note in CONTRIBUTING.md

Undistraction:
Also worth noting how the PR message itself should be formatted. It's the last commit by default, but we could give some guidance.

guillaumearm:
Woops I make a duplicate #313 .

Because there are more discussions on other issue, I close this one.

PR #301, opened 2018-01-21
chore(scripts): improve prepublishOnly

guillaumearm:
This script is useful before a push.

-----------

Every time I push, I like to check if all is good using : 
```bash
$ npm run clean && npm run lint && npm run test && npm run test:web && npm run build
```

What do you think about a `test:all` npm script ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
     "build:umd": "webpack --config webpack.config.js",
     "nsp:check": "nsp check",
-    "prepublishOnly": "npm run clean && npm run lint && npm run test && npm run build",
+    "prepublishOnly": "npm run clean && npm run lint && npm run test && npm run test:web && npm run build",
     "clean": "rimraf .nyc_output .tmp docs coverage tmp-test-bundle.js dist lib es"
   },
   "readme": "README.md",


char0n:
We already have a script for this. It' called `prepublishOnly`. It is being run before I create a release of before I commit directly to master. Instead of creating new script, modify `prepublishOnly` and add `npm run test:web` before `npm run build`.

Undistraction:
I think adding something to the README would be helpful here so people know to run it before submitting a PR. That would really cut down on the number of linting issues and CI issues in PRs. 

I think I would create a `beforePR` script that just points to `prepublishOnly` and could also print out a reminder regarding the formatting of commit messages.

char0n:
Rebarding commit message format, see https://github.com/char0n/ramda-adjunct/issues/96

Undistraction:
We could even print out a template to the console.

guillaumearm:
And why not a git prepublish hook script using [husky](https://github.com/typicode/husky) ?

char0n:
Let's drop `test:all` scripts. IMHO we don't need it. As @Undistraction pointed out, we can create `beforePR` script which is going to be just alias for `prepublishOnly`.

Regarding precommit hook, I think it is not needed becase we have CI to check every commit in every PR. If you wish to you can configure pre commit hook to run `beforePR` locally in you dev env.

Regarding commit linting, create a separate issue for it pls. We can integration into CI. But as I understand it it can check commits after they are created. Is there some mechanism to recovers from it after we make a mistake except for ammending the commit and force pushing ? Anyway lets continue this conversion into separate ticket.

guillaumearm:
In fact, no need to `beforePR`, I can use the `prepublishOnly` script.

Issue #306, opened 2018-01-23
Add eslint-plugin-ramda


Undistraction:
See: https://github.com/ramda/eslint-plugin-ramda

Adds some nice rules, though possibly a few we will want to disable.

Undistraction:
We are currently blocked due to lack of Code Climate support for the plugin. They whitelist eslint plugins for security reasons and this isn't on the whitelist yet. I have opened an issue on the Code Climate repo [here](https://github.com/codeclimate/codeclimate-eslint/issues/369)

PR #307, opened 2018-01-23
style(eslint): Add eslint-plugin-ramda

Undistraction:
See: https://github.com/ramda/eslint-plugin-ramda which adds a number of Ramda-related rules.

Note that this will currently fail. I've created a PR so you can take a look and decide which rules we want to enable. I'll happily go through and refactor when this is decided to get this PR passing.

Ref #306


char0n:
Let's go for it. Please drop the `package-lock.json`, and ammend the codebase to conform to all the rules which will fix all the 10 pending linting issues. We will disable some rules in future on use-case basis only, if we find some rules to be contraproductive.

One last thing. The configure in example looks different that it looks in this PR.

```js
{
  "name": "my-awesome-project",
  "eslintConfig": {
    "plugins": [
      "ramda"
    ],
    "extends": "plugin:ramda/recommended"
  }
}
```

Undistraction:
We are using a separate `.eslintrc` file so it doesn't make sense to copy the config in the project's README. It's better to keep all our eslint config in one place.

test/liftFN.js
```diff
@@ -1,13 +1,13 @@
+import { sum } from 'ramda';
```
char0n:
```js
import * as RA from 'ramda';
```

test/neither.js
```diff
@@ -1,3 +1,4 @@
+import { T } from 'ramda';
```
char0n:
```js
import * as RA from 'ramda';
```

test/notBoth.js
```diff
@@ -1,3 +1,4 @@
+import { F } from 'ramda';
```
char0n:
```js
import * as RA from 'ramda';
```

char0n: and  use `R.F` to that we don't fonfuse RA and R functions inside tests.

char0n:
Please rebase on top of master and resolve merge conflicts.

Undistraction:
I added a few more rules which are disabled in the default recommended configuration. They all seem sensible and didn't require flag any errors.

Undistraction:
Looks like Code Climate whitelists eslint plugins and this plugin hasn't been whitelisted (yet).

char0n:
That is unfortunate. Can we do something about that ? Adding exception for every line of code isn't a way to go ;]

Undistraction:
I'm looking into it at the moment.

guillaumearm:
Ref: codeclimate/codeclimate-eslint#369

Undistraction:
Unfortunately I think it's a matter of sitting on this until Code Climate add the plugin to their supported list.

char0n:
Ok we will wait. We have an issue where we track it ? Close this PR for
time beeing and thx for the research

Undistraction:
Closing because we are blocked.

PR #308, opened 2018-01-23
feat(notAllPass): Add notAllPass which is the complement of allPass

Undistraction:
Because of the auto-currying I had to use Ramda's implementation of `allPass`, but tweaked to return the correct result and to satisfy eslint.

Ref #234

### DIFF
--- a/src/notAllPass.js
+++ b/src/notAllPass.js
@@ -1,4 +1,4 @@
-import { curry, curryN, reduce, pluck, max } from 'ramda';
+import { complement, compose, allPass } from 'ramda';
 
 
 /**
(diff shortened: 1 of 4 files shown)


src/notAllPass.js
```diff
@@ -0,0 +1,43 @@
...
+ * const gt10 = R.gt(R.__, 10)
+ * const even = (x) => x % 2 === 0;
+ * const f = RA.notAllPass([gt10, even]);
+ *
+ * f(12); //=> false
+ * f(8); //=> true
+ * f(11); //=> true
+ * f(9); //=> true
+ */
+const notAllPass = curry(preds => curryN(reduce(max, 0, pluck('length', preds)), function testPredicates(...args) {
```
char0n: Wouldn't something like this work better ?

```js 
const notAllPass = (preds) => {
  const configuredAllPass = allPass(preds);
  return curryN(configuredAllPass.length, complement(configuredAllPass));
};
```
or 
```js
const notAllPass = pipe(allPass, complement);
```

char0n:
If you managed to finish this in half an hour I'll release if also for v2.4.0

src/notAllPass.js
Undistraction: That's so strange. I started off with `compose(complement, allPass)` but it was failing. Though it was because of the way `compose` was interacting with the currying, but looks like I was imagining things.

PR #341, opened 2018-02-08
test: fix UnhandledPromiseRejectionWarning

### LINKED ISSUE #264: UnhandledPromiseRejectionWarning in tests

guillaumearm:
should catch rejected promises in tests.

closes #264.

### DIFF
--- a/test/isPromise.js
+++ b/test/isPromise.js
@@ -40,8 +40,13 @@ describe('isPromise', function () {
     const hasNativePromise = typeof Promise !== 'undefined' && contains('[native code]', Promise.toString());
 
     specify('should return true', function () {
-      eq(RA.isPromise(Promise.resolve()), hasNativePromise);
-      eq(RA.isPromise(Promise.reject()), hasNativePromise);
+      const resolvedP = Promise.resolve();
+      const rejectedP = Promise.reject();
+
+      eq(RA.isPromise(resolvedP), hasNativePromise);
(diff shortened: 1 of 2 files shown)


test/isPromise.js
```diff
@@ -41,7 +41,7 @@ describe('isPromise', function () {
 
     specify('should return true', function () {
       eq(RA.isPromise(Promise.resolve()), hasNativePromise);
-      eq(RA.isPromise(Promise.reject()), hasNativePromise);
+      eq(RA.isPromise(Promise.reject().catch(() => {})), hasNativePromise);
```
char0n: Thank you for handling this, but this is not a correct fix. When creating a promise with the following signature
```js
Promise.reject().catch(() => {})
```
you are basically creating a resolved promise. The correct approach would be to create rejected promise, test it with `RA.isPromise` and then handle the rejection.

```js
  context('when value is an instance of native Promise', function () {
    const hasNativePromise = typeof Promise !== 'undefined' && contains('[native code]', Promise.toString());

    specify('should return true', function () {
      const resolvedP = Promise.resolve();
      const rejectedP = Promise.reject();

      eq(RA.isPromise(resolvedP), hasNativePromise);
      eq(RA.isPromise(rejectedP), hasNativePromise);

      rejectedP.catch(RA.noop);
    });
  });

guillaumearm: Oh, I did not know that o_Ô.
It's fixed.

Thanks for the review.

PR #344, opened 2018-02-08
fix: RA.defaults not exported

### LINKED ISSUE #338: defaults alias for mergeRight not exported

guillaumearm:
Closes #338

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -435,13 +435,6 @@ declare namespace RamdaAdjunct {
         resetToDefault(defaultOptions: object): (options: object) => object; // alias of mergeRight
 
         /**
-         * Set properties only if they don't exist. Useful for passing defaults.
-         * Alias of {@link http://ramdajs.com/docs/#merge|mergeRight}.
-         */
-        defaults(defaultOptions: object, options: object): object; // alias of mergeRight
-        defaults(defaultOptions: object): (options: object) => object; // alias of mergeRight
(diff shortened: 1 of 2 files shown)


char0n:
After reviewing this, I found out that what are we trying to do here is not correct. `defaults` should be alias of `R.merge` not `RA.mergeRight`. See [here](https://github.com/lodash/lodash/wiki/FP-Guide#installation) for reasoning.

So I suggest to revert back to original state (not export it) and remove it from [here](https://github.com/char0n/ramda-adjunct/blob/master/src/mergeRight.js#L10)

guillaumearm:
do we choose to **remove `RA.defaults`** ?
There are signature types in `src/index.d.ts`

char0n:
Yes, good point, I forgot about that. Remove that pls. Would not consider this breaking change as changing the major version of release because it got lost some time ago from the API and we did not even notice.

guillaumearm:
Done.
but I need to understand something, why don't expose `RA.defaults` as `R.merge` alias ?

char0n:
I guess because it's not part of RA itself. I think adding aliases directly for ramda functions isn't something we want to do. Adding it to direct RA functions makes sense though.

You can alias ramda itself by providing a PR to it or using syntaxes during import phase that allows this.

```js
const { merge: defaults } = require('ramda');
import { merge as defaults } from 'ramda';

PR #346, opened 2018-02-08
feat(isRegExp/isNotRegExp)

### LINKED ISSUE #325: isRegExp / isNotRegExp
Seems like an obvious addition to me. 

```
const t1 = new RegExp(/x/);
const t2 = /x/;
const t3 = 'x';

const isRegExp = is(RegExp);

console.log(`t1: `, isRegExp(t1)) // true
console.log(`t2: `, isRegExp(t2)) // true
console.log(`t3: `, isRegExp(t3)) // false
```

[REPL](http://ramdajs.com/repl/#?const%20t1%20%3D%20new%20RegExp%28%2Fx%2F%29%3B%0Aconst%20t2%20%3D%20%2Fx%2F%3B%0Aconst%20t3%20%3D%20%27x%27%3B%0A%0Aconst%20isRegExp%20%3D%20is%28RegExp%29%3B%0A%0Aconsole.log%28%60t1%3A%20%60%2C%20isRegExp%28t1%29%29%20%2F%2F%20true%0Aconsole.log%28%60t2%3A%20%60%2C%20isRegExp%28t2%29%29%20%2F%2F%20true%0Aconsole.log%28%60t3%3A%20%60%2C%20isRegExp%28t3%29%29%20%2F%2F%20false)

Undistraction:
Add `isRegExp` predicate and its complement `isNotRegExp` to test whether a value is, or is not a RegExp object.

closes #325


char0n:
Before we merge this, I'd like to open the discussion of the proper type checking. Currently this and couple of  other predicate functions that we have are using `R.is` which is fine in most cases (uses instanceOf internally) . But there are cases where this may not be sufficient and [tag checking](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) seems to be the right solution.

Also see this thread for more information: https://stackoverflow.com/questions/4339288/typeof-for-regexp

Affected usecases are e.g. when using window.postMessage and transfer regex object from one window to another it's still regex but our predicate will not satisfy because it may be from another realm. I tested in chrome and transfering the regex works there and they are instances of the same class. But it really depends on the implementation of browser.

Lodash is using tag detection mechanism because of that and it goes even futher by checking if `toStringTag` was not overriden. I wouldn't go that far, but IMHO I think that tag checking is safer then instance checking.

Undistraction:
I was actually going to open an issue regarding our handling of predicates in general. I think this is a perfect situation for generating tests. In effect all we need for each predicate is a list of values that satisfy and a list of values that don't (which we can reverse for the complements). We should then automate these values being fed into the predicates. I think this would give as much cleaner tests and make testing new predicates as simple as deciding on lists of accepted and unaccepted values.

Regarding tag checking, I don't know that approach well enough to call it. Given that we are effectively an extension of Randa (and indeed some of our complements are complementing their predicates), I think we should follow whichever approach they are using.

I would suggest:

1. Open issue on Randa's repo suggesting they switch `is` to use tag checking.
2. Add this PR as is (it covers everything but very edge cases).
3. Consider generation of tests for predicates.
4. If Ramda switches to tag checking, we will then be using tag checking to because we use `is`.

src/isNotRegExp.js
```diff
@@ -0,0 +1,23 @@
...
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.5.0|v2.5.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {Boolean}
+ * @see {@link RA.isRegExp|isRegExp}
+ * @example
+ *
+ * RA.isNotRegExp(1); //=> true
+ * RA.isNotRegExp(new RegExp()); //=> false
```
char0n: Please add
```js
RA.isNotRegExp(/(?:)/); //=> false

File: src/isRegExp.js [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.5.0|v2.5.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {Boolean}
+ * @see {@link RA.isNotRegExp|isNotRegExp}
+ * @example
+ *
+ * RA.isRegExp(new RegExp()); //=> true
```
char0n: Please add
```js
RA.isRegExp(/(?:)/); //=> true
```

char0n:
Please from https://github.com/char0n/ramda-adjunct/pull/346#issuecomment-364167002 create a separate issue so that we can discuss futher in isolation.

src/isNotRegExp.js
char0n: @Undistraction I ment to add them to @examples not tests. Sorry if I wasnt clear

Undistraction:
@char0n That's done.

PR #347, opened 2018-02-08
feat(appendFlipped): add flipped version of append

### LINKED ISSUE #333: appendFlipped
One thing I do a lot is `flip(append)`.  I know there is a `concatRight` so how about an `appendRight`? 

```
const appendRight = flip(append);
```

Useful with transformers too which expect (acc, value).

Undistraction:
Adds a version of Ramda's append with the argument order flipped.

closes #333

### DIFF
--- a/src/appendFlipped.js
+++ b/src/appendFlipped.js
@@ -13,7 +13,7 @@ import { append, flip } from 'ramda';
  * @param {Array} list The list of elements to add a new item to.
  * @param {*} el The element to add to the end of the new list.
  * @return {Array} A new list containing the elements of the old list followed by `el`.
- * @see R.prepend
+ * @see {@link http://ramdajs.com/docs/#append|append}
  * @example
  *
  *      R.append(['write', 'more'], 'tests'); //=> ['write', 'more', 'tests']
--- a/src/appendFlipped.js
+++ b/src/appendFlipped.js
@@ -3,22 +3,22 @@ import { append, flip } from 'ramda';
 
 /**
  * Returns a new list containing the contents of the given list, followed by the given element.
- * Ramda's `append`, but with argument order reversed: list then value.
+ * Like {@link http://ramdajs.com/docs/#append|R.append} but with argument order reversed.
  *
  * @func appendFlipped
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.5.0|v2.5.0}
  * @category List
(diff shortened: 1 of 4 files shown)


char0n:
Will review this tomorrow morning. Thanks

src/appendFlipped.js
```diff
@@ -0,0 +1,25 @@
...
+ * @func appendFlipped
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.5.0|v2.5.0}
+ * @category List
+ * @sig [a] -> a -> [a]
+ * @param {Array} list The list of elements to add a new item to.
+ * @param {*} el The element to add to the end of the new list.
+ * @return {Array} A new list containing the elements of the old list followed by `el`.
+ * @see {@link http://ramdajs.com/docs/#append|append}
+ * @example
```
char0n: Drop the padding.
```js
@example

R.append(['write', 'more'], 'tests'); //=> ['write', 'more', 'tests']
R.append([], 'tests'); //=> ['tests']
R.append(['write', 'more'], ['tests']); //=> ['write', 'more', ['tests']]
```

src/appendFlipped.js
```diff
@@ -0,0 +1,25 @@
+import { append, flip } from 'ramda';
+
+
+/**
+ * Returns a new list containing the contents of the given list, followed by the given element.
+ * Ramda's `append`, but with argument order reversed: list then value.
```
char0n:
```js
Returns a new list containing the contents of the given list, followed by the given element.
Like {@link http://ramdajs.com/docs/#append|R.append} but with arguments order reversed.
```
Would drop the `list then value` part. It's apprarent from the signature and jsdoc @params.

src/index.d.ts
```diff
@@ -790,6 +790,12 @@ declare namespace RamdaAdjunct {
          */
         compact(list: Array<any>): Array<any>;
 
+        /**
+         * Returns a new list containing the contents of the given list, followed by the given
```
char0n:
```ts
Returns a new list containing the contents of the given list, followed by the given element.
Like {@link http://ramdajs.com/docs/#append|R.append} but with arguments order reversed.
```

Undistraction:
@char0n That's done. It's not picking up the added test, but it is there.

char0n:
Looks good. Nice work !

PR #360, opened 2018-02-11
prettier integration POC

### LINKED ISSUE #336: Integrate prettier
https://github.com/prettier/prettier

Undistraction:
This PR adds prettier and integrates it with eslint. It's for discussion so please take a look. 

I have run it on all files in the `src/` and `test/` dirs. 

Prettier's minimal options are [here](https://prettier.io/docs/en/options.html). I would recommend:

1. Drop the  'Print Width' to 80.
2. Lose the semi-colons.
3. Use backticks instead of single quotes.

I have it configured (with VSCode) so it formats on save which I find to be very very useful. It really becomes part of your workflow - you don't need to think about formatting at all. 

I've added it to the `lint` script. It could go into the `lint:fix` script instead, but I think it should happen transparently when linting.

It'll take some tweaking to get eslint happy with what prettier does, but I'll hold off doing that for the moment until we decide it's something we want. 

closes #336


char0n:
Thank you for this POC.

> Drop the 'Print Width' to 80.
I'd go for that. Let see what it will do to our jsdoc comments that are pretty long 

> Lose the semi-colons.
Don't trust [ASI](http://2ality.com/2011/05/semicolon-insertion.html). I'd rather stay explicit with this one

> Use backticks instead of single quotes.
We can switch later if there are reasonable amount of usecases. 

Let's keep it on evolution level, not revolution ;]

Ok couple of things that I have questions about 


 1. ) How does this play with our eslint config ? https://github.com/prettier/prettier/issues/812

 2.) We will loose 2 lines after import statements; but I can live with that (https://github.com/prettier/prettier/issues/1610)

 3.) [This](https://github.com/char0n/ramda-adjunct/pull/360/files#diff-9fcb495f2a6fee14bb56261f25383f11R81) is something that really sucks

 4.) Spaces in functions declarations: https://github.com/prettier/prettier/issues/3847; Is there some options for this ?

Undistraction:
> Don't trust ASI. I'd rather stay explicit with this one

This is part of what prettier does. With semis disabled it will still insert semis for cases where their absence would cause problems. 

> I'd go for that. Let see what it will do to our jsdoc comments that are pretty long

I'll take a look tomorrow

> We can switch later if there are reasonable amount of usecases.

Np

> How does this play with our eslint config

Prettier will win, so its a matter of changing eslint rules to be harmonious with how prettier formats and disabling all eslint autofixing. When we are happy with how prettier is formatting I can look at this. 

> We will loose 2 lines after import statements

Yep. This is not optional

> This is something that really sucks

Not sure exactly what you mean - the wrapping?

> Spaces in functions declarations

Looks like this is going to land shortly: https://github.com/prettier/prettier/pull/3903

char0n:
> This is part of what prettier does. With semis disabled it will still insert semis for cases where their absence would cause problems.

Ok understand, let's keep the semicolons in for now.

> Not sure exactly what you mean - the wrapping?

Wrapping of chained method calls. Seems really unreadable for me. Let's see how will it look at 80 chars.

Undistraction:
Got everything playing nicely locally so please take a look when you get the chance. Prettier now runs prior to eslint and tslint on `npm run lint`. 

Looks like code climate has two issues (but is reporting 19). Not sure why they are showing now, but can't argue. That is one ugly function ;)

Undistraction:
> If everything else is OK I'll create a separate PR for isArrayLike that reduces the use of return and get that in first, then we can merge this one.

@char0n Shall I get this done before this PR gets too out of sync with master?

char0n:
I'll look at this during tomorrow. Meanwhile I will not merge any code related or configuration PRs. Sorry for delay

.eslintrc
```diff
@@ -10,23 +10,20 @@
...
   },
   "extends": [
-    "airbnb-base",
-    "plugin:ramda/recommended"
+    "eslint-config-airbnb-base",
+    "plugin:ramda/recommended",
+    "prettier"
   ],
-  "plugins": ["eslint-plugin-ramda"],
+  "plugins": ["eslint-plugin-ramda",
```
char0n: properly format  the array comprehension pls

package.json
```diff
@@ -57,8 +57,7 @@
   },
```
char0n: please drop the package-lock.json

test/isNotFloat.js
```diff
@@ -3,19 +3,18 @@ import MAX_SAFE_INTEGER from '../src/internal/polyfills/Number.MAX_SAFE_INTEGER'
...
-  it('tests a value for complement of a `float`', function () {
+describe('isNotFloat', function() {
+  it('tests a value for complement of a `float`', function() {
     eq(RA.isNotFloat(0), true);
     eq(RA.isNotFloat(1), true);
     eq(RA.isNotFloat(-100000), true);
     eq(RA.isNotFloat(MAX_SAFE_INTEGER), true);
     eq(RA.isNotFloat(MIN_SAFE_INTEGER), true);
-    eq(RA.isNotFloat(5e+0), true);
+    eq(RA.isNotFloat(5), true);
```
char0n: the scientific notation has been dropped here, but was here by design, can we ignore here ?

test/isNotFloat.js
```diff
@@ -3,19 +3,18 @@ import MAX_SAFE_INTEGER from '../src/internal/polyfills/Number.MAX_SAFE_INTEGER'
...
 
     eq(RA.isNotFloat(0.1), false);
     eq(RA.isNotFloat(Math.PI), false);
-    eq(RA.isNotFloat(5.56789e+0), false);
+    eq(RA.isNotFloat(5.56789), false);
```
char0n: the scientific notation has been dropped here, but was here by design, can we ignore here ?

char0n:
We can fix code climate issues after the merge. There are some conflicts to resolve in this PR. Can you check it out before we merge (by rebasing on top of currenct master) ? I'd like to rebase this instead of merging.

char0n:
Rerurning the ci build and if green, then merging.

Undistraction:
@char0n Those should now be addressed.

PR #375, opened 2018-02-17
feat: add ensureArray

### LINKED ISSUE #371: ensureArray
I just chat with a friend about using the following function :

```js
// a | [a] -> [a]
const ensureArray = R.when(RA.isNotArray, R.of);
```

e.g. : 

```js
ensureArray(42); // => [42]
ensureArray([42]); // => [42]
```

IMO, It is a recurrent pattern in JS, what do you think of adding it in ramda-adjunct?

guillaumearm:
Closes #371

@char0n, @Undistraction : Are you ok with this documentation ?

special thanks to @Poyoman39 for `RA.ensureArray` idea and name.


src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
+import { when, of } from 'ramda';
+import { isNotArray } from './index';
+
+
+/**
+ * Returns an array, even if not, wrap it in a singleton one.
```
Undistraction: If the argument is already an Array, return it, otherwise return a new Array with the argument as its only element.

src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @func ensureArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig a | [a] -> [a]
```
Undistraction: I think this signature should be `* -> [*]`, but not sure.

src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {*|Array.<*>} val the value to wrap when is not an array
```
Undistraction: the value to ensure as Array

src/index.d.ts
```diff
@@ -368,6 +368,11 @@ declare namespace RamdaAdjunct {
          */
         list(...items: any[]): any[];
 
+        /**
+         * Returns an array, even if not, wrap it in a singleton one.
```
Undistraction: As previous comment.

test/ensureArray.js
```diff
@@ -0,0 +1,29 @@
...
+describe('ensureArray', function () {
+  it('tests `Array` values', function () {
+    const emptyArray = [];
+    const filledArray = [42];
+    const arrayInstance = new Array();
+
+    chai.assert.strictEqual(RA.ensureArray(emptyArray), emptyArray);
+    chai.assert.strictEqual(RA.ensureArray(filledArray), filledArray);
+    chai.assert.strictEqual(RA.ensureArray(arrayInstance), arrayInstance);
+    chai.assert.strictEqual(RA.ensureArray(Array.prototype), Array.prototype);
```
Undistraction: Destructure `assert` for consistency with rest of codebase: `import { assert } from 'chai';`

Undistraction: Define `Array.prototype` as a const for consistency.

test/ensureArray.js
```diff
@@ -0,0 +1,29 @@
...
+  });
+
+  it('tests other values', function () {
```
Undistraction: `it('returns non-Array values as the only element of a new Array', function () {`

src/ensureArray.js
guillaumearm: `R.of` signature is `a → [a]`, so I did according to that.

guillaumearm: @char0n, need advice for this.

char0n: The signature is correct. It's called [parametrized types](https://github.com/ramda/ramda/wiki/Type-Signatures#parameterized-types).

char0n:
Please rebase on top of current master. We have prettier integrated already there.

src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
+import { when, of } from 'ramda';
+import { isNotArray } from './index';
+
+/**
+ * If the argument is already an Array, return it,
```
char0n: I'd prefer the following as description. It's basically a compilate of `when` and `of` descriptions.
```
Returns a singleton array containing the value provided.
If value is already an array, it is returned as is.

File: src/index.d.ts [hunk 8]
```diff
@@ -368,6 +368,12 @@ declare namespace RamdaAdjunct {
          */
         list(...items: any[]): any[];
 
+        /**
```
char0n: I'd prefer the following as description. It's basically a compilate of `when` and `of` descriptions.
```
Returns a singleton array containing the value provided.
If value is already an array, it is returned as is.

File: src/ensureArray.js [hunk 9]
```diff
@@ -0,0 +1,23 @@
+import { when, of } from 'ramda';
+import { isNotArray } from './index';
```
char0n: Insert one empty line between absolute and relative imports pls

src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * If the argument is already an Array, return it,
+ * otherwise return a new Array with the argument as its only element.
+ *
+ * @func ensureArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig a | [a] -> [a]
+ * @param {*|Array.<*>} val the value ensure as Array
```
char0n: `Array.<*>` -> `Array`

It's equivalent in jsdoc and it's shorted.

src/ensureArray.js
```diff
@@ -0,0 +1,23 @@
...
+ * @return {Array<*>}
```
char0n: Array.<*> -> Array

test/ensureArray.js
```diff
@@ -0,0 +1,31 @@
+import { assert } from 'chai';
+import * as RA from '../src/index';
```
char0n: insert one empty line between absolute and relative imports. Prettier seems to not modify this.

test/ensureArray.js
```diff
@@ -0,0 +1,31 @@
...
+import eq from './shared/eq';
+
+describe('ensureArray', function() {
+  const emptyArray = [];
+  const filledArray = [42];
+  const arrayInstance = new Array();
+  const arrayPrototype = Array.prototype;
+
+  it('tests `Array` values', function() {
```
char0n: Can we break  this down into separate tests ?
```js
context('when argument is empty array', function() {
    let emptyArray;

    beforeEach(function() {
      emptyArray = [];
    });

    specify('should return empty array', function() {
      assert.lengthOf(RA.ensureArray(emptyArray, 0));
    });

    specify('should be the same array instance', function() {
      assert.strictEqual(RA.ensureArray(emptyArray, emptyArray));
    });
});
```

And doing this for every case so that we know what we are really testing for.

test/ensureArray.js
```diff
@@ -0,0 +1,31 @@
...
+    assert.strictEqual(RA.ensureArray(emptyArray), emptyArray);
+    assert.strictEqual(RA.ensureArray(filledArray), filledArray);
+    assert.strictEqual(RA.ensureArray(arrayInstance), arrayInstance);
+    assert.strictEqual(RA.ensureArray(arrayPrototype), arrayPrototype);
+  });
+
+  it('returns non-Array values as the only element of a new Array', function() {
```
char0n: This is fine

src/ensureArray.js
guillaumearm: why eslint did not see that ?

char0n: May be related to prettier integration. Create a separate issue for this pls.

test/ensureArray.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  context('when argument is filled array', function() {
+    let filledArray;
+
+    beforeEach(function() {
+      filledArray = [42];
+    });
+
+    specify('should return array og length 1', function() {
```
char0n: og -> of

char0n:
One more typo to fixed. When tests pass, do the merging your self. Use `Squash and Merge` button to squash it into single commit and use the following format:
```
feat: add ensureArray (#375)

Closes #371
```
Nicely done!

src/ensureArray.js
Undistraction: TBH it was more the `a | [a]` that looked wrong to me. I could only find examples where it was used on output, not in input, but good to know it's OK. Sorry @guillaumearm

char0n: @Undistraction check [assocPath](http://ramdajs.com/docs/#assocPath)

It's bassically the same if we don't extract the union type outside

```js
[Idx] → a → {a} → {a}
Idx = String | Int
```

```js
[String | Int] → {a} → {a}
```

Issue #379, opened 2018-02-19
Ensure running 'prettier'


guillaumearm:
Because I just run prettier on master and there are some changes, I think we need to discuss on that, there are several solutions.

Ideas : 
- Separate style fixes in a a specific rule (`npm run lint:fix`)
- Run prettier on a precommit hook (staged files only, using husky and lint-staged #314)

@GingerPlusPlus : Can you confirm me your prettier does nothing on master ?

Undistraction:
By far the easiest solution is to have prettier run on save for your editor. I use VSCode and the integration is seamless - I get all the ESLint errors displayed inline and every time I save Prettier formats the code. This is particularly great because I no longer need to think about formatting my code at all. I can type whatever I need straight in, using snippets as needed, and just save to get everything formatted which is a huge timesaver (and ensures I save regularly). 

However this doesn't solve consistency issues with PRs etc. For my own projects I have Husky set up to run effectively `lint & eslint` before every commit and run `lint & eslint & test & build` before every push. 

Imho asking people to remember to do something - especially someone who might be committing to the project for the first time - is always going to cause issues. Linting / formatting of both code and commit messages should be automatic. This is the only way we can ensure consistency.

char0n:
IMHO I think the consistency needs to be enforced on the CI level. I suggest the following:

1.) extract prettier as independent `script`  in `package.json`.
2.) create a new script in ./scripts/prettier-changes-detection.js that will detect if there are unstaged changes on CI (using git internaly; return exit codes 0 or 1)

Workflow:

When somebody pushes commit or create PR, we will run `npm run prettier` and then `node ./scripts/prettier-changes-detection.js`. If script detects something changed the build will fail.


Documentation:

We will also mention prettier in CONTRIBUTING.md guilde line as part of the developer workflow.

What do you think ?

Undistraction:
I absolutely think it should be enforced at CI level, but if we can catch linting errors or formatting errors before then - before they end up in PRs then I'm up for that as well - it is less noise, less frustration and less wasted time for committers having to address small niggles. 

@char0n What is your argument against using Githooks for this - I can see there is a potential argument that it is unexpected, but I think this can be dealt with in the docs. There is also the fact it can be quite opaque if managing Git with an IDE - for example I use Tower for Git as I find great benefit from a good graphical representation of commit history etc (I think very visually) and it is non-obvious what is happening when the hooked scripts are being run.

guillaumearm:
[...]

PR #384, opened 2018-02-19
test(prettier): add scripts/prettier-test.sh

guillaumearm:
- move prettier from `lint` to `lint:fix` npm script
- run `prettier-test.sh` in `lint` npm script

Ref #379, #336

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
   },
   "betterScripts": {
     "docs": "jsdoc -c jsdoc.json && node ./scripts/jsdoc-inject-dist.js",
-    "lint": "prettier test/**.js src/**.js --write && eslint ./ && tslint -c tslint.json src/index.d.ts",
+    "lint": "./scripts/prettier-test.sh && eslint ./ && tslint -c tslint.json src/index.d.ts",
+    "lint:fix": "prettier test/**.js src/**.js --write",
     "changelog": "conventional-changelog -p angular -i ./CHANGELOG.md -s",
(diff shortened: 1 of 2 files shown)


guillaumearm:
Here test failed because `test/omitBy.js` has not been fixed by `prettier`.

@char0n : what do you think ?

GingerPlusPlus:
It'd be better if it only check files that differ from `master`, I don't think CI should fail for a good PR just because there's an issue elsewhere

guillaumearm:
I think CI should never fail on master, so a PR isn't a good PR if build fails.

GingerPlusPlus:
To simplify things, we could run Prettier from within Eslint, one step less to worry about: https://prettier.io/docs/en/eslint.html

package.json
```diff
@@ -62,7 +62,8 @@
   },
   "betterScripts": {
     "docs": "jsdoc -c jsdoc.json && node ./scripts/jsdoc-inject-dist.js",
-    "lint": "prettier test/**.js src/**.js --write && eslint ./ && tslint -c tslint.json src/index.d.ts",
+    "lint": "./scripts/prettier-test.sh && eslint ./ && tslint -c tslint.json src/index.d.ts",
+    "lint:fix": "prettier test/**.js src/**.js --write",
```
GingerPlusPlus: `lint:fix` should probably also run `eslint --fix` after prettier... Either directly, or by using [prettier-eslint](https://github.com/prettier/prettier-eslint)

char0n:
Closing until #387 is resolved.

Issue #387, opened 2018-02-19
Action points


char0n:
Before we move forward we need to deal with some HP issues. 

## omitBy tests

when running prettier on tests/omitBy.js it will change the source code. This also need to be resolved ASAP.

## Linting & prettier workflows

It is not clear how we currently use this. We have `eslint-config-prettier` and `eslint-plugin-prettier` that should integrate with eslint. So we should be OK just having `lint` and `lint:fix` scripts that are running prettier under the hood as an eslint rule implicitly. If I understand this correctly (https://github.com/prettier/eslint-plugin-prettier) than running `lint` should tell us what needs to prettyfied. If it works like that then all our problems are solved. Runnling `lint` shows us linting issues (including AST issues detected by prettier, and doesn't modify the code) and `lint:fix` will fix them (modifies the code). on CI we will still run only `lint` script and it should  tell us if somebody broke some eslint rule or did not properly format code with prettier running `lint:fix`. Having pre/post commit hooks is secondary now, this needs to be addressed ASAP. Also our `lint:fix` is currently defined as "lint:fix": "better-npm-run lint:fix" but bnr doesn't have any script named `lint:fix` so it fails.

There are also some incosistencies between eslint & prettier that need to be addressed.
```
> eslint --print-config .eslintrc | eslint-config-prettier-check

The following rules are unnecessary or might conflict with Prettier:

- comma-dangle
- semi
```

## Communication organization

I am finding it hard to maintaining contexts in all the issue threads that are currently being discussed. If you need decision on something, please summarize all the previous comments of the issue in one comment, include props and cons, describe the consensus or opinion differences and I'll try to decide so that we can move forward on the particular issue and have at least some working POC.

guillaumearm:
# omitBy tests
that's what triggered all these discussions.

I did not see before we have `eslint-plugin-prettier` and `eslint-config-prettier`.

**But now, I found the problem !**
`"prettier/prettier": "error"` is missing from `.eslintrc`file.

now we can remove **semi** and **comma-dangle** from eslint rules.
never mind for the 'omitLastInOneLineBlock' semi option.

So this is fix #379 for me

--------------------

# Linting & prettier workflows

For me, everything is clear now. git hooks is another deal and it's not HP.

char0n:
@guillaumearm "plugin:prettier/recommended" replaces "prettier/prettier": "error" if I'm corrected. PR now contains also fix to conflicting rules.

guillaumearm:
[...]

PR #388, opened 2018-02-19
chore(prettier): integrate prettier with eslint and property configur…

char0n:
Ref #387

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,7 +12,7 @@
   "extends": [
     "eslint-config-airbnb-base",
     "plugin:ramda/recommended",
-    "prettier"
+    "plugin:prettier/recommended"
   ],
   "plugins": ["eslint-plugin-ramda", "prettier"],
   "rules": {
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,9 +17,7 @@
   "plugins": ["eslint-plugin-ramda", "prettier"],
   "rules": {
     "strict": [2, "global"],
-    "semi": ["error", "always", { "omitLastInOneLineBlock": true}],
     "space-before-function-paren": "off",
-    "comma-dangle": ["error", "always-multiline"],
     "no-underscore-dangle": 0,
     "no-confusing-arrow": "off",
     "object-curly-newline": "off",
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,9 +12,9 @@
   "extends": [
     "eslint-config-airbnb-base",
     "plugin:ramda/recommended",
-    "plugin:prettier/recommended"
+    "prettier"
   ],
-  "plugins": ["eslint-plugin-ramda", "prettier"],
+  "plugins": ["eslint-plugin-ramda", "eslint-plugin-prettier", "prettier"],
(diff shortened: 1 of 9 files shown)


char0n:
Codeclime is having issues for sure...

guillaumearm:
revert and try my fix in #387

guillaumearm:
Maybe use the `prettier/prettier` rule instead of `prettier/recommended` preset fix the issue for codeclimate.

guillaumearm:
I just found related issue on codeclimate/codeclimate-eslint#359.
What is the version of eslint on codeclimate platform ?

Undistraction:
@guillaumearm I don't think the version is the problem. Code Climate whitelist eslint plugins for security reasons, so sometimes a plugin will error because they don't allow it. This happened with the Ramda plugin, but I requested they add it and they did so within a few days. I think this is more likely to be the issue.

char0n:
> You need to re-add the prettier preset (in extends) to enable eslint-config-prettier
Should I act on this ? IMHO it makes no difference

Undistraction:
@char0n Yes. Because that is setting `prettier/prettier` to error which is what will fail the `eslint` linting.

Undistraction:
@char0n No.

```
{
  "extends": [
    "plugin:prettier/recommended"
  ]
}
This does three things:

Enables eslint-plugin-prettier.
Sets the prettier/prettier rule to "error".
Extends the eslint-config-prettier configuration.
```

guillaumearm:
@Undistraction : I ask for the version because [eslint#9771](https://github.com/eslint/eslint/issues/9771) and [eslint#9767](https://github.com/eslint/eslint/issues/9771)

@char0n : yes, `plugin:prettier/recommended` turns on `plugin` and `config` (Ref https://prettier.io/docs/en/eslint.html#why-not-both)

Undistraction:
@guillaumearm From our `codeclimate.yaml`:

```
  eslint:
    enabled: true
    channel: "eslint-4"
```

char0n:
@Undistraction, @guillaumearm so the answer to my question is NO, the PR contains proper configuration ? ;]

guillaumearm:
@char0n : you should try, maybe we can bypass the codeclimate issue.

char0n:
Done

char0n:
Codeclimate ishappy now

guillaumearm:
I think this close #379

Undistraction:
@char0n Well done.

char0n:
yes and #387. Now we can build abstractions like hooks and other mechanism. We have a solid base of linting + lint:fixing that we all understand. Thanks a lot for helping on this!

guillaumearm:
Yes, git hooks is another deal, don't worry about that, we will continue to discuss about that, kindly ^^

char0n:
I have to go to sleep. If somebody is still up after tests are complete can you pls Squash and merge this with..
```
chore(eslint): integrate prettier as eslint rule

Closes #387, #379
```

Thanks

Undistraction:
@char0n I'll wait up.

Issue #313, opened 2018-01-23
Introduce commitizen


guillaumearm:
According to what was said in #301, it can be a good idea to use [commitlint](https://github.com/marionebl/commitlint) and [husky](https://github.com/typicode/husky)

guillaumearm:
Same thing as #314, @char0n, you can close this issue if you are ok ;-)

Personally, I use [commitizen](https://github.com/commitizen/cz-cli) to write my commit messages.

char0n:
IMHO I think we should has commitizen as devDependancy and provide a package.json script to run it.

guillaumearm:
Agree.

We can use it with [cz-conventional-changelog](https://github.com/commitizen/cz-conventional-changelog)

and adding this to package.json
```json
{
  "config": {
    "commitizen": {
      "path": "./node_modules/cz-conventional-changelog"
    }
  },
  "scripts": {
    "commit": "git-cz"
  }
}
```

Usage: 
```bash
$ git add .
$ npm run commit
```

char0n:
We use commit lint and husky, git staged and have a CONTRIBUTING.md document that documents creating commit messages. I think that is enought.

PR #402, opened 2018-02-22
WIP: Git hooks

### LINKED ISSUE #381: Git hooks (prepush/precommit)
This issue is a continuation of #379 discussion about git hooks.

Related issues : #301, #313, #314

I think adding a prepush hook is very good idea, I personally run `prepublishOnly` before every of my push.

But for precommit I disagree, it's unpleasant to wait for it, and [lint-staged](https://github.com/okonet/lint-staged) cannot be used because okonet/lint-staged#62

### LINKED ISSUE #386: Use commit-msg hook to validate commit message
While we're thinking about automating some stuff using git hooks, an idea:

Use [`commit-msg` hook](https://git-scm.com/docs/githooks#_commit_msg) to validate commit message according to [Commit Message Guidelines](https://github.com/char0n/ramda-adjunct/blob/master/CONTRIBUTING.md#-commit-message-guidelines). and drop the link if validation fails, and perhaps explain the problem.

Even validating just the first line would be helpful, as "typing" and "scoping" commits is not the standard practice.

GingerPlusPlus:
Resolves #381.
Resolves #386.

All the hooks we discussed, everything I came up with.

Every hook is added in a separate commit, so to not add a hook, just reverting a commit.

I suggest merge-squashing it after we're done, one `chore(package): added git hooks` would be completely fine IMO.

~~I think that push hook unnecessarily builds the project twice, one for testing, one for build. If that's the case, then `bnr build` in `prepush` is just wasting time. (It is there in the first place because of https://github.com/char0n/ramda-adjunct/issues/381#issuecomment-366826618)~~
`npm run build` creates 3 builds. Trying to build just one just to see if it succeeds would probably suffice on pre-push, especially when all of the builds are created on `prepublishOnly`.

~~I had trouble pushing this, I think they were network issues (When I tried to push from the terminal, I had nothing for a while, they it asked me about credentials, and then either said that they were invalid, or said something about http error), but still, I think it's a good idea that you know it, after all, I added `prepush` hook.~~ I'm getting similar errors without the hook... maybe I should just install Linux on my main machine, things wouldn't spontaneously break for no reason.

Haven't put the scripts in `betterScripts` for the simple reason of avoiding running additional program, `bnr`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "node": ">=0.10.3"
   },
   "scripts": {
+    "prepush": "bnr lint && bnr test && bnr build && bnr docs",
     "docs": "better-npm-run docs",
     "lint": "better-npm-run lint",
     "lint:fix": "better-npm-run lint:fix",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "node": ">=0.10.3"
   },
   "scripts": {
+    "precommit": "bash scripts/pre-commit.sh",
     "prepush": "bnr lint && bnr test && bnr build && bnr docs",
     "docs": "better-npm-run docs",
     "lint": "better-npm-run lint",
--- a/package.json
+++ b/package.json
@@ -121,6 +121,7 @@
     "eslint-plugin-ramda": "=2.4.0",
     "fantasy-land": "3.5.0",
     "glob": "=7.1.2",
+    "husky": "=0.14.3",
     "istanbul": "=0.4.5",
     "jsdoc": "=3.5.5",
     "jsverify": "=0.8.3",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
   "scripts": {
     "commitmsg": "commitlint -e $GIT_PARAMS",
     "precommit": "bash scripts/pre-commit.sh",
-    "prepush": "bnr lint && bnr test && bnr build && bnr docs",
     "docs": "better-npm-run docs",
     "lint": "better-npm-run lint",
     "lint:fix": "better-npm-run lint:fix",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "node": ">=0.10.3"
   },
   "scripts": {
+    "commitmsg": "commitlint -e $GIT_PARAMS",
     "precommit": "bash scripts/pre-commit.sh",
     "prepush": "bnr lint && bnr test && bnr build && bnr docs",
     "docs": "better-npm-run docs",
@@ -97,6 +98,8 @@
     "ramda": ">= 0.19.0 <= 0.25.0"
   },
   "devDependencies": {
+    "@commitlint/cli": "=6.1.0",
+    "@commitlint/config-conventional": "=6.1.0",
     "babel-cli": "=6.26.0",
     "babel-core": "=6.26.0",
     "babel-loader": "=7.1.2",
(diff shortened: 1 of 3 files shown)


scripts/pre-commit.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
```
char0n: Can we port this to javascript to make it platform independent ?

package.json
```diff
@@ -42,6 +42,9 @@
     "node": ">=0.10.3"
   },
   "scripts": {
+    "commitmsg": "commitlint -e $GIT_PARAMS",
+    "precommit": "bash scripts/pre-commit.sh",
+    "prepush": "bnr lint && bnr test && bnr build && bnr docs",
```
char0n: imho I think prepush is not necessary. I dont want to wait to be able to push to remote branches. We should create pr and let ci do the work. Another option is to make this optional and disabled by default and document it in readme. Let ppl decide if they wan to use it or not.

scripts/pre-commit.sh
GingerPlusPlus: It is platform independent, it just requires `bash` and `[` :)
I wrote it on Windows with git-bash installed, works for me.

Still, I agree it'd be cool to port to JS, to make it more performant. It matters here, it's `pre-commit`. Bash version spawns git process and node process (eslint) for every staged file, it could be rewitten in Node in such a way that it won't need any extra processes.

Sadly, I don't really have time / willingness to research which library for git is good enough, and how to use it.

We'd need 2 things, obtain list of staged files, and be able to obtain staged version of a file (not the one from working directory).

package.json
GingerPlusPlus: Looks like there's some disagreement within the team about which hooks to use:

> I hate waiting when I commit (https://github.com/char0n/ramda-adjunct/issues/386#issuecomment-366721586, @guillaumearm)

Also, I think that this hook could be optimized anyway:

> `npm run build` creates 3 builds. Trying to build just one just to see if it succeeds would probably suffice on pre-push, especially when all of the builds are created on `prepublishOnly`.

And, I'm not sure if trying to build is needed anyway, `npm test` works somehow even after `npm run clean`. And I can't tell what `bnr docs` actually does.

And, it has an issue:

> It probably runs everything on version from working dir instead of from last commit -- someone could fix the issues and forget to commit them, and push wouldn't stop him. (https://github.com/char0n/ramda-adjunct/issues/381#issuecomment-367982987)

> Another option is to make this optional [...]. Let ppl decide if they wan to use it or not.

Interesting idea. My take on this, we could make the hooks enabled by default and allow to disable each one of them separately by setting an env variable.

GingerPlusPlus: The fact that there is a CI set up is a valid point.

guillaumearm: @char0n : prepush hook is an habit from me, when I have no CI on my project, so it's fine, we can drop it.

@GingerPlusPlus : I'm ready to wait for `commitlint` and `eslint`, don't worry about me. next step is #313.

GingerPlusPlus:
I broke something horribly: `There isn’t anything to compare.
ramda:master and GingerPlusPlus:git-hooks are entirely different commit histories.`, I'm resolving it...

scripts/pre-commit.sh
guillaumearm: I think we can let the bash script for now, and create another issue for replace it by some je code later.

PR #404, opened 2018-02-23
feat: concatAll

### LINKED ISSUE #109: concatAll, concatAllRight
Abstraction for following functionality + right variant.

```javascript
const a = [1];
const b = [2];
const c = [3];

const result = [...a, ...b, ...c];
```
Something like this could work. We can get Identity value by calling R.empty on String or Array types (Monoid)
```javascript
const concatAll = R.converge(R.reduce(R.concat), [R.empty]);
const concatAllRight = R.converge(R.reduce(R.flip(R.concat)), [R.empty]);
```

- [x] concatAll (one test needs this a fix)
- [ ] concatAllRight

GingerPlusPlus:
Resolves #109.

### DIFF
--- a/src/concatAll.js
+++ b/src/concatAll.js
@@ -1,0 +1,33 @@
+import { concat, identical, identity, pipe, when, reduce } from 'ramda';
+
+import stubNull from './stubNull';
+
+const nullSemigroup = { concat: identity };
+
+/**
+ * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
+ * Returns null if empty foldable was passed.
+ *
+ * @func concatAll
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig Semigroup s => Foldable s f => f -> (s | null)
+ * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
+ * @return {Semigroup|null} concatenated semigroups, or null, if empty foldable was passed
+ * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
+ * @see {@link http://ramdajs.com/docs/#concat|concat}
+ * @see {@link http://ramdajs.com/docs/#unnest|unnest}
+ * @see {@link http://ramdajs.com/docs/#join|join}
+ * @example
+ *
+ * concatAll([[1], [2], [3]]); //=> [1, 2, 3]
+ * concatAll(['1', '2', '3']); //=> '123'
+ * concatAll([]); //=> null;
+ */
+const concatAll = pipe(
+  reduce(concat, nullSemigroup),
+  when(identical(nullSemigroup), stubNull)
+);
+
+export default concatAll;
--- a/src/concatAll.js
+++ b/src/concatAll.js
@@ -12,6 +12,8 @@ const nullSemigroup = { concat: identity };
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
  * @category List
+ * @sig [[a]] -> ([a] | undefined)
+ * @sig [String] -> (String | undefined)
  * @sig Semigroup s => Foldable s f => f -> (s | undefined)
  * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
  * @return {Semigroup|null} concatenated semigroups, or undefined, if empty foldable was passed
--- a/src/concatAll.js
+++ b/src/concatAll.js
@@ -2,7 +2,7 @@ import { concat, identical, identity, pipe, when, reduce } from 'ramda';
 
 import stubUndefined from './stubUndefined';
 
-const nullSemigroup = { concat: identity };
+const leftIdentitySemigroup = { concat: identity };
 
 /**
  * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
@@ -28,8 +28,8 @@ const nullSemigroup = { concat: identity };
  * concatAll([]); //=> undefined;
  */
 const concatAll = pipe(
-  reduce(concat, nullSemigroup),
-  when(identical(nullSemigroup), stubUndefined)
+  reduce(concat, leftIdentitySemigroup),
+  when(identical(leftIdentitySemigroup), stubUndefined)
 );
 
 export default concatAll;
--- a/src/concatAll.js
+++ b/src/concatAll.js
@@ -5,20 +5,22 @@ import stubUndefined from './stubUndefined';
 const leftIdentitySemigroup = { concat: identity };
 
 /**
- * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
- * Returns undefined if empty foldable was passed.
+ * Returns the result of concatenating the given lists or strings.
+ * Note: RA.concatAll expects all elements to be of the same type. It will throw an error if you concat an Array with a non-Array value.
+ * Dispatches to the concat method of the preceding element, if present. Can also concatenate multiple elements of a [fantasy-land compatible semigroup](https://github.com/fantasyland/fantasy-land#semigroup).
+ * Returns undefined if empty array was passed.
  *
  * @func concatAll
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
  * @category List
- * @sig [[a]] -> ([a] | undefined)
- * @sig [String] -> (String | undefined)
- * @sig Semigroup s => Foldable s f => f -> (s | undefined)
- * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
- * @return {Semigroup|null} concatenated semigroups, or undefined, if empty foldable was passed
- * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
+ * @sig [[a]] -> [a] | Undefined
+ * @sig [String] -> String | Undefined
+ * @sig Semigroup s => Foldable s f => f -> s | Undefined
+ * @param {Array.<Array|string>} list List containing elements that will be concatenated
+ * @return {Array|string|undefined} Concatenated elements
  * @see {@link http://ramdajs.com/docs/#concat|concat}
+ * @see {@link RA.concatRight|concatRight}
  * @see {@link http://ramdajs.com/docs/#unnest|unnest}
  * @see {@link http://ramdajs.com/docs/#join|join}
  * @example
--- a/src/concatAll.js
+++ b/src/concatAll.js
@@ -1,20 +1,20 @@
 import { concat, identical, identity, pipe, when, reduce } from 'ramda';
 
-import stubNull from './stubNull';
+import stubUndefined from './stubUndefined';
 
 const nullSemigroup = { concat: identity };
 
 /**
  * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
- * Returns null if empty foldable was passed.
+ * Returns undefined if empty foldable was passed.
  *
  * @func concatAll
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
  * @category List
- * @sig Semigroup s => Foldable s f => f -> (s | null)
+ * @sig Semigroup s => Foldable s f => f -> (s | undefined)
  * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
- * @return {Semigroup|null} concatenated semigroups, or null, if empty foldable was passed
+ * @return {Semigroup|null} concatenated semigroups, or undefined, if empty foldable was passed
  * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
  * @see {@link http://ramdajs.com/docs/#concat|concat}
  * @see {@link http://ramdajs.com/docs/#unnest|unnest}
@@ -23,11 +23,11 @@ const nullSemigroup = { concat: identity };
  *
  * concatAll([[1], [2], [3]]); //=> [1, 2, 3]
  * concatAll(['1', '2', '3']); //=> '123'
- * concatAll([]); //=> null;
+ * concatAll([]); //=> undefined;
  */
 const concatAll = pipe(
   reduce(concat, nullSemigroup),
-  when(identical(nullSemigroup), stubNull)
+  when(identical(nullSemigroup), stubUndefined)
 );
 
 export default concatAll;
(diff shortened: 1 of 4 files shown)


src/index.d.ts
```diff
@@ -376,6 +385,12 @@ declare namespace RamdaAdjunct {
          */
         ensureArray<T>(value: T | T[]): T[];
 
+        /**
+         * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
+         * Returns null if empty foldable was passed.
+         */
+        concatAll<S extends Semigroup>(foldable: Foldable<S>): S | null;
```
char0n: @BjornMelgaard can you review this pls ?

src/concatAll.js
```diff
@@ -0,0 +1,34 @@
...
+ * Returns null if empty foldable was passed.
+ *
+ * @func concatAll
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig Semigroup s => Foldable s f => f -> (s | null)
+ * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
+ * @return {Semigroup|null} concatenated semigroups, or null, if empty foldable was passed
+ * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
```
char0n: why defaultTo ?

test/concatAll.js
```diff
@@ -0,0 +1,30 @@
...
+describe('concatAll', function() {
+  it('concatenates arrays', function() {
+    eq(concatAll([[1], [2], [3]]), [1, 2, 3]);
+  });
+
+  it('concatenates strings', function() {
+    eq(concatAll(['1', '2', '3']), '123');
+  });
+
+  it('returns null if empty foldable was passed', function() {
```
char0n: Your tests are failing on some ramda version. Ramdas utilities become fantasy land compatible in some specific version. We have a pattern of future detection in tests.

See here https://github.com/char0n/ramda-adjunct/blob/a3109fc1020c6f4a8686d032cb7d539a230af258/test/notBoth.js#L50

test/concatAll.js
```diff
@@ -0,0 +1,30 @@
...
+    eq(concatAll([]), null);
+  });
+
+  it('throws if non-foldable is passed', function() {
+    assert.throws(() => concatAll(null), TypeError);
```
char0n: I don't see any tests with Semigroups.

Use `RA.Identity` or monet to create couple of tests; see here https://github.com/char0n/ramda-adjunct/blob/a3109fc1020c6f4a8686d032cb7d539a230af258/test/notBoth.js#L52

src/concatAll.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
+ * Returns null if empty foldable was passed.
+ *
+ * @func concatAll
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig Semigroup s => Foldable s f => f -> (s | null)
+ * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
```
char0n: `Foldable<Semigroup>` is not true for all version of ramda we are supporting.
I'd go for the following just like [R.concat](http://ramdajs.com/docs/#concat) do
```js
@sig [a] -> [a] -> [a]
@sig String -> String -> String
```

And add a sentence: `Can also concatenate multiple members of a fantasy-land compatible semigroup.`

src/concatAll.js
```diff
@@ -0,0 +1,34 @@
+import { concat, identical, identity, pipe, when, reduce } from 'ramda';
+
+import stubNull from './stubNull';
+import fl from './fantasy-land/mapping';
+
+const nullSemigroup = { [fl.concat]: identity };
+
+/**
+ * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
```
char0n:
```
Returns the result of concatenating the given lists or strings.
Note: R.concat expects all items to be of the same type. It will throw an error if you concat an Array with a non-Array value.
Dispatches to the concat method of the item, if present. Can also concatenate multiple members of a fantasy-land compatible semigroup.
```

test/concatAll.js
```diff
@@ -0,0 +1,30 @@
+import { assert } from 'chai';
+
+import { concatAll } from '../src';
+import eq from './shared/eq';
+
+describe('concatAll', function() {
+  it('concatenates arrays', function() {
```
char0n: Prefer should notations;  the same applies for other tests
```js
it('should concatenate arrays'

File: src/concatAll.js [hunk 8]
```diff
@@ -0,0 +1,34 @@
+import { concat, identical, identity, pipe, when, reduce } from 'ramda';
+
+import stubNull from './stubNull';
+import fl from './fantasy-land/mapping';
+
+const nullSemigroup = { [fl.concat]: identity };
```
char0n: I guess we'll need to change this to be compatible from ramda 0.19.1 to ramda 0.25.0

src/concatAll.js
GingerPlusPlus: Because it's a way to get rid of the returned `null`.

char0n: Yes but it's just an implementation detail. Consumer of the api should not be conserned with that. Actually I'm more inclined to use `undefined` instead of null. I've used some ramda functions today and noticed the ramda is never returning `null`. Always undefined. 

http://ramdajs.com/docs and search for `| undefined`

src/concatAll.js
GingerPlusPlus: Already done. Changed back to `'concat'`.

test/concatAll.js
GingerPlusPlus: I think I'll test it on [`NonEmptyList`](https://monet.github.io/monet.js/#non-empty-lists)

src/concatAll.js
GingerPlusPlus: Agree with `undefined`.

Well, actually the user of it should be aware of concerned that it'll return null or undefined when it receives empty foldable.

src/concatAll.js
char0n: We should also add info about returning undefined. It's missing in this description. Please do so.

src/concatAll.js
char0n: Yes but link to `defaultTo` using `@see` doesn't do the job. Actually I think that there should only bee two `@see` tags with reference to `R.concat` and `RA.concatRight`. We should also add `@see` to `concatRight` that referece to `RA.concatAll`.

test/concatAll.js
char0n: Try I'm looking forward to see it

src/concatAll.js
GingerPlusPlus: I don't think it's necessary to reference `RA.concatRight`.

GingerPlusPlus: Also, I'm of two minds about whether or not `@see` `R.unnest` and `R.join`, from one perspective, `concatAll` does a similar thing, except it has no sensible return value for empty foldable. From the other, `R.unnest` and `concatAll` have significantly different requirements about their parameters (`Chain` vs `Foldable Semigroup`).

test/concatAll.js
GingerPlusPlus: Done

src/concatAll.js
GingerPlusPlus: Partly done, added the signatures, but didn't remove the previous one.

test/concatAll.js
GingerPlusPlus: Done.

test/concatAll.js
```diff
@@ -0,0 +1,35 @@
...
+
+  it('should throw if non-foldable is passed', function() {
+    assert.throws(() => concatAll(null), TypeError);
+  });
+
+  it('should throw if foldable contains non-semigroups', function() {
+    assert.throws(() => concatAll([1, 2, null, true]), TypeError);
+  });
+
+  it('should throw if foldable contains non-compatible semigroups', function() {
```
char0n: This test is not passing on ramda 0.23.0

test/concatAll.js
```diff
@@ -0,0 +1,35 @@
...
+
+  it('should returns undefined if empty foldable was passed', function() {
+    eq(concatAll([]), undefined);
+  });
+
+  it('should throw if non-foldable is passed', function() {
+    assert.throws(() => concatAll(null), TypeError);
+  });
+
+  it('should throw if foldable contains non-semigroups', function() {
```
char0n: Your test description contains branching logic. We use more proper mechanism for this.
```js
context('when foldable contains non-semigroups', function () {
  specify('should throw error', function () {
    assert.throws(() => concatAll([1, 2, null, true]), TypeError);
  });
});
```

Ref: https://github.com/char0n/ramda-adjunct/blob/a3109fc1020c6f4a8686d032cb7d539a230af258/test/notBoth.js#L41

src/index.d.ts
char0n:
```ts
concatAll<S extends Semigroup>(foldable: Foldable<S>): S | undefined;

File: src/index.d.ts [hunk 11]
```diff
@@ -376,6 +385,12 @@ declare namespace RamdaAdjunct {
          */
         ensureArray<T>(value: T | T[]): T[];
 
+        /**
```
char0n: Don't forget to align this description with the one I proposed in previous comment

test/concatAll.js
```diff
@@ -0,0 +1,35 @@
+import { assert } from 'chai';
+import { NEL, Nil } from 'monet';
+
+import { concatAll } from '../src';
```
char0n: Use following notation in tests pls. 
```js
import * as RA from '../src';
```

Ref: https://github.com/char0n/ramda-adjunct/blob/2f5173a6699dbf090dee56f389f05bf36dbf66c0/test/isBoolean.js#L4

GingerPlusPlus: Will do. Why tho?

src/concatAll.js
```diff
@@ -0,0 +1,35 @@
+import { concat, identical, identity, pipe, when, reduce } from 'ramda';
+
+import stubUndefined from './stubUndefined';
+
+const nullSemigroup = { concat: identity };
```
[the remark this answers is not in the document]
GingerPlusPlus: Done (but not yet pushed, but it's still useful to mark it as done for me)

test/concatAll.js
char0n: It's a convention we use through out our tests to distinguish between R and RA functions.

test/concatAll.js
GingerPlusPlus: Should I remove it then?

char0n: We have to find out why it's not passing and do feature detection to enable/disable test depending on features R provides for us in particular version

GingerPlusPlus: That version just doesn't throw for concating string and array, so I changed the test to try to concat array and string ;)

GingerPlusPlus: Done.

src/concatAll.js
```diff
@@ -0,0 +1,35 @@
...
+ * @func concatAll
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category List
+ * @sig [[a]] -> ([a] | undefined)
+ * @sig [String] -> (String | undefined)
+ * @sig Semigroup s => Foldable s f => f -> (s | undefined)
+ * @param {Foldable<Semigroup>} foldable foldable with semigroups to concatenate
+ * @return {Semigroup|null} concatenated semigroups, or undefined, if empty foldable was passed
+ * @see {@link http://ramdajs.com/docs/#defaultTo|defaultTo}
```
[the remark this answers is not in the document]
GingerPlusPlus: Remove or leave those in?

```js
/*
 * @see {@link http://ramdajs.com/docs/#unnest|unnest}
 * @see {@link http://ramdajs.com/docs/#join|join}
 */
```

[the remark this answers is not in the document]
char0n: You explained these two in prev comments and I guess it make sense to keep them.

src/concatAll.js
```diff
@@ -0,0 +1,35 @@
+import { concat, identical, identity, pipe, when, reduce } from 'ramda';
+
+import stubUndefined from './stubUndefined';
+
+const nullSemigroup = { concat: identity };
+
+/**
+ * Returns the result of concatenating all semigroups (such as arrays or strings) in passed foldable (such as an array).
```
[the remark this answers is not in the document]
GingerPlusPlus: Right.

> Returns undefined if empty array was passed.

Should I mention `R.defaultTo` there somehow? "pipe the result through `R.defaultTo` to replace the undefined with something useful"

The best wording I came up with.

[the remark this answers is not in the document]
char0n: Nope we should not do that. For me it's the same as using `R.find` function with signature `(a → Boolean) → [a] → a | undefined`. They don't tell us there what to do with undefined.

Some functions just returns undefined, as an expected output. I trying to understand your reasoning but I don't see it, I am sorry. Should we also mention `defaultTo` when documenting `R.prop` function or `R.nth` ? I think not.

test/concatAll.js
GingerPlusPlus: Done. It looks so neat now. Thanks, I have no experience in testing.

test/concatAll.js
GingerPlusPlus: Done.

GingerPlusPlus:
Oh ffs, I thought CI would pass it by now

test/concatAll.js
GingerPlusPlus: This fails in 0.21 tho!

I think I'll just remove the test.

GingerPlusPlus: Ramda should be testing that in `R.concat` anyway.

char0n: Just comment the test out, don't remove it. I'll play with that after merge.

GingerPlusPlus:
Finally.

If everything's fine, merge-squash it or something, don't rebase all of the commits on master, that'd be messy.

If you don't want to merge-squash, I can rebase-squash and force-push.

guillaumearm:
@GingerPlusPlus : that's fine, here is "squash and merge" for each PRs.

GingerPlusPlus:
@guillaumearm The thing is, I don't know who does it show as the creator of the merged commit

guillaumearm:
@GingerPlusPlus  : don't worry about that. 
Your `RA.omitBy` commit has been squashed : https://github.com/char0n/ramda-adjunct/commit/c7c9c7e07f938d688c26ad91d1b6a6b3abd3c5cf

PR #405, opened 2018-02-24
refactor: change anyPass to either where appropriate

### LINKED ISSUE #320: Use either instead of anyPass
use R.either instead of R.anyPass for usecases where there are only two predicates

char0n:
Closes #320


BjornMelgaard:
because either should be faster? @char0n

char0n:
@BjornMelgaard more a semantic issue in this case and also theoretically faster.

PR #407, opened 2018-02-24
feat: add dispatch

char0n:
Ref #66

See the tests for usage. I'll add the required docs if we agree this is something we want. I personally have many usecase for this. See #66 for more info. It was proposed originaly by Michael Fogus.

### DIFF
--- a/src/dispatch.js
+++ b/src/dispatch.js
@@ -31,13 +31,15 @@ const getMaxArity = pipe(
 const iteratorFn = curry((args, accumulator, fn) => {
   const result = tryCatch(fn, stubFalse)(...args);
 
-  return isTruthy(result) ? reduced(result) : accumulator.concat(result);
+  return isTruthy(result) ? reduced(result) : accumulator;
 });
 
 const dispatch = functions => {
   const arity = getMaxArity(functions);
 
-  return curryN(arity, (...args) => reduce(iteratorFn(args), [], functions));
+  return curryN(arity, (...args) =>
+    reduce(iteratorFn(args), undefined, functions)
+  );
 };
 
 export default ifElse(isNonEmptyArray, dispatch, stubUndefined);
--- a/src/dispatch.js
+++ b/src/dispatch.js
@@ -5,7 +5,6 @@ import {
   pipe,
   head,
   curryN,
-  defaultTo,
   reduce,
   reduced,
   curry,
@@ -18,13 +17,7 @@ import stubUndefined from './stubUndefined';
 
 const byArity = comparator((a, b) => a.length > b.length);
 
-const getMaxArity = pipe(
-  sort(byArity),
-  head,
-  defaultTo({ length: 1 }),
-  prop('length'),
-  defaultTo(1)
-);
+const getMaxArity = pipe(sort(byArity), head, prop('length'));
 
 const iteratorFn = curry((args, accumulator, fn) => {
   const result = fn(...args);
--- a/src/dispatch.js
+++ b/src/dispatch.js
@@ -9,36 +9,35 @@ import {
   reduce,
   reduced,
   tryCatch,
+  curry,
+  ifElse,
   F as stubFalse,
 } from 'ramda';
 
 import isTruthy from './isTruthy';
-import isEmptyArray from './isEmptyArray';
+import isNonEmptyArray from './isNonEmptyArray';
+import stubUndefined from './stubUndefined';
+
+const byArity = descend(prop('length'));
+
+const getMaxArity = pipe(
+  sort(byArity),
+  head,
+  defaultTo({ length: 1 }),
+  prop('length'),
+  defaultTo(1)
+);
+
+const iteratorFn = curry((args, accumulator, fn) => {
+  const result = tryCatch(fn, stubFalse)(...args);
+
+  return isTruthy(result) ? reduced(result) : accumulator.concat(result);
+});
 
 const dispatch = functions => {
-  if (isEmptyArray(functions)) {
-    return undefined;
-  }
+  const arity = getMaxArity(functions);
 
-  const byArity = descend(prop('length'));
-  const arity = pipe(
-    sort(byArity),
-    head,
-    defaultTo({ length: 1 }),
-    prop('length')
-  )(functions);
-
-  return curryN(arity, (...args) =>
-    reduce(
-      (accumulator, fn) => {
-        const result = tryCatch(fn, stubFalse)(...args);
-
-        return isTruthy(result) ? reduced(result) : accumulator.concat(result);
-      },
-      [],
-      functions
-    )
-  );
+  return curryN(arity, (...args) => reduce(iteratorFn(args), [], functions));
 };
 
-export default dispatch;
+export default ifElse(isNonEmptyArray, dispatch, stubUndefined);
--- a/src/dispatch.js
+++ b/src/dispatch.js
@@ -11,6 +11,42 @@ import {
   ifElse,
 } from 'ramda';
 
+/**
+ * Can be used as a way to compose multiple invokers together to form polymorphic functions,
+ * or functions that exhibit different behaviors based on their argument(s).
+ * Consumes dispatching functions and keep trying to invoke each in turn, until a non-nil value is returned.
+ *
+ * Accepts a list of dispatching functions and returns a new function.
+ * When invoked, this new function is applied to some arguments,
+ * each dispatching function is applied to those same arguments until one of the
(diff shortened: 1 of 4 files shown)


test/dispatch.js
```diff
@@ -0,0 +1,75 @@
...
+
+import * as RA from '../src';
+import eq from './shared/eq';
+
+describe.only('dispatch', function() {
+  let zeroStub;
+  let nullStub;
+  let undefinedStub;
+  let positiveNumberStub;
+
```
char0n: Add test for empty array as input

test/dispatch.js
```diff
@@ -0,0 +1,75 @@
...
+import eq from './shared/eq';
+
+describe('dispatch', function() {
+  let zeroStub;
+  let nullStub;
+  let undefinedStub;
+  let positiveNumberStub;
+
+  beforeEach(function() {
+    zeroStub = sinon.stub().returns(0);
```
char0n: Incorporate stubs into the test

char0n:
`R.anyPass` cannot replace it, works differently

src/dispatch.js
```diff
@@ -0,0 +1,44 @@
...
+
+import isTruthy from './isTruthy';
+import isEmptyArray from './isEmptyArray';
+
+const dispatch = functions => {
+  if (isEmptyArray(functions)) {
+    return undefined;
+  }
+
+  const byArity = descend(prop('length'));
```
char0n: this can be extracted out.

src/dispatch.js
```diff
@@ -0,0 +1,44 @@
...
+  const arity = pipe(
+    sort(byArity),
```
char0n: this can be extracted out

src/dispatch.js
```diff
@@ -0,0 +1,44 @@
...
+    head,
+    defaultTo({ length: 1 }),
+    prop('length')
+  )(functions);
+
+  return curryN(arity, (...args) =>
```
char0n: the iterator function can be extracted too

guillaumearm:
Very nice, So everything is fine for me.

You just need to passing tests and add documentation.

char0n:
Of course I do ;]

Undistraction:
So it's basically switch? Or am I missing something? If so can we alias to `switch` as dispatch is non-intuitive (though I understand why it has that name). 

I don't understand returning the accumulated errors from `reduce`:

```
accumulator.concat(result)
```

How will it handle values that fall through all the supplied functions?

char0n:
> So it's basically switch? Or am I missing something more? If so can we alias to switch as dispatch is non-> intuitive (though I understand why it has that name), or I suppose it is more flexible than switch because > each step is self contained.

Switch is just one of use use-case of this. I'll borrow the explanation from the Michael Foguses book what it really is:

_"Can be used as a way to compose multiple invokers together to form polymorphic functions, or functions that exhibit different behaviors based on their argument(s). To do this, I’ll need a way to take one or more functions and keep trying to invoke each in turn, until a NIL value is returned."_

And IMHO that's why it's called **dispatch**

> I don't understand returning the accumulated errors from reduce:
> accumulator.concat(result)
> How will it handle values that fall through all the supplied functions?

It was nonsensical and unfinished. Check it now.

> Other than that it looks useful to me.
Cool

src/dispatch.js
```diff
@@ -0,0 +1,43 @@
...
+  ifElse,
+} from 'ramda';
+
+import isNotNil from './isNotNil';
+import isNonEmptyArray from './isNonEmptyArray';
+import stubUndefined from './stubUndefined';
+
+const byArity = comparator((a, b) => a.length > b.length);
+
+const getMaxArity = pipe(
```
guillaumearm: I don't get why use `defaultTo` twice here ? 

IMHO, because we use `isNonEmptyArray` above, `getMaxArity` assume that take a non empty array and just be implemented like this :
```js
const getMaxArity = compose(length, head, sort(byArity))
```


---------------
In addition (just for informations), [R.curryN](https://github.com/ramda/ramda/blob/master/source/internal/_curryN.js) can be used with arity 0.

e.g.
```js
R.curryN(0, () => 42)() // => gives 42, no support for ramda placeholder
R.curryN(1, () => 42)() // => gives a Function, supports ramda placeholder
```

char0n: That is true I overlooked this; pushing changes. Thanks

Do you guys still want to brainstorm on the name ?

guillaumearm: I cannot find any better name, `dispatch` is the best we have IMHO.

char0n:
Ok then I'll add typing and docs during the week. Thanks for input

src/dispatch.js
```diff
@@ -0,0 +1,36 @@
...
+import stubUndefined from './stubUndefined';
+
+const byArity = comparator((a, b) => a.length > b.length);
+
+const getMaxArity = pipe(sort(byArity), head, prop('length'));
+
+const iteratorFn = curry((args, accumulator, fn) => {
+  const result = fn(...args);
+
+  return isNotNil(result) ? reduced(result) : accumulator;
```
char0n: Hm still one thing to fix
```js
return isNotNil(result) ? reduced(result) : result;

**Comment by char0n** [comment]
@guillaumearm, @Undistraction can you do a final review ? Everything should be in place now. Thanks

File: src/dispatch.js [hunk 8]
```diff
@@ -11,6 +11,42 @@ import {
   ifElse,
 } from 'ramda';
 
+/**
+ * Can be used as a way to compose multiple invokers together to form polymorphic functions,
+ * or functions that exhibit different behaviors based on their argument(s).
+ * Consumes dispatching functions and keep trying to invoke each in turn, until a non-nil value is returned.
```
Undistraction: Consumes a list of dispatching functions and tries each in turn until a non-nil value is returned.

src/dispatch.js
```diff
@@ -11,6 +11,42 @@ import {
...
+ *
+ * Accepts a list of dispatching functions and returns a new function.
+ * When invoked, this new function is applied to some arguments,
```
Undistraction: When invoked, each dispatching function is applied to all arguments supplied to this new function in sequence, until one of the dispatching functions returns a non-nil value.

src/dispatch.js
```diff
@@ -11,6 +11,42 @@ import {
...
+ * each dispatching function is applied to those same arguments until one of the
+ * dispatching functions returns a non-nil value.
+ *
+ * @func dispatch
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category Function
+ * @sig [((a, b, ...) -> x1), ((a, b, ...) -> x2), ...] -> x1 | x2 | ...
+ * @param {!Array} functions A list of functions
+ * @return {*|undefined} Returns the first not-nil value or undefined if empty list provided
```
Undistraction: Add note about what happens if no function returns a non-nil value.

test/dispatch.js
```diff
@@ -0,0 +1,86 @@
...
+      });
+    });
+  });
+
+  context('when empty array provided as input', function() {
+    specify('should return undefined', function() {
+      eq(RA.dispatch([]), undefined);
+    });
+  });
+
```
Undistraction: Might be worth adding a test for implicit undefined returns (`return;` and no return as opposed to `return undefined;`) given Ramda's different treatment of implicit and explicit `undefined` values in function invocation.

char0n: Ok will do

src/dispatch.js
char0n: Maybe this one is simple to read ?

`Returns the first not-nil value, or undefined if either an empty list is provided or none of the dispatching functions returns a non-nil value.`

src/dispatch.js
char0n: I stole this from converge description http://ramdajs.com/docs/#converge

src/dispatch.js
Undistraction: LGTM

src/dispatch.js
Undistraction: Fair enough. I just find the phrase 'some arguments' to be very vague.

test/dispatch.js
char0n: > Also might be worth testing that undefined is returned when the last function returns a nil value that isn't undefined - i.e. make sure null doesn't end up returned.

This one is already covered by last test

char0n:
@Undistraction made required changes. This one was quite challenging, took me 12 commits to make it right ;]

Issue #395, opened 2018-02-20
RA.lensSlice


GingerPlusPlus:
Python-style pure slice and set slice encapsulated in a lens:

```js
const lensSlice = (bgn, end) => R.lens(
    R.slice(bgn, end),
    (newVal, oldVal) => oldVal.slice(0, bgn).concat(newVal).concat(oldVal.slice(end)))
```

Probably it could be rewritten to be more readable, but it's just proof-of-concept.

Example:

```js
const lens = lensSlice(0, 2)
R.set(x, '1212', 'aasdf') //=> "1212sdf"
```

Python for comparison (used list because strings are immutable, and set slice is mutable):

```python
>>> a = list('aasdf')
>>> a[0:2] = '1212'
>>> a
['1', '2', '1', '2', 's', 'd', 'f']
```

```js
R.set(x, Array.from('1212'), Array.from('aasdf')) //=> ["1", "2", "1", "2", "s", "d", "f"]
```

GingerPlusPlus:
Related: https://github.com/calmm-js/partial.lenses#L-prefix
https://github.com/calmm-js/partial.lenses#L-slice

char0n:
Legitimate; but I would simplify the composition

```js
const lensSlice = (bgn, end) => R.lens(
    R.slice(bgn, end),
    (value, target) =>  R.useWith(
      R.apply(RA.concatAll), 
      [RA.sliceTo(bgn), R.identity, RA.sliceFrom(end)]
    )(oldVal, newVal, oldVal)
```

Ref https://www.linkedin.com/pulse/functional-lenses-javascript-vladim%C3%ADr-gorej/ for naming
Ref [concatAll](https://github.com/char0n/ramda-adjunct/issues/109) need to implemented first to allow this composition

GingerPlusPlus:
We could also define `setSlice`:
```js
const setSlice = R.curry((bgn, end, replacement, target) =>
    RA.concatAll([RA.sliceTo(bgn, target), replacement, RA.sliceFrom(end, target)]))

// Then
const lensSlice = R.curry((bgn, end) => R.lens(R.slice(bgn, end), setSlice(bgn, end)))
// const lensSlice = R.converge(R.lens, [R.slice, setSlice]) // doesn't work
```

[...]

PR #410, opened 2018-02-24
feat: updateSlice

GingerPlusPlus:
I'd ask someone else to write typings. When I tried to do something like (simplified): `<S extends string|any[]>(replacement: S, target: S): S`, when I passed string literals `"a"` and `"s"`, it inferred `S` to be `"a" | "s"` ._.

We could also consider extending the function to work on and return "Slice-able" (which'd extend Semigroup), but there's no such thing in Fantasy Land, hence I didn't go for it.

What should we do in case `fromIndex > toIndex`? I'd say, either do the current thing, let it do whatever it pleases (consistent with `R.slice`), or throw, because it was probably a mistake (consistent with `R.clamp`).

This function wasn't discussed before (it was born in https://github.com/char0n/ramda-adjunct/issues/395#issuecomment-367297854), so feel free to raise any concerns and toss any ideas here.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -765,6 +765,31 @@ declare namespace RamdaAdjunct {
         sliceTo<T>(toIndex: number, list: string|T[]): string|T[];
         sliceTo(toIndex: number): <T>(list: string|T[]) => string|T[];
 
+        updateSlice(fromIndex: number, toIndex: number, replacement: string, list: string): string;
+        updateSlice(fromIndex: number, toIndex: number, replacement: string): (list: string) => string;
+        updateSlice(fromIndex: number, toIndex: number): {
+            (replacement: string): (list: string) => string;
+            (replacement: string, list: string): string;
+        }
+        updateSlice(fromIndex: number): {
+            (toIndex: number): (replacement: string) => (list: string) => string;
+            (toIndex: number): (replacement: string, list: string) => string;
+            (toIndex: number, replacement: string): (list: string) => string;
+            (toIndex: number, replacement: string, list: string): string;
+        }
+        updateSlice<T>(fromIndex: number, toIndex: number, replacement: T[], list: T[]): T[];
+        updateSlice<T>(fromIndex: number, toIndex: number, replacement: T[]): (list: T[]) => T[];
+        updateSlice<T>(fromIndex: number, toIndex: number): {
+            (replacement: T[]): (list: T[]) => T[];
+            (replacement: T[], list: T[]): T[];
+        }
+        updateSlice<T>(fromIndex: number): {
+            (toIndex: number): (replacement: T[]) => (list: T[]) => T[];
+            (toIndex: number): (replacement: T[], list: T[]) => T[];
+            (toIndex: number, replacement: T[]): (list: T[]) => T[];
+            (toIndex: number, replacement: T[], list: T[]): T[];
+        }
+
         /**
          * Returns a partial copy of an array omitting the indexes specified.
          */
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -765,6 +765,14 @@ declare namespace RamdaAdjunct {
         sliceTo<T>(toIndex: number, list: string|T[]): string|T[];
         sliceTo(toIndex: number): <T>(list: string|T[]) => string|T[];
 
+        /**
+         * Returns a new copy of the list or string
+         * with slice from fromIndex (inclusive) to toIndex (exclusive)
+         * replaced by the provided value.
+         *
+         * If toIndex is lower then fromIndex,
+         * then the behavior of the updateSlice is unspecified.
+         */
         updateSlice(fromIndex: number, toIndex: number, replacement: string, list: string): string;
         updateSlice(fromIndex: number, toIndex: number, replacement: string): (list: string) => string;
         updateSlice(fromIndex: number, toIndex: number): {
--- a/src/index.js
+++ b/src/index.js
@@ -101,6 +101,7 @@ export { default as reduceP } from './reduceP';
 export { default as reduceRightP } from './reduceRightP';
 export { default as sliceFrom } from './sliceFrom';
 export { default as sliceTo } from './sliceTo';
+export { default as updateSlice } from './updateSlice';
 export { default as omitIndexes } from './omitIndexes';
 export { default as compact } from './compact';
 export { default as appendFlipped } from './appendFlipped';
--- a/src/updateSlice.js
+++ b/src/updateSlice.js
@@ -9,6 +9,9 @@ import sliceTo from './sliceTo';
  * with slice from fromIndex (inclusive) to toIndex (exclusive)
  * replaced by the provided value.
  *
+ * If toIndex is lower then fromIndex,
+ * then the behavior of the updateSlice is unspecified.
+ *
  * @func updateSlice
  * @memberOf RA
  * @category List
@@ -25,7 +28,6 @@ import sliceTo from './sliceTo';
  * RA.updateSlice(1, 3, 'DDDD', 'ABBCCC'); //=> 'ADDDDCCC'
  * RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')); //=> [ 'A', 'D', 'D', 'D', 'D', 'C', 'C', 'C' ]
  * RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC'); //=> 'ADDDDCCC'
- * RA.updateSlice(3, 1, 'DDDD', 'ABBCCC'); //=> 'ABBDDDDBBCCC'
  */
 const updateSlice = curry((fromIndex, toIndex, replacement, list) =>
   concatAll([sliceTo(fromIndex, list), replacement, sliceFrom(toIndex, list)])
--- a/src/updateSlice.js
+++ b/src/updateSlice.js
@@ -5,31 +5,30 @@ import sliceFrom from './sliceFrom';
 import sliceTo from './sliceTo';
 
 /**
- * Returns target with given slice replaced using passed replacement.
+ * Returns a new copy of the list or string
+ * with slice from fromIndex (inclusive) to toIndex (exclusive)
+ * replaced by the provided value.
  *
  * @func updateSlice
  * @memberOf RA
  * @category List
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
  * @sig Number -> Number -> [a] -> [a] -> [a]
  * @sig Number -> Number -> String -> String -> String
  * @param {number} fromIndex
  * @param {number} toIndex
- * @param {string|any[]} replacement
- * @param {string|any[]} target
- * @returns {string|any[]} target with given slice replaced using passed replacement
+ * @param {string|Array} replacement
+ * @param {string|Array} list
+ * @returns {string|Array} list with given slice replaced using passed replacement
  * @example
  *
- * RA.updateSlice(1, 3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
- * RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')) //=> [ 'A', 'D', 'D', 'D', 'D', 'C', 'C', 'C' ]
- * RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
- * RA.updateSlice(3, 1, 'DDDD', 'ABBCCC') //=> 'ABBDDDDBBCCC'
+ * RA.updateSlice(1, 3, 'DDDD', 'ABBCCC'); //=> 'ADDDDCCC'
+ * RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')); //=> [ 'A', 'D', 'D', 'D', 'D', 'C', 'C', 'C' ]
+ * RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC'); //=> 'ADDDDCCC'
+ * RA.updateSlice(3, 1, 'DDDD', 'ABBCCC'); //=> 'ABBDDDDBBCCC'
  */
-const updateSlice = curry((fromIndex, toIndex, replacement, target) =>
-  concatAll([
-    sliceTo(fromIndex, target),
-    replacement,
-    sliceFrom(toIndex, target),
-  ])
+const updateSlice = curry((fromIndex, toIndex, replacement, list) =>
+  concatAll([sliceTo(fromIndex, list), replacement, sliceFrom(toIndex, list)])
 );
 
 export default updateSlice;
--- a/src/updateSlice.js
+++ b/src/updateSlice.js
@@ -1,0 +1,35 @@
+import { curry } from 'ramda';
+
+import concatAll from './concatAll';
+import sliceFrom from './sliceFrom';
+import sliceTo from './sliceTo';
+
+/**
+ * Returns target with given slice replaced using passed replacement.
+ *
+ * @func updateSlice
+ * @memberOf RA
+ * @category List
+ * @sig Number -> Number -> [a] -> [a] -> [a]
+ * @sig Number -> Number -> String -> String -> String
+ * @param {number} fromIndex
+ * @param {number} toIndex
+ * @param {string|any[]} replacement
+ * @param {string|any[]} target
+ * @returns {string|any[]} target with given slice replaced using passed replacement
+ * @example
+ *
+ * RA.updateSlice(1, 3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
+ * RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')) //=> [ 'A', 'D', 'D', 'D', 'D', 'C', 'C', 'C' ]
+ * RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
+ * RA.updateSlice(3, 1, 'DDDD', 'ABBCCC') //=> 'ABBDDDDBBCCC'
+ */
+const updateSlice = curry((fromIndex, toIndex, replacement, target) =>
+  concatAll([
+    sliceTo(fromIndex, target),
+    replacement,
+    sliceFrom(toIndex, target),
+  ])
+);
+
+export default updateSlice;
--- a/test/updateSlice.js
+++ b/test/updateSlice.js
@@ -13,4 +13,14 @@ describe('updateSlice', function() {
   it('should support negative indexes', function() {
     eq(RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC'), 'ADDDDCCC');
   });
+
+  it('is curried', function() {
+    eq(RA.updateSlice(1, 3, 'DDDD')('ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1, 3)('DDDD', 'ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1, 3)('DDDD')('ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1)(3, 'DDDD', 'ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1)(3, 'DDDD')('ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1)(3)('DDDD', 'ABBCCC'), 'ADDDDCCC');
+    eq(RA.updateSlice(1)(3)('DDDD')('ABBCCC'), 'ADDDDCCC');
+  });
 });
--- a/test/updateSlice.js
+++ b/test/updateSlice.js
@@ -7,10 +7,7 @@ describe('updateSlice', function() {
   });
 
   it('should work on arrays', function() {
-    eq(
-      RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')),
-      Array.from('ADDDDCCC')
-    );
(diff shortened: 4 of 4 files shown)


char0n:
I'll look at this tommorow and provide typings.

test/updateSlice.js
```diff
@@ -0,0 +1,16 @@
+import * as RA from '../src';
+import eq from './shared/eq';
+
+describe('updateSlice', function() {
+  it('should work on strings', function() {
+    eq(RA.updateSlice(1, 3, 'DDDD', 'ABBCCC'), 'ADDDDCCC');
+  });
+
```
char0n: Missing test for cases like this 
```js
RA.updateSlice(3, 1, 'DDDD', 'ABBCCC')
```
when fromIndex is higher number then to index

src/updateSlice.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {number} toIndex
+ * @param {string|any[]} replacement
+ * @param {string|any[]} target
+ * @returns {string|any[]} target with given slice replaced using passed replacement
+ * @example
+ *
+ * RA.updateSlice(1, 3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
+ * RA.updateSlice(1, 3, Array.from('DDDD'), Array.from('ABBCCC')) //=> [ 'A', 'D', 'D', 'D', 'D', 'C', 'C', 'C' ]
+ * RA.updateSlice(-5, -3, 'DDDD', 'ABBCCC') //=> 'ADDDDCCC'
+ * RA.updateSlice(3, 1, 'DDDD', 'ABBCCC') //=> 'ABBDDDDBBCCC'
```
char0n: This doesn't make sense to me. See how [R.slice](http://ramdajs.com/docs/#slice) handles this case. IMHO we should return empty array or empty string in these cases.

See also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice
and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice

src/index.js
```diff
@@ -101,6 +101,7 @@ export { default as reduceP } from './reduceP';
 export { default as reduceRightP } from './reduceRightP';
 export { default as sliceFrom } from './sliceFrom';
 export { default as sliceTo } from './sliceTo';
+export { default as updateSlice } from './updateSlice';
```
char0n: Typings:
```ts
updateSlice(fromIndex: number, toIndex: number, replacement: string|any[], list: string|any[]): string|any[];
updateSlice(fromIndex: number, toIndex: number, replacement: string|any[]) =>  list: string|any[]): string|any[];
updateSlice(fromIndex: number, toIndex: number): {
  (replacement: string|any[], list: string|any[]) => string|any[];
  (replacement: string|any[]):  (list: string|any[]) => string|any[];
}
updateSlice(fromIndex: number): {
  (toIndex: number, replacement: string|any[], list: string|any[]) => string|any[];
  (toIndex: number, replacement: string|any[]): (list: string|any[]) => string|any[];
  (toIndex: number): {
    (replacement: string|any[], list: string|any[]) => string|any[];
    (replacement: string|any[]): (list: string|any[]) => string|any[];
  }
}
```

@BjornMelgaard can you double check pls ?

test/updateSlice.js
GingerPlusPlus: I'm not sure what's the correct behavior in this case.

src/updateSlice.js
GingerPlusPlus: replace empty selection = return empty array? That doesn't seem right.

Python treats it as insertion.
I was about to say that we don't need it, until I noticed that `R.insertAll` works only for arrays.

Default behavior is not very useful, IMO, we'd need to special case it if we want to support it.

test/updateSlice.js
```diff
@@ -0,0 +1,16 @@
+import * as RA from '../src';
+import eq from './shared/eq';
+
+describe('updateSlice', function() {
+  it('should work on strings', function() {
```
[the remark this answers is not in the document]
GingerPlusPlus: I have no idea what you mean

src/index.js
guillaumearm: IMO, it's better using extends
```ts
updateSlice<T extends string|any[]>(fromIndex: number, toIndex: number, replacement: T, list: T): T;
updateSlice<T extends string|any[]>(fromIndex: number, toIndex: number, replacement: T) =>  list: T): T;
updateSlice<T extends string|any[]>(fromIndex: number, toIndex: number): {
  (replacement: T, list: T) => T;
  (replacement: T):  (list: T) => T;
}
updateSlice<T extends string|any[]>(fromIndex: number): {
  (toIndex: number, replacement: T, list: T) => T;
  (toIndex: number, replacement: T): (list: T) => T;
  (toIndex: number): {
    (replacement: T, list: T) => T;
    (replacement: T): (list: T) => T;
  }
}
```

GingerPlusPlus: @guillaumear

> When I tried to do something like (simplified): `<S extends string|any[]>(replacement: S, target: S): S`, when I passed string literals "a" and "s", it inferred S to be "a" | "s" ._.

test/updateSlice.js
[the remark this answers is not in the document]
char0n: Sorry made couple of typos in the comment that made in unreadable.

I mean tests for cases describing this behaviors:
when we do `sliceFrom(fromIndex, list)` then `fromIndex` is inclusive. When we do `sliceTo(toIndex, list)` then `toIndex` is exclusive. You are combining both of these functions so IMHO it make sense to create explicit tests that describe these facts.

src/updateSlice.js
char0n: Upon reflection, I agree that returning empty array or string is confusing and doesn't seem correct. Alternative is to return the original list unmodified or throw. 

@Undistraction, @guillaumearm opinions on this ?

char0n: @Undistraction, @guillaumearm just pinging you, take a look at this when you have time.

guillaumearm: I think we should have same behavior as `RA.inRange` and throw.
IMO, having to handle this case is weird.

Undistraction: It's a tricky one. Current behaviour is definitely wrong. `R.slice` will return an empty list if both ends of the supplied range don't sit within the list, but here that behaviour doesn't make much sense. I would say return the list untouched, but what if there is a partial overlap? In JS doesn't even support negative indexes so shouldn't we should clamp them to zero anyway? As for the top end, if the upper index is over, we replace the whole thing. If the lower index is over we return the list untouched.

char0n: @guillaumearm no you understood it correctly. So we have two sensible approaches. Throw or return original string.

Undistraction: We need to handle:

A. both negative indices
B. 1. negative index, 2. positive within range
C. 1. positive within range, 2. negative index
D. 1. positive within range, 2. positive out of range
E. both positive out of range.

I think different situations require different results.

- We can coerce negative indices to 0 which will work for A (returns original list) and B (replaces from 0 - second index).
- C should throw as this is completely invalid
- D should replace and extend
- E should return the original list

What do you think?

GingerPlusPlus: @Undistraction I wouldn't coerce negative indexes to 0. Negative indexes are valid, look at tests, they count from the end. But, it's a valid concern, `fromIndex` may be greater that `toIndex`, because `toIndex` might be negative.

What's worse, wheter there will be overlap or not depends on length of `list`...

src/index.js
BjornMelgaard: better not use extend [because you cant use `updateSlice(1, 3, 'DDDD', 'ABBCCC') == 'adsf'
`](https://www.typescriptlang.org/play/index.html#src=declare%20function%20updateSlice%3C%20T%20extends%20string%20%7C%20any%5B%5D%20%3E%20(fromIndex%3A%20number%2C%20toIndex%3A%20number%2C%20replacement%3A%20T%2C%20list%3A%20T)%3A%20T%3B%0D%0Adeclare%20function%20updateSlice%3CT%20extends%20string%7Cany%5B%5D%3E(fromIndex%3A%20number%2C%20toIndex%3A%20number%2C%20replacement%3A%20T)%3A%20%20(list%3A%20T)%20%3D%3E%20T%3B%0D%0Adeclare%20function%20updateSlice%3CT%20extends%20string%7Cany%5B%5D%3E(fromIndex%3A%20number%2C%20toIndex%3A%20number)%3A%20%7B%0D%0A%20%20(replacement%3A%20T%2C%20list%3A%20T)%3A%20T%3B%0D%0A%20%20(replacement%3A%20T)%3A%20%20(list%3A%20T)%20%3D%3E%20T%3B%0D%0A%7D%0D%0Adeclare%20function%20updateSlice%3CT%20extends%20string%7Cany%5B%5D%3E(fromIndex%3A%20number)%3A%20%7B%0D%0A%20%20(toIndex%3A%20number%2C%20replacement%3A%20T%2C%20list%3A%20T)%3A%20T%3B%0D%0A%20%20(toIndex%3A%20number%2C%20replacement%3A%20T)%3A%20(list%3A%20T)%20%3D%3E%20T%3B%0D%0A%20%20(toIndex%3A%20number)%3A%20%7B%0D%0A%20%20%20%20(replacement%3A%20T%2C%20list%3A%20T)%3A%20T%3B%0D%0A%20%20%20%20(replacement%3A%20T)%3A%20(list%3A%20T)%20%3D%3E%20T%3B%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0AupdateSlice(1%2C%203%2C%20'DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1%2C%203%2C%20'DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1%2C%203)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1)(3%2C%20'DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1)(3)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1)(3)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1%2C%203)('DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1)(3)('DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0A%2F%2F%20dont%20use%20extend%20because%0D%0AupdateSlice(1%2C%203%2C%20'DDDD'%2C%20'ABBCCC')%20%3D%3D%20'adsf'%0D%0A)

[this works](https://www.typescriptlang.org/play/index.html#src=declare%20function%20updateSlice(fromIndex%3A%20number%2C%20toIndex%3A%20number%2C%20replacement%3A%20string%7Cany%5B%5D%2C%20list%3A%20string%7Cany%5B%5D)%3A%20string%7Cany%5B%5D%3B%0D%0Adeclare%20function%20updateSlice(fromIndex%3A%20number%2C%20toIndex%3A%20number%2C%20replacement%3A%20string%7Cany%5B%5D)%3A%20%20(list%3A%20string%7Cany%5B%5D)%20%3D%3E%20string%7Cany%5B%5D%3B%0D%0Adeclare%20function%20updateSlice(fromIndex%3A%20number%2C%20toIndex%3A%20number)%3A%20%7B%0D%0A%20%20(replacement%3A%20string%7Cany%5B%5D%2C%20list%3A%20string%7Cany%5B%5D)%3A%20string%7Cany%5B%5D%3B%0D%0A%20%20(replacement%3A%20string%7Cany%5B%5D)%3A%20%20(list%3A%20string%7Cany%5B%5D)%20%3D%3E%20string%7Cany%5B%5D%3B%0D%0A%7D%0D%0Adeclare%20function%20updateSlice(fromIndex%3A%20number)%3A%20%7B%0D%0A%20%20(toIndex%3A%20number%2C%20replacement%3A%20string%7Cany%5B%5D%2C%20list%3A%20string%7Cany%5B%5D)%3A%20string%7Cany%5B%5D%3B%0D%0A%20%20(toIndex%3A%20number%2C%20replacement%3A%20string%7Cany%5B%5D)%3A%20(list%3A%20string%7Cany%5B%5D)%20%3D%3E%20string%7Cany%5B%5D%3B%0D%0A%20%20(toIndex%3A%20number)%3A%20%7B%0D%0A%20%20%20%20(replacement%3A%20string%7Cany%5B%5D%2C%20list%3A%20string%7Cany%5B%5D)%3A%20string%7Cany%5B%5D%3B%0D%0A%20%20%20%20(replacement%3A%20string%7Cany%5B%5D)%3A%20(list%3A%20string%7Cany%5B%5D)%20%3D%3E%20string%7Cany%5B%5D%3B%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0AupdateSlice(1%2C%203%2C%20'DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1%2C%203%2C%20'DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1%2C%203)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1)(3%2C%20'DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1)(3)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1)(3)('DDDD'%2C%20'ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0AupdateSlice(1%2C%203)('DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A%0D%0A%0D%0AupdateSlice(1)(3)('DDDD')('ABBCCC')%20%2F%2F%3D%3E%20'ADDDDCCC'%0D%0A)

src/updateSlice.js
Undistraction: >  I wouldn't coerce negative indexes to 0. Negative indexes are valid, look at tests, they count from the end

Is that something we want? If we want consistency (with Ramda's clamp), we should error here. This also applies to `inRange` Do we want to support the idea of negative indexes? At present we don't in `inRange`.

GingerPlusPlus: `inRange` supports negative indexes: `const isSingleDigit inRange(-9, 10)`

Undistraction: `inRange` is a different case because we are just talking about a range of numbers and numbers can be negative. So -x literally means -x. In this function (in the current implementation -x doesn't literally mean -x, it means the index that is -x from the last item in the list. 

Personally I find this behaviour pretty confusing, however given that a negative index wouldn't make sense at all when used on a list or string in any other context, I think adding this isn't so much of a problem - we can trust that a user probably hasn't accidentally used a negative index. 

So now that I understand the negatives - that they are just a mirror of positives, counting from the last item, the list becomes what happens if:

A.1 `fromIndex` is negative and abs(fromIndex) is greater than list length
A.2 `fromIndex` is positive and greater than list length
B.1 `toIndex` is negative and abs(toIndex) is greater than list length
B.1 `toIndex` is positive and  greater than list length

However, this is assuming that negative indexes are always counted from the end of the list. The tests do not make clear what should happen if `fromIndex` is negative and `toIndex` is positive and vice-versa. So what is the correct behaviour is when:

1. `fromIndex` is positive and `toIndex` is negative.
2. `fromIndex` is negative and `toIndex` is positive.

GingerPlusPlus: I'd make a function `absIndex` taking index and array or length, and returning normal index. That allows to just (pseudocode) `if absIndex(from) > absIndex(to) throw`.

Undistraction: The `abs` is just shorthand. Forget the implementation. What should the behaviour be?

GingerPlusPlus: Behavior of what?

test/updateSlice.js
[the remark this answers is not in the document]
GingerPlusPlus: I don't think it's necessary. It's implied by other tests.

src/updateSlice.js
char0n: @GingerPlusPlus behavior of the `updateSlice` util in cases that @Undistraction described.

char0n:
What is that status here ? @GingerPlusPlus do you want to continue with this ? I still don't see any of our comments translated into code changes or typings added. Let us know If we can close this or you wish to continue with the efforts.

GingerPlusPlus:
I was doing other things while waiting till we settle on who to do in case of an overlap. I'll continue. I'll handle what we settled on when I'll have time and I'll feel like coding (whole last week I've been working on a feature in another project that I expected to finish in few hours, it's not done yet)

src/updateSlice.js
```diff
@@ -0,0 +1,35 @@
...
+ * @func updateSlice
+ * @memberOf RA
+ * @category List
+ * @sig Number -> Number -> [a] -> [a] -> [a]
+ * @sig Number -> Number -> String -> String -> String
+ * @param {number} fromIndex
+ * @param {number} toIndex
+ * @param {string|any[]} replacement
+ * @param {string|any[]} target
+ * @returns {string|any[]} target with given slice replaced using passed replacement
```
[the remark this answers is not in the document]
GingerPlusPlus: I don't like this wording... I prefer the current, mine.

test/updateSlice.js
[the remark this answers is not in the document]
char0n: Yes but add them anyway please.

test/updateSlice.js
char0n: Let's say that the behavior is unspecified in this usecase. Document it here https://github.com/char0n/ramda-adjunct/pull/410/files#diff-3d5f0b697b4314c9dbc340c8783967b5R8.

We need to add this sentence.

`If toIndex is lower then fromIndex  then the behavior of the updateSlice is unspecified`

What  about the following usecase ?
```js
RA.updateSlice(3, 3, 'DDDD', 'ABBCCC');
```

src/index.d.ts
```diff
@@ -765,6 +765,31 @@ declare namespace RamdaAdjunct {
         sliceTo<T>(toIndex: number, list: string|T[]): string|T[];
         sliceTo(toIndex: number): <T>(list: string|T[]) => string|T[];
 
+        updateSlice(fromIndex: number, toIndex: number, replacement: string, list: string): string;
```
char0n: Missing description block before the signatures.

test/updateSlice.js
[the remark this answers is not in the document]
char0n: Also add test for currying, tests for negative indexes and so on. Current tests don't even cover the examples in jsdoc in `src/updateSlice.js`.

char0n:
Let's move this forward. This can be a great addition to the library.

test/updateSlice.js
GingerPlusPlus: It clearly means insertion, result should be `'ABBDDDDCCC'`.

test/updateSlice.js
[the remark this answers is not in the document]
GingerPlusPlus: Added test for currying, I'm not sure how the other ones should look.

Undistraction:
I think we should wait until any issues are resolved before merging.

char0n:
@Undistraction well that's what I'm asking. Do we still see any issues here ? The only issue I have with this is tests at this point which I am willing to amend. But cannot do that directly because it's on forked repository.

Undistraction:
Feels like there are a lot of things to resolve with it. The docs aren't correct:

> If toIndex is lower then fromIndex then the behavior of the updateSlice is unspecified

However one of the examples includes two negative indexes with the toIndex lower than fromIndex returning a valid result. 

I don't think the result of calling it with identical from and to indexes is properly defined. If we are saying from is inclusive and to is exclusive I don't know where that leaves:

```
RA.updateSlice(3, 3, 'DDDD', 'ABBCCC');
```

If we are saying the result is `ABBDDDDCCC` then doesn't that mean that it has the same output as:

```
RA.updateSlice(3, 4, 'DDDD', 'ABBCCC');
```

Which doesn't make sense to me.

I also don't see any tests or explanation of what happens in the following scenario:

```
RA.updateSlice(-4, 4, `ABBCCC`)
```

`toIndex` is not lower than `fromIndex` but what should happen here?

char0n:
@Undistraction thank you valid questions and I agree this should not go into master. We've already spent too much time with this and I'm inclined to close it without merging at this point. I was trying to somehow handle this in positive way by testing it myself and resolving all the pending issues myself. But as you pointed out there are so much problems here, that this function needs a specification before an implementation.

It all goes to this [comment](https://github.com/char0n/ramda-adjunct/issues/395#issuecomment-367989672). @GingerPlusPlus I've adviced you to open separate issue for this (@Undistraction did the same) - opening and issue usually means providing a function body and examples to see how it should work. This function doesn't seems to be carefully thought of and it's not our job to specify it for you. So please if you want to continue, open a separate issue for it, and provide the examples of input and output and how you plan to handle the corner cases. And if you're unsure then document the corner cases and invite us to collaborate about how we can resolve them.

I'm closing this, if anybody want to continue with this, we need a separate issue for this, specification and documentation of corner cases with proposed solutions.

Undistraction:
Agree. It's potentially a really useful addition, but it needs to be carefully thought through.

PR #434, opened 2018-03-07
Add isNonPositive and isNonNegative

### LINKED ISSUE #324: RA.isNonNegative, RA.isNonPositive
There are currently `isPositive` and `isNegative`, but we don't offer a way of checking for positive numbers including zero or negative numbers including zero. There is currently no way to check explicity for zero, therefore it isn't possible to do `either(isZero, isPositive)`. 

I suggest adding an `isZero` which would then allow us to add `isPositiveIncludingZero` and `isNegativeIncludingZero`.

Undistraction:
Add predicates for testing numbers for being non-positive (zero and all negative numbers) or non-negative (zero and all positive numbers)

Closes #324

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -770,6 +770,7 @@ declare namespace RamdaAdjunct {
          */
         sliceFrom<T>(fromIndex: number, list: string | T[]): string | T[];
         sliceFrom(fromIndex: number): <T>(list: string | T[]) => string | T[];
+        sliceFrom<T>(fromIndex: number, list: string | T[]): string | T[];
 
         /**
          * Returns the elements of the given list or string (or object with a slice method)
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -251,12 +251,14 @@ declare namespace RamdaAdjunct {
         isNotNumber(val: any): boolean;
 
         /**
-         * Checks if value is a positive `Number` primitive or object.
+         * Checks if value is a positive `Number` primitive or object. Zero is not considered
+         * negative.
          */
         isPositive(val: any): val is number;
 
         /**
-         * Checks if value is a negative `Number` primitive or object.
+         * Checks if value is a negative `Number` primitive or object. Zero is not considered
+         * negative.
          */
         isNegative(val: any): val is number;
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -251,14 +251,14 @@ declare namespace RamdaAdjunct {
         isNotNumber(val: any): boolean;
 
         /**
-         * Checks if value is a positive `Number` primitive or object. Zero is not considered
-         * negative.
+         * Checks if value is a positive `Number` primitive or object. Zero is considered neither
+         * positive or negative.
          */
         isPositive(val: any): val is number;
 
         /**
-         * Checks if value is a negative `Number` primitive or object. Zero is not considered
-         * negative.
+         * Checks if value is a negative `Number` primitive or object. Zero is considered neither
+         * positive or negative.
          */
         isNegative(val: any): val is number;
 
@@ -269,7 +269,7 @@ declare namespace RamdaAdjunct {
         isNonPositive(val: any): val is number;
 
         /**  
-         * Checks if value is a non-positive `Number` primitive or object. This includes all
+         * Checks if value is a non-negative `Number` primitive or object. This includes all
          * positive numbers and zero.
          */
         isNonNegative(val: any): val is number;
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -263,6 +263,18 @@ declare namespace RamdaAdjunct {
         isNegative(val: any): val is number;
 
         /**
+         * Checks if value is a non-positive `Number` primitive or object. This includes all
+         * negative numbers and zero.
+         */
+        isNonPositive(val: any): val is number;
+
+        /**  
+         * Checks if value is a non-positive `Number` primitive or object. This includes all
+         * positive numbers and zero.
+         */
+        isNonNegative(val: any): val is number;
+
+        /**
          * Checks whether the passed value is a finite `Number`.
          */
         isFinite(val: any): boolean;
@@ -428,18 +440,18 @@ declare namespace RamdaAdjunct {
          * will be called with the left value.
          */
         cata<V1, V2, T1, T2>(
-          leftFn: (leftValue: V1) => T1,
-          rightFn: (rightValue: V2) => T2,
-          either: Catamorphism<V1|V2>,
-        ): T1|T2;
+            leftFn: (leftValue: V1) => T1,
+            rightFn: (rightValue: V2) => T2,
+            either: Catamorphism<V1 | V2>,
+        ): T1 | T2;
         cata<V1, V2, T1, T2>(
-          leftFn: (leftValue: V1) => T1,
-          rightFn: (rightValue: V2) => T2):
-          (either: Catamorphism<V1|V2>,
-        ) => T1|T2;
+            leftFn: (leftValue: V1) => T1,
+            rightFn: (rightValue: V2) => T2):
+            (either: Catamorphism<V1 | V2>,
+            ) => T1 | T2;
         cata<V1, V2, T1, T2>(leftFn: (leftValue: V1) => T1): {
-            (rightFn: (rightValue: V2) => T1, either: Catamorphism<V1|V2>): T1|T2;
-            (rightFn: (rightValue: V2) => T1): (either: Catamorphism<V1|V2>) => T1|T2
+            (rightFn: (rightValue: V2) => T1, either: Catamorphism<V1 | V2>): T1 | T2;
+            (rightFn: (rightValue: V2) => T1): (either: Catamorphism<V1 | V2>) => T1 | T2
         };
 
         /**
@@ -577,7 +589,7 @@ declare namespace RamdaAdjunct {
         ): {
                 (acc: TResult): (list: R) => TResult;
                 (acc: TResult, list: R): TResult
-        };
+            };
 
         /**
          * Given an `Iterable`(arrays are `Iterable`), or a promise of an `Iterable`,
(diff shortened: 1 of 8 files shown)


src/index.d.ts
```diff
@@ -251,15 +251,29 @@ declare namespace RamdaAdjunct {
         isNotNumber(val: any): boolean;
 
         /**
-         * Checks if value is a positive `Number` primitive or object.
+         * Checks if value is a positive `Number` primitive or object. Zero is not considered
+         * negative.
```
char0n: Checks if value is a positive `Number` primitive or object. Zero is considered neither positive nor negative.

src/index.d.ts
```diff
@@ -251,15 +251,29 @@ declare namespace RamdaAdjunct {
...
          */
         isPositive(val: any): val is number;
 
         /**
-         * Checks if value is a negative `Number` primitive or object.
+         * Checks if value is a negative `Number` primitive or object. Zero is not considered
```
char0n: Checks if value is a negative `Number` primitive or object. Zero is considered neither positive nor negative.

src/index.d.ts
```diff
@@ -251,15 +251,29 @@ declare namespace RamdaAdjunct {
...
         isNegative(val: any): val is number;
 
+        /**
+         * Checks if value is a non-positive `Number` primitive or object. This includes all
+         * negative numbers and zero.
+         */
+        isNonPositive(val: any): val is number;
+
+        /**  
+         * Checks if value is a non-positive `Number` primitive or object. This includes all
```
char0n: Checks if value is a non-negative `Number` primitive or object. This includes all positive numbers and zero.

src/index.d.ts
```diff
@@ -754,16 +768,16 @@ declare namespace RamdaAdjunct {
          * from fromIndex (inclusive).
          * Dispatches to the slice method of the third argument, if present.
          */
-        sliceFrom<T>(fromIndex: number, list: string|T[]): string|T[];
-        sliceFrom(fromIndex: number): <T>(list: string|T[]) => string|T[];
+        sliceFrom<T>(fromIndex: number, list: string | T[]): string | T[];
```
char0n: Just for info: All these changes in typings are related to new version of tslint ?

src/isNegative.js
```diff
@@ -3,7 +3,7 @@ import { both, gt } from 'ramda';
 import isNumber from './isNumber';
 
 /**
- * Checks if value is a negative `Number` primitive or object.
+ * Checks if value is a negative `Number` primitive or object. Zero is not considered negative.
```
char0n: Checks if value is a negative `Number` primitive or object. Zero is considered neither positive nor negative.

src/isPositive.js
```diff
@@ -3,7 +3,7 @@ import { both, lt } from 'ramda';
 import isNumber from './isNumber';
 
 /**
- * Checks if value is a positive `Number` primitive or object.
+ * Checks if value is a positive `Number` primitive or object. Zero is not considered positive.
```
char0n: Checks if value is a positive `Number` primitive or object. Zero is considered neither positive nor negative.

test/isNonPositive.js
```diff
@@ -0,0 +1,42 @@
+import * as RA from '../src/index';
+import MAX_SAFE_INTEGER from '../src/internal/polyfills/Number.MAX_SAFE_INTEGER';
+import MIN_SAFE_INTEGER from '../src/internal/polyfills/Number.MIN_SAFE_INTEGER';
+import eq from './shared/eq';
+import args from './shared/arguments';
+import Symbol from './shared/Symbol';
+
+describe('isNonPositive', function() {
+  it('tests a value for positive `Number`', function() {
```
char0n: tests a value for non-positive `Number`'

src/isNonNegative.js
```diff
@@ -0,0 +1,30 @@
+import { both, gte, flip } from 'ramda';
+
+import isNumber from './isNumber';
+
+/**
+ * Checks if value is a non-positive `Number` primitive or object. This includes all positive
```
char0n: Checks if value is a non-negative `Number` primitive or object. This includes all positive numbers and zero.

src/isNonNegative.js
```diff
@@ -0,0 +1,30 @@
...
+ * numbers and zero.
+ *
+ * @func isNonNegative
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.15.0|v1.15.0}
```
char0n: `@since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}`

src/isNonNegative.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link RA.isPositive|isPositive}
```
char0n: `@see {@link RA.isPositive|isPositive}, {@link RA.isNonPositive|isNonPositive}`

Also I would add the following into the `isNegative` source:
`@see {@link RA.isNonNegative|isNonNegative}`

src/isNonNegative.js
```diff
@@ -0,0 +1,30 @@
...
+ * RA.isNonNegative(0); // => true
+ * RA.isNonNegative(1); // => true
+ * RA.isNonNegative(Infinity); // => true
+ * RA.isNonNegative(Number.MAX_VALUE); // => true
+ *
+ * RA.isNonNegative(-Infinity); // => false
+ * RA.isNonNegative(Number.MIN_VALUE); // => false
+ * RA.isNonNegative(NaN); // => false
+ */
+const isNonNegative = both(isNumber, flip(gte)(0));
```
char0n: Would this be better for composition ?
```js
const isNonNegative = both(isNumber, complement(isNegative));
```

src/isNonPositive.js
```diff
@@ -0,0 +1,30 @@
...
+
+import isNumber from './isNumber';
+
+/**
+ * Checks if value is a non-positive `Number` primitive or object. This includes all negative
+ * numbers and zero.
+ *
+ * @func isNonPositive
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.15.0|v1.15.0}
```
char0n: `@since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}`

src/isNonPositive.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {Boolean}
+ * @see {@link RA.isNegative|isNegative}
```
char0n: `@see {@link RA.isNegative|isNegative}, {@link RA.isNonNegative|isNonNegative}`

Also I would add the following into the `isPositive` source:
`@see {@link RA.isNonPositive|isNonPositive}`

src/isNonPositive.js
```diff
@@ -0,0 +1,30 @@
...
+ * RA.isNonPositive(0); // => true
+ * RA.isNonPositive(-1); // => true
+ * RA.isNonPositive(-Infinity); // => true
+ * RA.isNonPositive(Number.MIN_VALUE); // => true
+ *
+ * RA.isNonPositive(Infinity); // => false
+ * RA.isNonPositive(Number.MAX_VALUE); // => false
+ * RA.isNonPositive(NaN); // => false
+ */
+const isNonPositive = both(isNumber, flip(lte)(0));
```
char0n: Would this be better for composition ?
```js
const isNonPositive = both(isNumber, complement(isPositive));
```

Undistraction: Does that make sense though given that `isPositive` is also checking for `isNumber`? I was thinking that this might complicate things, but I guess we're rechecking `isNumber` anyway. However this would mean we are checking it twice.

char0n: I always prefer composition even though It may not be the most effective thing how to do things. We can always optimize later. I consider checking `isNumber` in `isPositive` an implementation detail in this case.

Undistraction: OK. I'll buy that for a dollar.

Undistraction: @char0n My original intuition was correct. The problem comes with `NaN`:

```
const isNonPositive = both(isNumber, complement(isPositive));
isNonPositive(NaN) // true
```

`NaN` is neither positive or non-positive, so this should be `false` shouldn't it? Using `complement` we are effectively doing `!both(isNumber, isPositive)`,  so NaN passes `isNumber` but fails `isPositive` giving a `false`, however the complement is `true`.

src/index.d.ts
char0n: Yes could be part of this PR. The question was rather if  the tslint rules changes to create these changes or did you do them manually ?

src/isNonPositive.js
char0n: Yeah, so the question is - should we consider NaN a non-positive number ? IMHO, I think no (but that is debatable). By this assumption we call use your original signature that seems to avoid NaN problem. But I would add note to function description that NaN is not considered non-positive number.

src/index.d.ts
Undistraction: No. I haven't made any changes outside of the immediate scope of the PR.

Undistraction:
@char0n Just to summarise the outstanding changes: 

- Refactor isNonPositive to use `complement` of `isPositive` - Problem with `NaN` (see comment above).
- Refactor isNonNagative to use `complement` of `isPositive` - Problem with `NaN` (see comment above).
- sliceFrom typescript defs - Weren't changed as part of this PR.

char0n:
@Undistraction thanks for summary, fallback to your original composition and let me know when you want final review.

src/isNonPositive.js
Undistraction: I think this is the most practical solution.

Undistraction:
@char0n Should be good to go.

PR #436, opened 2018-03-07
feat: add inRange

### LINKED ISSUE #393: RA.inRange
As discussed in #324

Signature
```
Number -> Number -> Number -> Boolean
```

Possible Implementation
```
const inRange = curry((low, high, v) =>
  compose(apply(useWith(both, [flip(gte), flip(lte)])), sort(gt), list)(
    low,
    high
  )(v)
);
```
[Codepen](https://codepen.io/Pedr/pen/PQRMWL)

Does it need to do anything else?

Undistraction:
Test whether a value falls within an inclusive range.

- I've added an internal `throwError` function
- What else should we test for?
- What should happen if non-numeric values are passed in?
- Is my description OK?
- Is the signature OK (I was going to crib Ramda's [clamp](http://ramdajs.com/docs/#clamp) signature, but it is opaque to me and I don't want to blindly copy/paste). 
- Is my TypeScript signature OK?

Looks like PhantomJS is having issues with `Number.MAX_SAFE_INTEGER` and `Number.MIN_SAFE_INTEGER`. Shall I ditch those tests or try and do some kind of feature detection? Looks like we are on the latest version of PhantomJS and Webkit supports those, but I can't find any related issues regarding Phantom.

Closes #393


src/inRange.js
```diff
@@ -0,0 +1,44 @@
...
+ * @throws {Error} When `low` is greater than `high`
+ * @example
+ *
+ * RA.inRange(0, 5, 3); //=> true
+ * RA.inRange(0, 5, 0); //=> true
+ * RA.inRange(0, 5, 5); //=> true
+ * RA.inRange(0, 5, -1); //=> false
+ * RA.inRange(0, 5, 6); //=> false
+ */
+export default curry((low, high, value) =>
```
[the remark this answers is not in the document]
GingerPlusPlus: https://github.com/char0n/ramda-adjunct/issues/393#issuecomment-368257240

src/index.d.ts
```diff
@@ -685,6 +685,11 @@ declare namespace RamdaAdjunct {
             (value: any): (obj: object) => boolean;
         };
 
+        /** 
+         * Checks if input value falls within the supplied inclusive range.
+         */
+        inRange(low: number, high: number, value: number): boolean;
```
GingerPlusPlus: I'd add curried typings, at least this one:

```ts
inRange(low: number, high: number): (value: number) => boolean;
```

I'd also add typings in tests for `Date`s, they're coerced into ms timestamp and comparison operators work corrently with them.

```ts
inRange(low: Date, high: Date, value: Date): boolean;
inRange(low: Date, high: Date): (value: Date) => boolean;
```

src/inRange.js
```diff
@@ -0,0 +1,44 @@
...
+  gte,
+  lte,
+} from 'ramda';
+
+import throwError from './internal/throwError';
+
+const errorMessage = `low must not be greater than high in inRange(low, high, value)`;
+
+/**
+ * Checks if input value falls within the supplied inclusive range.
```
GingerPlusPlus: Inclusive?

> end is **exclusive** (https://github.com/char0n/ramda-adjunct/issues/393#issuecomment-367093887)

Undistraction: Thanks. Good catch.

Undistraction: Thanks. Good catch.

char0n:
@Undistraction what is that of this ? Can we do the CR or are there still some things that needs to be discussed ?

Undistraction:
@char0n Well my comments at the top still stand. I guess it makes sense for you to take a look and see what you think if you have the time.

src/internal/throwError.js
```diff
@@ -0,0 +1,3 @@
+export default message => {
```
char0n: Not cool with this. I think we talked about it in past. This is a side effect. If we want to introduce the side effect in our functions like `inRange` (which I am for) the side effect (error) should originate in that function and not in some internal `throwError` function. Every stacktract will start at line `2` of `trowError.js`. Been there, done that and falled back to throwing the error from the place where it should be thrown.

Undistraction: OK. Though it would be alright internally but I see your argument.

src/inRange.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+export default curry((low, high, value) =>
+  ifElse(
+    gte,
+    () => {
+      throw new Error(errorMessage);
+    },
+    useWith(both, [flip(gte), flip(lt)])
+  )(low, high)(value)
+);
```
char0n: I'd extract the implementation so that we don't have to compose in every composition.

```js
const inRangeImpl = ifElse(
  gte,
  () => {
    throw new Error('low must not be greater than high in inRange(low, high, value)');
  },
  useWith(both, [flip(gte), flip(lt)])
);

const inRange = curry((low, high, value) => inRangeImpl(low, high)(value));
```

Also I'm looking for some translation of `flip(gte)` into some function that is already in R without composition..isn't it true that `flip(gte)` is equivalent to `lt` ? I'm not sure if the logic is found, is early morning ;]

Undistraction: What do you mean by

> we don't have to compose in every run

Undistraction: > isn't it true that flip(gte) is equivalent to lt

No, but it is equivalent to `lte`, and `flip(lt)` is equivalent to `gt`. I find the order of args really hard to reason about, but it makes sense to avoid the `flip`.

char0n: > What do you mean by

`ifElse` is HOC. In every call to `inRange` the `ifElse` has to always be composed. If you extract it outside, it is composed once, and used in `inRange` function. Does it make sense ?

char0n:
OK I'll check the PR overally once again this evening and we can merge. Nice job on this !

src/inRange.js
Undistraction: Thanks. Yes it makes sense.

src/inRange.js
```diff
@@ -0,0 +1,35 @@
...
+  () => {
+    throw new Error(
+      'low must not be greater than high in inRange(low, high, value)'
+    );
+  },
+  useWith(both, [lte, gt])
+);
+
+/**
+ * Checks if input value falls within the supplied range. The bottom of the range is inclusive, the
```
char0n: Checks if `value` is between `low` and up to, but not including, `high`.

src/inRange.js
```diff
@@ -0,0 +1,35 @@
...
+ * top of the range is exclusive.
+ *
+ * @func inRange
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
```
char0n: 2.7.0

src/inRange.js
```diff
@@ -0,0 +1,35 @@
...
+ * @category Relation
+ * @sig Number -> Number -> Number -> Boolean
+ * @param {Number} low The lowest value within the range
```
char0n: @param {Number} low Start of the range

src/inRange.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Number} high The first value outside the range
```
char0n: The end of the range

src/index.d.ts
```diff
@@ -699,6 +699,12 @@ declare namespace RamdaAdjunct {
             (value: any): (obj: object) => boolean;
         };
 
+        /** 
+         * Checks if input value falls within the supplied range. The bottom of the range is
```
char0n: Checks if `value` is between `low` and up to, but not including, `high`.

src/index.d.ts
```diff
@@ -699,6 +699,12 @@ declare namespace RamdaAdjunct {
...
+         * inclusive, the top of the range is exclusive.
+         */
+        inRange(low: number, high: number, value: number): boolean;
```
char0n: You have to create curried alternatives.

This is how to do that when you have function with 3 arguments: https://github.com/char0n/ramda-adjunct/blob/b9ac42e0a182958f5930218b4a3bde900c7be4ea/src/index.d.ts#L684

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+
+const supportsMinMaxSafeInt = R.all(RA.isNotUndefined, [
+  Number.MAX_SAFE_INTEGER,
+  Number.MIN_SAFE_INTEGER,
+]);
+
+describe('inRange', function() {
+  it('tests whether a value falls within supplied range', function() {
+    const f = RA.inRange(5, 10);
+    eq(f(5), true);
```
char0n: Can we insert empty line here or prettier will remove it ?

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    eq(f(9), true);
+    eq(f(10), false);
+    eq(f(4), false);
+    eq(f(11), false);
+  });
+
+  it('supports negative values', function() {
+    const f = RA.inRange(-10, -5);
```
char0n: again empty line

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+  it('tests whether a value falls within supplied range', function() {
+    const f = RA.inRange(5, 10);
+    eq(f(5), true);
+    eq(f(9), true);
+    eq(f(10), false);
+    eq(f(4), false);
+    eq(f(11), false);
+  });
+
+  it('supports negative values', function() {
```
char0n: 'should support negative values'

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(-10, -5);
+    eq(f(-6), true);
+    eq(f(-5), false);
+    eq(f(-10), true);
+    eq(f(-4), false);
+    eq(f(-11), false);
+  });
+
+  it('supports `Number.POSITIVE_INFINITY`', function() {
```
char0n: should support `Number.POSITIVE_INFINITY`

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(0, Number.POSITIVE_INFINITY);
+    eq(f(Number.MAX_VALUE), true);
+  });
+
+  it('supports `Number.NEGATIVE_INFINITY`', function() {
```
char0n: should support `Number.NEGATIVE_INFINITY`

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(Number.NEGATIVE_INFINITY, 0);
```
char0n: insert empty line

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    eq(f(Number.NEGATIVE_INFINITY), true);
+  });
+
+  it('supports `Number.MIN_VALUE`', function() {
```
char0n: should support `Number.MIN_VALUE`

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(Number.MIN_VALUE, 1);
```
char0n: insert empty line

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    eq(f(Number.MIN_VALUE), true);
+  });
+
+  it('supports `Number.MAX_VALUE`', function() {
```
char0n: should + empy line

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(0, Number.MAX_VALUE);
+    eq(f(Number.MAX_VALUE), false);
+  });
+
+  it('throws if `low` value is greater than `high` value', function() {
```
char0n: should trhow Error if ...

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(10, 5);
```
char0n: empty line

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    chai.assert.throws(function fn() {
```
char0n: Equivalent with following (pick of them)
```js
chai.assert.throws(() => f(5));
chai.assert.throws(f.bind(null, 5));

File: test/inRange.js [hunk 24]
```diff
@@ -0,0 +1,76 @@
...
+      return f(5);
+    }, Error);
+  });
+
+  it('throws if `low` value is same as `high` value', function() {
```
char0n: The same as above

test/inRange.js
```diff
@@ -0,0 +1,76 @@
...
+    const f = RA.inRange(10, 10);
+    chai.assert.throws(function fn() {
+      return f(5);
+    }, Error);
+  });
+
+  if (supportsMinMaxSafeInt) {
```
char0n: Don't need to do that, we have ponyfills
https://github.com/char0n/ramda-adjunct/blob/e3b30bc7456fd4b867182ef442510bfc053b7c6e/src/internal/polyfills/Number.MAX_SAFE_INTEGER.js#L4-L3

https://github.com/char0n/ramda-adjunct/blob/e3b30bc7456fd4b867182ef442510bfc053b7c6e/src/internal/polyfills/Number.MIN_SAFE_INTEGER.js

char0n:
Regarding your questions:

> What else should we test for?

Floats for sure. There is no test for handling floats. Maybe test for `NaN` which can be considered a Number too. Just so that we know what the behavior is.

> What should happen if non-numeric values are passed in?

The behavior is undefined. We don't care. The functional is total and the category for inputs is defined absolutely as `Number`

> Is my description OK?

See my CR comments

> Is the signature OK (I was going to crib Ramda's clamp signature, but it is opaque to me and I don't want to blindly copy/paste).

It's ok

> Is my TypeScript signature OK?

See my CR comments

> Looks like PhantomJS is having issues with Number.MAX_SAFE_INTEGER and Number.MIN_SAFE_INTEGER. Shall I ditch those tests or try and do some kind of feature detection? Looks like we are on the latest version of PhantomJS and Webkit supports those, but I can't find any related issues regarding Phantom.

Handled by our ponyfill

test/inRange.js
Undistraction: Prettier will remove.

test/inRange.js
Undistraction: Prettier will remove.

test/inRange.js
Undistraction: prettier will remove.

src/inRange.js
```diff
@@ -0,0 +1,34 @@
...
+  () => {
+    throw new Error(
+      'low must not be greater than high in inRange(low, high, value)'
+    );
+  },
+  useWith(both, [lte, gt])
+);
+
+/**
+ * Checks if `value` is between `low` and upto but not including `high`.
```
char0n: up to

src/index.d.ts
```diff
@@ -699,6 +699,17 @@ declare namespace RamdaAdjunct {
             (value: any): (obj: object) => boolean;
         };
 
+        /** 
+         * Checks if `value` is between `low` and upto but not including `high`.
```
char0n: up to

test/inRange.js
char0n: What I ment here was not to remove the tests. But to keep the tests but import these ponyfills and test on them. That way you always have this numbers available in any env.

test/inRange.js
```diff
@@ -0,0 +1,81 @@
...
+    const f = RA.inRange(10, 5);
+    chai.assert.throws(() => f(5), Error);
+  });
+
+  it('should throw error if `low` value is same as `high` value', function() {
+    const f = RA.inRange(10, 10);
+    chai.assert.throws(() => f(5), Error);
+  });
+
+  it('supports `Number.MAX_SAFE_INTEGER`', function() {
```
char0n: should support

test/inRange.js
```diff
@@ -0,0 +1,81 @@
...
+    const f = RA.inRange(0, MAX_SAFE_INTEGER);
+    eq(f(MAX_SAFE_INTEGER), false);
+  });
+
+  it('supports `Number.MIN_SAFE_INTEGER`', function() {
```
char0n: should support

test/inRange.js
```diff
@@ -0,0 +1,81 @@
...
+    const f = RA.inRange(MIN_SAFE_INTEGER, 0);
+    eq(f(MIN_SAFE_INTEGER), true);
+  });
+});
```
char0n: Last thing - add test for currying
```js
it('should be curried', function () ...
```

test/inRange.js
```diff
@@ -0,0 +1,86 @@
...
+    const f = RA.inRange(-10, -5);
+    eq(f(-6), true);
+    eq(f(-5), false);
+    eq(f(-10), true);
+    eq(f(-4), false);
+    eq(f(-11), false);
+  });
+
+  it('should support currying', function() {
+    eq(RA.inRange(5)(10)(5), true);
```
char0n: I think to test the currying you have to test the arity combinations (as we do in other tests)
```js
eq(RA.inRange(5)(10)(5), true);
eq(RA.inRange(5, 10)(5), true);
eq(RA.inRange(5, 10, 5), true);
 eq(RA.inRange(5)(10, 5), true);
```
Do this pls and squash and merge. Will be pristine ;] Thanks for you patience

PR #435, opened 2018-03-07
Add isTrue and isFalse predicates

Undistraction:
Adds predicates for testing the Boolean primatives `true` and `false`.

Note: I have taken the decision not to try and support values created using the Boolean function as a constructor - `new Boolean(false)`, as the values returned are not Boolean primitives but are actually objects. I think this is such an edge-case and is potentially such a deep rabbit hole that it isn't worth the added complexity. Very much open to discussion on this though.

Closes #255

PR #437, opened 2018-03-07
feat: add argsPass

### LINKED ISSUE #370: RA.argsPass
Build a predicate that maps multiple args to multiple predicates:

[(* -> Boolean)] -> *... -> Boolean

```
const multiPredicate = predicates => useWith(and, predicates)
const configuredMultiPredicate = multiPredicate([is(Boolean), is(String)])
configuredMultiPredicate(false, 's')
```

[REPL](http://ramdajs.com/repl/?v=0.25.0#?const%20multiPredicate%20%3D%20predicates%20%3D%3E%20useWith%28and%2C%20predicates%29%0Aconst%20configuredMultiPredicate%20%3D%20multiPredicate%28%5Bis%28Boolean%29%2C%20is%28String%29%5D%29%0AconfiguredMultiPredicate%28false%2C%20%27s%27%29)

Undistraction:
Adds function for mapping arguments to predicates and combining those values using a further function.

- What other tests do we need?
- Please check the signature
- Does the description make sense?
- Please check the typescript definitions 

I think the most important decision is what predicate to use when combining the results. We have the choice of either `isTruthy` or `isTrue` (which is about to drop [PR](https://github.com/char0n/ramda-adjunct/pull/435)). When more arguments are supplied than predicates, any remaining args are passed to the combining predicate untouched. If we choose to use `isTrue` this will mean they will always be `false` unless the argument is the Boolean primative `true`. If we choose to use `isTruthy` then truthy args will be coerced to `true`.  This would also mean that other functions can be used instead of predicates and will have whatever they return coerced to a Boolean. 

Closes #370


src/index.d.ts
```diff
@@ -858,6 +858,18 @@ declare namespace RamdaAdjunct {
...
+         * Takes a combining predicate–predicate that combines a list of Boolean values (such as
+         * `any` or `all`)–and a list of predicates and returns a function which will map the
+         * arguments it receives to the list of predicates and returns the result of passing the
+         * boolean values returned from each predicate to the combining predicate. 
+         * 
+         * The function returned is curried to the number of predicates supplied, and if called with
+         * more arguments than predicates, any remaining arguments are passed in to the combining
+         * predicate untouched.
+         */
+        argsPass(combiningPredicate: Function, predicates: Function[]): Function;
```
guillaumearm: add the curried form

src/argsPass.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * The function returned is curried to the number of
+ * predicates supplied, and if called with more arguments than predicates, any remaining arguments
+ * are passed in to the combining predicate untouched.
+ *
+ * @func compact
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.6.0|v2.6.0}
+ * @category Logic
+ * @sig ([*] -> Boolean) -> [(* -> Boolean), …] -> (*…) -> Boolean
```
guillaumearm:
```
((* -> Boolean) -> [*] -> Boolean) -> [(* -> Boolean), ...] -> (*...) -> Boolean
```

src/argsPass.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {Function} combiningPredicate The predicate used to combine the Boolean values returned
+ * from the other predicates
+ * @param {Array} predicates List of predicates
+ * @return {Boolean} Returns the result of the combined result of mapping arguments to predicates
+ * @example
+ *
+ * RA.argsPass(R.all, RA.isArray, RA.isBoolean, RA.isString)([], false, 'abc') //=> true
+ * RA.argsPass(R.all, RA.isArray, RA.isBoolean, RA.isString)([], false, 1) //=> false
```
guillaumearm: Add more examples with R.any and R.none.
IMHO, examples here are very important to understand behavior of this function.

src/argsPass.js
```diff
@@ -0,0 +1,32 @@
+import { useWith, curry, compose } from 'ramda';
+
+import list from './list';
+import isTruthy from './isTruthy';
+
+/**
+ * Takes a combining predicate–predicate that combines a list of Boolean values (such as `any` or
```
guillaumearm: (such as `any`, `all` or `none`)

test/argsPass.js
```diff
@@ -0,0 +1,155 @@
...
+      });
+    });
+
+    context('with no predicate passing', function() {
+      specify('returns true', function() {
+        const f = argsPass(any, [p1False, p2False, p3False]);
+        assert.isFalse(f(1, 2, 3));
+      });
+    });
+  });
```
guillaumearm: please add `with 'none' as the combinator` tests

char0n:
I'll make a CR tommorow. Pls remove `WIP` from PRs that are readdy to be reviewied.

char0n:
Will do this one tommorow. Fixing Webpack configs took me more time than I planned to spend ;]

Undistraction:
Spent more of my life fixing Webpack configs than I care to remember.

char0n:
I've started too look into this.

src/argsPass.js
```diff
@@ -0,0 +1,39 @@
...
+ * @category Logic
+ * @sig ([*] -> Boolean) -> [(* -> Boolean), …] -> (*…) -> Boolean
+ * @sig ((* -> Boolean) -> [*] -> Boolean) -> [(* -> Boolean), ...] -> (*...) -> Boolean
+ * @param {Function} combiningPredicate The predicate used to combine the values returned from the
+ * list of functions
+ * @param {Array} functions List of functions
+ * @return {Boolean} Returns the combined result of mapping arguments to functions
+ * @example
+ *
+ * RA.argsPass(R.all, RA.isArray, RA.isBoolean, RA.isString)([], false, 'abc') //=> true
```
char0n:
```js
RA.argsPass(R.all, [RA.isArray, RA.isBoolean, RA.isString])([], false, 'abc')
```
Accepts list of functions. This relates to other examples also.

src/argsPass.js
```diff
@@ -0,0 +1,39 @@
...
+ * The function returned is curried to the number of functions supplied, and if called with more
+ * arguments than functions, any remaining arguments are passed in to the combining predicate
+ * untouched.
+ *
+ * @func argsPass
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category Logic
+ * @sig ([*] -> Boolean) -> [(* -> Boolean), …] -> (*…) -> Boolean
+ * @sig ((* -> Boolean) -> [*] -> Boolean) -> [(* -> Boolean), ...] -> (*...) -> Boolean
```
char0n: I am not sure about the first signature, the second is legitimate. Can you provide example code for the first one or explain ?

src/index.d.ts
char0n: Yes we're missing currying here.

src/argsPass.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * RA.argsPass(R.all, RA.isArray, RA.isBoolean, RA.isString)([], false, 'abc') //=> true
+ * RA.argsPass(R.all, RA.isArray, RA.isBoolean, RA.isString)([], false, 1) //=> false
+ * RA.argsPass(R.any, RA.isArray, RA.isBoolean, RA.isString)({}, 1, 'abc') //=> true
+ * RA.argsPass(R.any, RA.isArray, RA.isBoolean, RA.isString)({}, 1, false) //=> false
+ * RA.argsPass(R.none, RA.isArray, RA.isBoolean, RA.isString)({}, 1, false) //=> true
+ * RA.argsPass(R.none, RA.isArray, RA.isBoolean, RA.isString)({}, 1, 'abc') //=> false
+ */
+export default curry((combiningPredicate, predicates) =>
+  useWith(compose(combiningPredicate(isTruthy), list), predicates)
```
char0n: Truthy is the right tool for the job.

src/argsPass.js
```diff
@@ -0,0 +1,39 @@
+import { useWith, curry, compose } from 'ramda';
+
+import list from './list';
+import isTruthy from './isTruthy';
+
+/**
+ * Takes a combining predicate and a list of functions and returns a function which will map the
+ * arguments it receives to the list of functions and returns the result of passing the values
```
char0n: Although this is long, it describes what the functions does. So I'm fine with this.

test/argsPass.js
```diff
@@ -0,0 +1,175 @@
...
+    p2True = stub().returns(true);
+    p3False = stub().returns(false);
+    p3True = stub().returns(true);
+    c2True = stub().returns(true);
+    c1True = stub().returns(c2True);
+    c2False = stub().returns(false);
+    c1False = stub().returns(c2False);
+  });
+
+  it('returns a new curried function for each argument supplied', function() {
```
char0n: change all description to `should` convention pls

test/argsPass.js
char0n: Yes, we should test our own code examples.

test/argsPass.js
```diff
@@ -0,0 +1,175 @@
...
+        assert.isTrue(c2True.calledWith([true, true, true]));
+        assert.isTrue(p1True.calledOnce);
+        assert.isTrue(p2True.calledOnce);
+        assert.isTrue(p3True.calledOnce);
+        assert.isTrue(c1True.calledOnce);
+        assert.isTrue(c2True.calledOnce);
+      });
+    });
+
+    context('with one function returning a falsy value', function() {
```
char0n: Missing test for currying

src/index.d.ts
```diff
@@ -884,6 +884,18 @@ declare namespace RamdaAdjunct {
          */
         nonePass(predicates: Function[]): Function;
 
+        /**
+         * Takes a combining predicate–predicate that combines a list of Boolean values (such as
+         * `any` or `all`)–and a list of predicates and returns a function which will map the
```
char0n: replace `–and` with ` and`

src/argsPass.js
char0n: It seems to me that the description here and on typescript signature is not the same. Can we unify them pls ?

test/argsPass.js
char0n: Yes we should separate the scope of the tests. Test should explicitly tell us what it is testing and should test for one thing. Currying in this test is implicit. So please extract to separate `it/context/specify`.

Undistraction: @char0n Sorry. I'm confused. There is already an (explicit)  test for currying on line 32:
```
it('should return a new curried function for each argument supplied', function() {
```
The test linked above is not using currying and is testing for something else. 

I'm happy to add more tests, but I'm unclear of what exactly you'd like me to test for given that currying is already covered.

char0n: Right. Sorry read the code in hurry. PR is fine. Go for a merge.

PR #439, opened 2018-03-07
style: enable eslint rule for ramda both vs either

### LINKED ISSUE #438: Enable prefer-both-either ramda linting rule
`prefer-both-either` has been [added](https://github.com/ramda/eslint-plugin-ramda/commit/6e5879be4c24a903ab57587af14f90e2746f1212) to eslint-plugin-ramda and should be enabled.

Undistraction:
Enables eslint-plugin-ramda linting rule to flag use of `RA.any` with only two args for replacement with `RA.either`.

Closes #438


Undistraction:
Looks like Code Climate haven't updated to the latest version of `eslint-plugin-ramda`.

char0n:
Let's keep this open then and w8 for them to update. Nice job

Undistraction:
Requested update [here](https://github.com/codeclimate/codeclimate/issues/819)

guillaumearm:
Maybe a dumb question but why do we enable eslint in codeclimate while we have a circleci job for that ? what is the point ?

Undistraction:
That is a very good question.

guillaumearm:
IMHO, we can just disable eslint in codeclimate and move on.

@char0n, any suggestions ?

char0n:
If it's possible, let's do that. We need codeclimate to expose complexity problems and things like that.

guillaumearm:
Yes I think it's possible.

@Undistraction : try to modifiy `.codeclimate.yml` like this : 

```yml
engines:
  eslint:
    enabled: false
```

PR #464, opened 2018-03-23
ci(codeclimate): disable eslint

Undistraction:
We are already running eslint via CircleCI so no need to run in Code Climate as well.


Undistraction:
Just making sure it works

Issue #456, opened 2018-03-17
import RA default export doest not works


guillaumearm:
default export seems to not works like `ramda` does.

```js
import RA from 'ramda-adjunct' // RA is undefined
import R from 'ramda' // R is OK

import * as RA from 'ramda-adjunct' // RA is OK
import * as R from 'ramda' // R is OK
```

I don't understand why, maybe need to investigate https://github.com/ramda/ramda/blob/master/.babelrc

guillaumearm:
I have made some investigations on ramda : 

```js
import R from 'ramda' // use "main" field (commonjs)
import * as R from 'ramda' // use "main" field (commonjs)
```

It seems like Ramda have a bug and never really expose es module.
In my case, "module" field is ignored, but "jsnext:main" works fine.

```js
import RA from 'ramda-adjunct' // use "jsnext:main" field (es)
import * as RA from 'ramda-adjunct' // use "jsnext:main" field (es)
```

I can't find a case where "module" is used, I test this with babel 6 and this configuration : 


```json
{
  "ignore": [
    "node_modules/**"
  ],
  "presets": [
    ["es2015"],
    ["stage-0"]
  ]
}

```

Andarist:
How do u consume `ramda-adjunct`? node? webpack? rollup? other?

guillaumearm:
In my case I've just node, so `import` are replaces by `require`, commonJs is used, everything seems normal.

> You are not supposed to be able to import `default` from RA, it's a namespace library and import * as RA is correct for this use case.

You right, but I have to know why `import R from 'ramda'` is working while `import RA from 'ramda-adjunct'` doesn't

Andarist:
Might be because `ramda` uses custom transform from esm to cjs - https://github.com/ramda/ramda/blob/9e5107e6e021b35179401eb719283cd835687d51/scripts/transpile/esm-to-cjs.js . This is just a guess though - you should be able to debug this easily by inspecting transformed outputs of both.

Andarist:
@quangv You are not supposed to be able to import default from RA, it's a namespace library and `import * as RA` is correct for this use case.

char0n:
Yes the current behavior `* as RA` is due to the tree shaking. We can add additional mechanism that supports `import RA from 'ramda-adjunct'` but now we don't support. I'll fix the documentation to reflect that.

Andarist:
@guillaumearm `ramda` does not export default, so you are already consistent 😉

[...]

PR #506, opened 2018-04-22
docs(README.md): fix RA E56 imports

char0n:
Ref #456


Undistraction:
@char0n Shall I go ahead and merge?

Issue #392, opened 2018-02-20
RA.contained/RA.notContained


guillaumearm:
Also known as `contained` or `within`, I prefer `oneOf` anyway.

```js
const oneOf = R.flip(R.contains)
```

Ref #38

Undistraction:
I like it, but don't understand why the name `oneOf`.

guillaumearm:
For me it's very more clear.

e.g.

```js
R.reject(RA.oneOf([1, 2, 3, 4]));
```

```js
if (RA.oneOf([1, 2, 3, 4]), n) {
  console.log('n is one of them');
}
```

Sometimes, we had to choose semantic over consistance, what do you think ?

Anyway, `within`, `isContained`, `isContainedBy` and `containedBy` are fine too.

char0n:
I vote for `within`, `isWithin`, or fallback to `containesFlipped`

`isWithin` evokes predicate and `containesFlipped` is a last resort. So for me `within` is a clear winner.


`oneOf` evokes a functiona that works similiar to [lodash.sample](https://lodash.com/docs/4.17.5#sample)

guillaumearm:
I don't like `containsFlipped`, I prefer use `flip(contains)` directly.

But `within` or `isWithin` is fine for me ;-)

char0n:
Ok votes pls

`+1 for within`

Undistraction:
I agree that `containsFlipped` is the worst option, but prefer `isContained` or `contained` to `isWithin` or `within` because of the link to `contains` and the fact it is a clear opposite.

`+1 for contained`

guillaumearm:
@Undistraction : you coninced me.

`+1 for contained`

char0n:
it's `contained` then

GingerPlusPlus:
I'd alias it as `within` tho, dunno, it just sounds better to me in cases like `RA.within([1, 2, 3, 4])`.

Or to `oneOf`. Even if some of you don't get it, I do. It's not about array includes, it's more of `value === 1 || value === 2 || value === 3 || value === 4`.

guillaumearm:
Yup, that's why I prefer `oneOf`.
For me it's more like a `anyPass` but with values.

Undistraction:
@gingerplusplus It`s not about not getting it. It's about not agreeing.

[...]

PR #509, opened 2018-04-22
fix(appendFlipped): add curried typescript signature

guillaumearm:
By doing #392, I noticed that `RA.appendFlipped` typescript signature was missing.


char0n:
I reworded the commit message type to `refactor`. IMHO this is not a fix, neither a feature.

guillaumearm:
IMHO, this is a fix for typescript users.

PR #510, opened 2018-04-22
feat: add contained

guillaumearm:
Ref #392

### DIFF
--- a/src/contained.js
+++ b/src/contained.js
@@ -9,12 +9,12 @@ import { flip, contains } from 'ramda';
  *
  * @func contained
  * @memberOf RA
- * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
  * @category List
- * @sig [a] → a → Boolean
+ * @sig [a] -> a -> Boolean
  * @param {Array|String} list The list to consider
  * @param {*} a The item to compare against
- * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise.
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise
  * @see {@link http://ramdajs.com/docs/#contains|R.contains}
  * @example
  *
--- a/src/contained.js
+++ b/src/contained.js
@@ -1,0 +1,27 @@
+import { flip, contains } from 'ramda';
+
+/**
+ * Returns true if the specified value is equal, in R.equals terms,
+ * to at least one element of the given list; false otherwise.
+ *
+ * Like {@link http://ramdajs.com/docs/#contains|R.contains} but with argument order reversed.
+ *
+ * @func contained
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category List
+ * @sig [a] → a → Boolean
+ * @param {Array} list The array to consider
+ * @param {*} a The item to compare against
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list, `false` otherwise.
+ * @see {@link http://ramdajs.com/docs/#contains|R.contains}, {@link http://ramdajs.com/docs/#flip|R.flip}
+ * @example
+ *
+ * RA.contained([1, 2, 3], 3); //=> true
+ * RA.contained([1, 2, 3], 4); //=> false
+ * RA.contained([{ name: 'Fred' }], { name: 'Fred' }); //=> true
+ * RA.contained([[42]], [42]); //=> true
+ */
+const contained = flip(contains);
+
+export default contained;
--- a/src/contained.js
+++ b/src/contained.js
@@ -15,7 +15,7 @@ import { flip, contains } from 'ramda';
  * @param {Array|String} list The list to consider
  * @param {*} a The item to compare against
  * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise.
- * @see {@link http://ramdajs.com/docs/#contains|R.contains}, {@link http://ramdajs.com/docs/#flip|R.flip}
+ * @see {@link http://ramdajs.com/docs/#contains|R.contains}
  * @example
  *
  * RA.contained([1, 2, 3], 3); //=> true
--- a/src/contained.js
+++ b/src/contained.js
@@ -2,7 +2,8 @@ import { flip, contains } from 'ramda';
 
 /**
  * Returns true if the specified value is equal, in R.equals terms,
- * to at least one element of the given list; false otherwise.
+ * to at least one element of the given list or false otherwise.
+ * Given list can be a string.
  *
  * Like {@link http://ramdajs.com/docs/#contains|R.contains} but with argument order reversed.
  *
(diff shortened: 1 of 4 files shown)


src/contained.js
```diff
@@ -0,0 +1,27 @@
+import { flip, contains } from 'ramda';
+
+/**
+ * Returns true if the specified value is equal, in R.equals terms,
+ * to at least one element of the given list; false otherwise.
```
Undistraction: … given list **or** false otherwise.

src/contained.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Like {@link http://ramdajs.com/docs/#contains|R.contains} but with argument order reversed.
+ *
+ * @func contained
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category List
+ * @sig [a] → a → Boolean
```
Undistraction: Should this maybe be `@sig [*] -> * -> Boolean given that the value may or may not be in the list?

src/contained.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Array} list The array to consider
```
Undistraction: 'list' instead of 'array'

Also, if it supports strings, this should be `{Array|String} I think.

src/contained.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {*} a The item to compare against
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list, `false` otherwise.
```
Undistraction: …list **or** `false` otherwise.

test/contained.js
```diff
@@ -0,0 +1,38 @@
+import R from 'ramda';
+
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('contained', function() {
+  context('R.contains tests', function() {
+    it('returns true if an element is in a list', function() {
```
Undistraction: '…**should** return true' (@char0n prefers this form). Same for all others.

src/contained.js
```diff
@@ -0,0 +1,27 @@
+import { flip, contains } from 'ramda';
+
+/**
+ * Returns true if the specified value is equal, in R.equals terms,
```
Undistraction: Should mention that it supports strings as well.

test/contained.js
```diff
@@ -0,0 +1,38 @@
...
+      eq(RA.contained([-0], 0), false);
+      eq(RA.contained([0], -0), false);
+      eq(RA.contained([NaN], NaN), true);
+      eq(RA.contained([new Just([42])], new Just([42])), true);
+    });
+
+    it('returns true if substring is part of string', function() {
+      eq(RA.contained('banana', 'ba'), true);
+    });
+  });
```
Undistraction: Add tests for curried form.

src/contained.js
Undistraction: No. If that is the signature for `R.contains` then it is correct.

src/contained.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @func contained
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category List
+ * @sig [a] → a → Boolean
+ * @param {Array|String} list The list to consider
+ * @param {*} a The item to compare against
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise.
+ * @see {@link http://ramdajs.com/docs/#contains|R.contains}, {@link http://ramdajs.com/docs/#flip|R.flip}
```
Undistraction: Don't think we need the link to `R.flip` as that is an implementation detail.

guillaumearm: I don't understand, it's already done.

Undistraction: I'm suggesting you *remove* the reference to `R.flip`.

guillaumearm: it's done.

Undistraction:
@guillaumearm Nice work.

src/contained.js
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Returns true if the specified value is equal, in R.equals terms,
+ * to at least one element of the given list or false otherwise.
+ * Given list can be a string.
+ *
+ * Like {@link http://ramdajs.com/docs/#contains|R.contains} but with argument order reversed.
+ *
+ * @func contained
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
```
char0n: Change to 2.8.0

src/contained.js
```diff
@@ -0,0 +1,28 @@
...
+ * @category List
+ * @sig [a] → a → Boolean
```
char0n: Change `→` for `->`. We're using to chars to represent the continuation.

src/contained.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {Array|String} list The list to consider
+ * @param {*} a The item to compare against
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise.
```
char0n: Drop the `.` at the end

test/contained.js
```diff
@@ -0,0 +1,43 @@
+import * as R from 'ramda';
+
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('contained', function() {
+  context('R.contains tests', function() {
```
char0n: drop the `context` block. It's necessary.

test/contained.js
```diff
@@ -0,0 +1,43 @@
...
+      eq(RA.contained([0], -0), false);
+      eq(RA.contained([NaN], NaN), true);
+      eq(RA.contained([new Just([42])], new Just([42])), true);
+    });
+
+    it('should return true if substring is part of string', function() {
+      eq(RA.contained('banana', 'ba'), true);
+    });
+
+    it('is curried', function() {
```
char0n: 'should be curried'

test/contained.js
```diff
@@ -0,0 +1,43 @@
...
+
+    it('should return false if an element is not in a list', function() {
+      eq(RA.contained([1, 2, 3, 9, 8, 7, 100, 200, 300], 99), false);
+    });
+
+    it('should return false for the empty list', function() {
+      eq(RA.contained([], 1), false);
+    });
+
+    it('has R.equals semantics', function() {
```
char0n: 'should have R.equals semantics'

guillaumearm:
@char0n: It's done 😉

src/contained.js
```diff
@@ -0,0 +1,28 @@
...
+ * Like {@link http://ramdajs.com/docs/#contains|R.contains} but with argument order reversed.
+ *
+ * @func contained
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig [a] -> a -> Boolean
+ * @param {Array|String} list The list to consider
+ * @param {*} a The item to compare against
+ * @return {Boolean} Returns Boolean `true` if an equivalent item is in the list or `false` otherwise
```
char0n:
```js
@return {boolean} ...
```
Should return bool primitive

Undistraction:
Nice work @guillaumearm

PR #512, opened 2018-04-24
Feat/while nil

jackmellis:
Two new methods that we've been using a lot lately: `allEqual` and `whileNil`

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -94,6 +94,7 @@ export { default as Y } from './Y';
 export { default as seq } from './seq';
 export { default as sequencing } from './seq';
 export { default as dispatch } from './dispatch';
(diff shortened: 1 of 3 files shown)


guillaumearm:
It's the same behavior that [RA.dispatch](https://char0n.github.io/ramda-adjunct/2.6.0/RA.html#.dispatch), so I think we can close this PR.

@jackmellis : This is why we like 1 feature by PR.

jackmellis:
Thanks. I hadn't realised dispatch did the same thing!

Undistraction:
@jackmellis I think `RA.dispatch`'s naming needs revisiting or aliasing as it is non-obviois (though it does makes sense).

PR #513, opened 2018-04-24
feat: allEqual method

jackmellis:
assert that all array items are equal

### DIFF
--- a/src/allEqual.js
+++ b/src/allEqual.js
@@ -1,0 +1,23 @@
+import { uniq, length, lte, __, pipe } from 'ramda';
+
+/* eslint-disable */
+/**
+  * Checks that all elemnets in a list are the same
+  *
+  *
+  *
+  * @func allEqual
+  * @memberOf RA
+  * @category List
+  * @sig [Iterable a] -> boolean
+  * @params {Iterable.<*>}
+
+  * @example
+  *
+  * allEqual([ 1, 2, 3, 4 ]); //=> false
+  * allEqual([ 1, 1, 1, 1 ]); //=> true
+  */
+/* eslint-enable */
+const allEqual = pipe(uniq, length, lte(__, 1));
+
+export default allEqual;
--- a/test/allEqual.js
+++ b/test/allEqual.js
@@ -1,0 +1,15 @@
+import { assert } from 'chai';
+
+import * as RA from '../src/index';
+
+describe('allEqual', function() {
+  it('returns true when all items are equal', function() {
+    assert.equal(RA.allEqual([4, 4, 4, 4]), true);
+  });
+  it('returns true when all items are deeply equal', function() {
+    assert.equal(RA.allEqual([{ key: 'foo' }, { key: 'foo' }]), true);
+  });
+  it('returns false when items are not equal', function() {
+    assert.equal(RA.allEqual([1, 1, 2, 1, 1]), false);
+  });
+});
--- a/src/index.js
+++ b/src/index.js
@@ -109,6 +109,7 @@ export { default as sliceTo } from './sliceTo';
 export { default as omitIndexes } from './omitIndexes';
 export { default as compact } from './compact';
 export { default as appendFlipped } from './appendFlipped';
+export { default as allEqual } from './allEqual';
 // Object
 export { default as paths } from './paths';
 export { default as renameKeys } from './renameKeys';


guillaumearm:
I propose a more flexible version :

**e.g.** : 
```js
const allEqual = R.curry((x, arr) => (
  R.all(R.equals(x), arr)
));

allEqual(42, [42, 42, 42]) // => true
allEqual(21, [42, 42, 42]) // => false
```

jackmellis:
The purpose of this method was so that we didn't have to pass in a value to compare to, we just want to check that all of the values are the same, regardless of what that value happens to be.

Undistraction:
I think it's worth adding. Might also be worth adding an `allUnique` that checks there are no duplicates.

If @char0n is happy with this addition I'll review.

guillaumearm:
@Undistraction : `allUnique`is not really the complement, the complement is `notAllEqual`.

Effectively, we don't want to add a `notAllEqualTo` because using `notAllEqual` is enough.

Anyway, maybe we would add a complement of `allUnique` too.

Undistraction:
@guillaumearm You are correct. So that would mean:

- `allEqual` (all values are equal)
- `notAllEqual` (at least one value is different from the others)
- `allUnique` (all values are unique)
- `notAllUnique` (at least two values are identical). 

I wonder if there is a better word than `equal` to use in the naming? Unique captures it perfectly, but 'equal' poses the question 'equal to what?'.

guillaumearm:
> but 'equal' poses the question 'equal to what?'.

This is exactly the point I thought when I proposed `allEqualTo`, maybe @char0n will have an idea.

Undistraction:
I think 'allEqual' is perfectly valid but it needs to be called `allIdentical` or something similar.

guillaumearm:
Yup, I think the same thing but `allIdentical` is the best name IMO.

src/allEqual.js
```diff
@@ -0,0 +1,23 @@
+import { uniq, length, lte, __, pipe } from 'ramda';
+
+/* eslint-disable */
+/**
+  * Checks that all elemnets in a list are the same
+  *
+  *
```
char0n: Drop to redundant new lines pls

src/allEqual.js
```diff
@@ -0,0 +1,23 @@
+import { uniq, length, lte, __, pipe } from 'ramda';
+
+/* eslint-disable */
```
char0n: Why disable linting ?

test/allEqual.js
```diff
@@ -0,0 +1,15 @@
+import { assert } from 'chai';
+
+import * as RA from '../src/index';
+
+describe('allEqual', function() {
+  it('returns true when all items are equal', function() {
+    assert.equal(RA.allEqual([4, 4, 4, 4]), true);
+  });
+  it('returns true when all items are deeply equal', function() {
```
char0n: Use `should` convention in tests. Can be seen here: https://github.com/char0n/ramda-adjunct/blob/f7d7f7c454c2e49806b3fc0e1e521577a38ea544/test/argsPass.js#L32

src/allEqual.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+  * Checks that all elemnets in a list are the same
+  *
+  *
+  *
+  * @func allEqual
+  * @memberOf RA
+  * @category List
+  * @sig [Iterable a] -> boolean
```
char0n: Boolean is a category
```js
@sig [Iterable a] -> Boolean
```

src/allEqual.js
```diff
@@ -0,0 +1,23 @@
+import { uniq, length, lte, __, pipe } from 'ramda';
+
+/* eslint-disable */
+/**
+  * Checks that all elemnets in a list are the same
```
char0n: Returns true if all elements of the list match the each other, false if there are any that don't.

src/allEqual.js
char0n: Add
```js
@since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}

File: src/allEqual.js [hunk 6]
```diff
@@ -0,0 +1,23 @@
...
+  *
+  *
+  *
+  * @func allEqual
+  * @memberOf RA
+  * @category List
+  * @sig [Iterable a] -> boolean
+  * @params {Iterable.<*>}
```
char0n:
```js
@param {Iterable.<*>} iterable

File: src/allEqual.js [hunk 7]
```diff
@@ -0,0 +1,23 @@
...
+
```
char0n: Add
```js
@return {boolean}

File: src/allEqual.js [hunk 8]
```diff
@@ -0,0 +1,23 @@
...
+  * @example
+  *
+  * allEqual([ 1, 2, 3, 4 ]); //=> false
+  * allEqual([ 1, 1, 1, 1 ]); //=> true
+  */
+/* eslint-enable */
+const allEqual = pipe(uniq, length, lte(__, 1));
```
char0n: Prefer
```js
const allEqual = curry((x, arr) => (
  all(equals(x), arr)
));
```

char0n:
`allIdentical` will not be correct name, because we're doing deep equal here. I imagine identity to be different than equal.

src/allEqual.js
guillaumearm: It's not the same thing, this implementation had arity 2 and should be named `allEqualTo`

guillaumearm:
@char0n, you're totally right, we have the same problem with `allUnique` IMO, it suggests that all values are unique in term of references.

anyway, we should back to the initial proposed `allEqual` name, it suggests : `all equal to each other`.

src/allEqual.js
Undistraction: I agree with @guillaumearm. I think this is a separate function: `allEqualTo`.

Undistraction:
@guillaumearm I don't think 'unique' implies that. `R.uniq` is using `R.equals` for comparison - so in Ramda terms 'unique' means values that are not deep-equal. So I think we can legitimately use 'allUnique' (or possibly `allUniq` for consistency.

Undistraction:
This is a tough one as `equals` is verbose and `uniq` is contracted. But given that this is a sister function to `allEquals` I vote for `allUnique`, especially as the use of `R.uniq` is actually an implementation detail.

guillaumearm:
So, I'm a bit confused for this PR, It's `allEqual` or `allEquals` ?

Let's talk about `AllUnique` in the concerned thread.

Undistraction:
As far as I understand it:

1. This PR just needs to be completed as per @charOn's review. Current name is correct (allEqual).

2. We need to vote for naming of allUniq Vs allUnique (I vote for the latter).

3. We need to decide if we want allEqualTo which takes an additional value to compare values to.

guillaumearm:
2. Let's vote in #515 
3. I just created #525

src/allEqual.js
char0n: Agree. I missed the previous comments. Removing this comment.

char0n:
@Undistraction, @guillaumearm as this haven't been touched for longer time, there are comuple of options what we can do here:

1.) @jackmellis will  continue with the PR and finishes it (@jackmellis is that something you're willing to do?)
2.) we close this PR and create a new one with the same functionality
3.) we close it

jackmellis:
I'm quite happy to fix up the PR for the `allEqual` method - just been a bit busy with work lately.

char0n:
Have some usecases for this so I managed to continue with it here: https://github.com/char0n/ramda-adjunct/pull/572

Undistraction:
I'm going to go ahead and close as this is now added by #513

Issue #444, opened 2018-03-12
RA.lengthGt, RA.lengthLt, RA.lengthGte, RA.lengthLte, RA.lengthEq, RA.lengthNotEq


Undistraction:
```
const isLengthGt = curry((vLength, value) => compose(flip(gt)(vLength), length)(value))
const isLengthLt = curry((vLength, value) => compose(flip(lt)(vLength), length)(value))
const isLengthGte = complement(isLengthLt)
const isLengthLte = complement(isLengthGt)
const isLengthEq = curry((vLength, value) => compose(equals(vLength), length)(value))
const isNotLengthEq = complement(isLengthEq)
```

guillaumearm:
Soon, I never implements predicates anymore, I use ramda-adjunct :-D

Jokes apart, It's all good for me, if @char0n is OK, we can set a 'feature' label and move on PR.

char0n:
Why not, then this can be easily used in lenses

guillaumearm:
In lenses ? How ?

char0n:
```js
const a = { b: [1] };

RA.lensSatisfies(RA.lengthEq(3), R.lensProp('b'), a);
```

guillaumearm:
Yes ! With lensSatisfies, this is awesome !

Undistraction:
Should these be 'isLengthGt`, `isLengthGte` etc. insead?

Undistraction:
I've updated names. Are we happy with these? Think I will work on these next if so.

Undistraction:
@char0n Are you happy for me to go ahead with these?

PR #523, opened 2018-04-27
Add RA.lengthLt, RA.lengthGt, RA.lengthLte, RA.lengthGte

Undistraction:
See discussion here: #444

I'll add `RA.isLengthEqualTo` and `RA.isLengthNotEqualTo` in a separate PR once this is merged.


Undistraction:
@guillaumearm Thanks. 

> All functions implemented with complement return true when R.length computes a NaN

Good catch. We should do what Ramda does which is what the underlying JS does, so for the functions using `complement` we must return false for `undefined`, `null` and `NaN` or anything that isn't a number. I'll address this by not using `complement`.

> This functions works with objects that have the 'length' field

Looking at Ramda's docs for `R.length` (which does support objects) there is no mention of the fact. I think this is pretty edge-case and adding to documentation just confuses matters.

Undistraction:
I've removed usage of `R.complement`. I could extract a common function that accepts a comparator predicate:

```
const isLength = predicate => curry((valueLength, value) =>
  compose(flip(predicate)(valueLength), length)(value)
);
```

This should probably be internal.

guillaumearm:
> This is currently internal, but should we expose this publicly?

Very good idea, I think you should open an issue for this.

char0n:
Just my quick observation before I go deeper. I'd drop the `is` prefix from these functions to be consistent with e.g. `path*` or `prop*` functions. `isLengthLt` -> `lengthLt`.

src/index.d.ts
```diff
@@ -810,6 +809,32 @@ declare namespace RamdaAdjunct {
...
+         * Returns `true` if the supplied list or string has a length less than `valueLength`.
+         */
+        lengthLt<T>(valueLength: number, list: string | T[]): boolean;
+        lengthLt(valueLength: number): <T>(list: string | T[]) => boolean;
+
+        /**
+         * Returns `true` if the supplied list or string has a length less than or equal to
+         * `valueLength`.
+         */
+        isLenghtLte<T>(valueLength: number, list: string | T[]): boolean;
```
char0n: Drop `is` prefix here

src/index.d.ts
```diff
@@ -810,6 +809,32 @@ declare namespace RamdaAdjunct {
...
+        isLenghtLte(valueLength: number): <T>(list: string | T[]) => boolean;
```
char0n: Drop is prefix here

src/index.d.ts
```diff
@@ -810,6 +809,32 @@ declare namespace RamdaAdjunct {
...
+
+        /**
+         * Returns `true` if the supplied list or string has a length greater than or equal to
+         * `valueLength`.
+         */
+        isLenghtGte<T>(valueLength: number, list: string | T[]): boolean;
```
char0n: Drop is prefix here

src/index.d.ts
```diff
@@ -810,6 +809,32 @@ declare namespace RamdaAdjunct {
...
+        isLenghtGte(valueLength: number): <T>(list: string | T[]) => boolean;
```
char0n: Drop is prefix here

src/internal/compareLength.js
```diff
@@ -0,0 +1,8 @@
+import { curry, compose, flip, length } from 'ramda';
+
+const compareLength = comparator =>
```
char0n: Pls do the changes that were discussed in https://github.com/char0n/ramda-adjunct/issues/527

src/lengthGt.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Returns `true` if the supplied list or string has a length greater than `valueLength`.
+ *
+ * @func lengthGt
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param Number valueLength The length of the list or string
```
char0n:
```js
@param {number} ....
```
This also concerns every other jsdoc param that we have in this PR

src/lengthGte.js
```diff
@@ -0,0 +1,26 @@
+import { gte } from 'ramda';
+
+import compareLength from './internal/compareLength';
+
+/**
+ * Returns `true` if the supplied list or string has a length greater than or equal to
+ * `valueLength`.
+ *
+ * @func isLenghtGte
```
char0n: drop `is` prefix

src/lengthGte.js
```diff
@@ -0,0 +1,26 @@
...
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param Number valueLength The length of the list or string
+ * @param Number value The list or string
+ * @return {boolean}
+ * @see {@link http://ramdajs.com/docs/#gte|gte}, {@link http://ramdajs.com/docs/#length|length}
+ * @example
+ *
+ * RA.isLenghtGte(3, [1,2,3,4]); //=> true
```
char0n: Drop `is` prefix from example

src/lengthLt.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Returns `true` if the supplied list or string has a length less than `valueLength`.
+ *
+ * @func lengthLt
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param Number valueLength The length of the list or string
```
char0n: fix param sytnax

src/lengthLte.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Returns `true` if the supplied list or string has a length less than or equal to `valueLength`.
+ *
+ * @func lengthLte
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param Number valueLength The length of the list or string
```
char0n: fix param syntax

test/lengthGt.js
```diff
@@ -0,0 +1,30 @@
...
+    eq(RA.lengthGt(1, undefined), false);
+    eq(RA.lengthGt(1, null), false);
+    eq(RA.lengthGt(1, {}), false);
+    eq(RA.lengthGt(1, true), false);
+    eq(RA.lengthGt(1, false), false);
+    eq(RA.lengthGt(1, 5), false);
+  });
+
+  it(`should be curried`, function() {
+    eq(RA.lengthGt(1)([1, 2]), true);
```
char0n: Add
```js
eq(RA.lengthGt(1, [1, 2]), true);
```

Undistraction:
@char0n Those should all be addressed.

src/lengthGt.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Returns `true` if the supplied list or string has a length greater than `valueLength`.
+ *
+ * @func lengthGt
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {Number} valueLength The length of the list or string
```
char0n: Number should be lowercase `number`. It's a primitive not a wrapper object.

Undistraction:
@char0n The codebase is full of inconsistencies on this front (try searching for `@return {Boolean}` for example. I resisted fixing it all in this PR to keep it focussed but I'll try and find time to fix all the other issues in another PR.

char0n:
@Undistraction can we pls rebase this onto master ASAP. Nothing is preventing this for being in next release.

Undistraction:
@char0n Are you happy for me to squash and merge?

char0n:
@Undistraction yep go fo it!

PR #536, opened 2018-05-04
chore(commitlint): camel or kebab case for scope

### LINKED ISSUE #472: Restrict scope case
There is a [PR](https://github.com/marionebl/commitlint/issues/307#issuecomment-376480920) adding support for restricting scope to a subset of scopes to commit-lint. We currently have no restriction on scope because we couldn't limit to just one case. Once this PR drops we should restrict to:

```
"rules": {
    "scope-case": [2, ['camel-case', 'kebab-case']],
  }
```

Undistraction:
commitlint now supports multiple cases for scopes, so this PR adds config to support either camelcase or kebabcase in a commit message's scope:

```
scope(functionName): foo
```
or
```
scope(example-value): foo
```

Closes #472

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,6 +1,15 @@
 {
-  "extends": ["@commitlint/config-conventional"],
+  "extends": [
+    "@commitlint/config-conventional"
+  ],
   "rules": {
-    "scope-case": [0]
+    "scope-case": [
+      2,
+      "always",
+      [
+        "camel-case",
+        "kebab-case"
+      ]
+    ]
   }
 }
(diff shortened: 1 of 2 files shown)


char0n:
No package.json update ?

Undistraction:
@char0n My mistake. Added now.

char0n:
@Undistraction not sure I follow. We've updated the commitlint by removing the commitlint/cli from devDependencies ?

Undistraction:
@char0n Sorry. Something strange going on with project - can't seem to install new dependencies - tried on laptop and desktop. `npm install` succeeds but they aren't added to `package.json`. I've added it manually.

PR #551, opened 2018-05-13
feat: add thenP

char0n:
Ref #35

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -721,7 +721,7 @@ declare namespace RamdaAdjunct {
             (high: number, value: number): boolean;
             (high: number): (value: number) => boolean;
         };
-        
+
         /**
          * Returns whether or not an object has an own property with the specified name at a given path.
          */
@@ -776,6 +776,15 @@ declare namespace RamdaAdjunct {
          * Returns a Promise object that is rejected with the given reason.
          */
         rejectP<T>(value?: T): Promise<T>;
+
+        /**
+         * Composable shortcut for `Promise.then`.
+         * The thenP returns a Promise. It takes two arguments: callback function for the success of the Promise
+         * and the Promise instance itself.
+         */
+        thenP<T>(onFulfilled: Function, thenable: Promise<T>): Promise<T>;
+        thenP<T>(onFulfilled: Function): (thenable: Promise<T>) => Promise<T>;
+
 
         /**
          * Runs the given list of functions in order with the supplied object, then returns the object.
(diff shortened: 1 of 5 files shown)


src/index.d.ts
```diff
@@ -777,6 +777,15 @@ declare namespace RamdaAdjunct {
          */
         rejectP<T>(value?: T): Promise<T>;
 
+        /**
+         * Composable shortcut for `Promise.then`.
+         * The thenP returns a Promise. It takes two arguments: callback function for the success of the Promise
```
Undistraction: 'a callback function'

Undistraction: Either:

> The thenP function returns …

or

> thenP returns …

src/thenP.js
```diff
@@ -0,0 +1,27 @@
+import { invoker } from 'ramda';
+
+/**
+ * Composable shortcut for `Promise.then`.
+ * The thenP returns a Promise. It takes two arguments: callback function for the success of the Promise
```
Undistraction: See comments for description in `index.d.ts`

src/thenP.js
```diff
@@ -0,0 +1,27 @@
...
+ * and the promise instance itself.
+ *
+ * @func thenP
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category Function
+ * @sig (a -> Promise b | b) -> Promise b
+ * @param {Function} onFulfilled A Function called if the Promise is fulfilled. This function has one argument, the fulfillment value
```
Undistraction: Consider using 'success' here instead of 'fulfilled' for consistency with description (or changing 'success' in description for 'fulfilled').

src/thenP.js
```diff
@@ -0,0 +1,27 @@
...
+ * @param {Promise} promise Any Promise or Thenable object
+ * @return {Promise} A Promise in the pending status
```
Undistraction: A Promise in pending state

src/thenP.js
char0n: Copied directly from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

src/thenP.js
char0n: Copied directly from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

src/thenP.js
char0n: Will stick with `fulfilled` to be consistent with mdn

char0n:
@Undistraction are we OK here ? Can I go for merge ?

Undistraction:
Sorry. Yes. Looks good.

PR #552, opened 2018-05-13
feat: add lengthEq and lengthNotEq

### LINKED ISSUE #444: RA.lengthGt, RA.lengthLt, RA.lengthGte, RA.lengthLte, RA.lengthEq, RA.lengthNotEq

Undistraction:
Adds remaining functions to test of list or string length is equal or notEqual to the supplied length.

Note: I'm unsure of the correct behaviour for `lengthNotEq` when the value supplied does not have a length property. I am returning false, but should we do something else?

Closes #444

### DIFF
--- a/src/lengthEq.js
+++ b/src/lengthEq.js
@@ -11,10 +11,9 @@ import compareLength from './internal/compareLength';
  * @category List
  * @sig Number -> [*] -> Boolean
  * @param {number} valueLength The length of the list or string
- * @param {number} value The list or string
+ * @param {Array|string} value The list or string
  * @return {boolean}
- * @see {@link http://ramdajs.com/docs/#equals|equals},
- * {@link http://ramdajs.com/docs/#length|length}
+ * @see {@link RA.lengthNotEq|lengthNotEq}, {@link RA.lengthLt|lengthLt}, {@link RA.lengthGt|lengthGt}, {@link RA.lengthLte|lengthLte}, {@link RA.lengthGte|lengthGte},, {@link http://ramdajs.com/docs/#equals|equals}, {@link http://ramdajs.com/docs/#length|length}
  * @example
  *
  * RA.lengthEq(3, [1,2,3]); //=> true
--- a/src/lengthEq.js
+++ b/src/lengthEq.js
@@ -1,0 +1,25 @@
+import { equals } from 'ramda';
+
+import compareLength from './internal/compareLength';
+
+/**
+ * Returns `true` if the supplied list or string has a length equal to `valueLength`.
+ *
+ * @func lengthEq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {number} valueLength The length of the list or string
+ * @param {number} value The list or string
+ * @return {boolean}
+ * @see {@link http://ramdajs.com/docs/#equals|equals},
+ * {@link http://ramdajs.com/docs/#length|length}
+ * @example
+ *
+ * RA.lengthEq(3, [1,2,3]); //=> true
+ * RA.lengthEq(3, [1,2,3,4]); //=> false
+ */
+const lengthEq = compareLength(equals);
+
+export default lengthEq;
--- a/src/lengthNotEq.js
+++ b/src/lengthNotEq.js
@@ -11,10 +11,9 @@ import compareLength from './internal/compareLength';
  * @category List
  * @sig Number -> [*] -> Boolean
  * @param {number} valueLength The length of the list or string
- * @param {number} value The list or string
+ * @param {Array|string} value The list or string
  * @return {boolean}
- * @see {@link http://ramdajs.com/docs/#equals|equals},
- * {@link http://ramdajs.com/docs/#length|length}
+ * @see {@link RA.lengthEq|lengthEq}, {@link RA.lengthLt|lengthLt}, {@link RA.lengthGt|lengthGt}, {@link RA.lengthLte|lengthLte}, {@link RA.lengthGte|lengthGte}, {@link http://ramdajs.com/docs/#equals|equals}, {@link http://ramdajs.com/docs/#length|length}
  * @example
  *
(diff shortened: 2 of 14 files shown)


char0n:
CI is complaining about some jsdoc problems.

> Note: I'm unsure of the correct behaviour for lengthNotEq when the value supplied does not have a length property. I am returning false, but should we do something else?

IMHO is absolutely ok to return false in cases like these.

Undistraction:
> CI is complaining about some jsdoc problems.

It's strange - `npm run lint` doesn't pick up those for me locally. Anyway they are now fixed. 

> IMHO is absolutely ok to return false in cases like these.

OK.

src/lengthEq.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @func lengthEq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {number} valueLength The length of the list or string
+ * @param {number} value The list or string
+ * @return {boolean}
+ * @see {@link http://ramdajs.com/docs/#equals|equals},
```
char0n: @see is on multiple lines, should be on one line, or prettier is doing this?

src/lengthEq.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Returns `true` if the supplied list or string has a length equal to `valueLength`.
+ *
+ * @func lengthEq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {number} valueLength The length of the list or string
+ * @param {number} value The list or string
```
char0n: `@param {Array|string} value ...`

This also apply to all the `length*` functions that you did before. Include changes to these function into this PR.

src/lengthNotEq.js
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @func lengthNotEq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {number} valueLength The length of the list or string
+ * @param {number} value The list or string
+ * @return {boolean}
+ * @see {@link http://ramdajs.com/docs/#equals|equals},
```
char0n: Again on multiple lines.

Undistraction:
@char0n Btw don't feel you have to review this if you want to get a release out today. It can wait for the next release - just wanted to close out the remaining functions in that issue and thought it made sense for them all to go in the same release. But I don't think anyone will lose sleep if they don't.

src/lengthNotEq.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Returns `true` if the supplied list or string has a length not equal to `valueLength`.
+ *
+ * @func lengthNotEq
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.8.0|v2.8.0}
+ * @category List
+ * @sig Number -> [*] -> Boolean
+ * @param {number} valueLength The length of the list or string
+ * @param {number} value The list or string
```
char0n: `@param {Array|string} value ...`

This also apply to all the `length*` functions that you did before. Include changes to these function into this PR.

test/lengthEq.js
```diff
@@ -0,0 +1,33 @@
+import * as RA from '../src/index';
+import eq from './shared/eq';
+
+describe('lengthEq', function() {
+  it(`should return true if the length of a list is equal to the supplied length`, function() {
```
char0n:
```js
context('when the length of a list is equal to the supplied length', function() {
  specify('should return true', function() {
    ...
  });
});

File: test/lengthEq.js [hunk 6]
```diff
@@ -0,0 +1,33 @@
...
+    eq(RA.lengthEq(3, [1, 2, 3]), true);
+    eq(RA.lengthEq(3, [1, 2, 3, 4]), false);
+    eq(RA.lengthEq(3, [1, 2]), false);
+    eq(RA.lengthEq(0, []), true);
+  });
+
+  it(`should return true if the length of a string is equal to the supplied length`, function() {
```
char0n:
```js
context('the length of a string is equal to the supplied length', function() {
  specify('should return true', function() {
    ...
  });
});

File: test/lengthEq.js [hunk 7]
```diff
@@ -0,0 +1,33 @@
...
+    eq(RA.lengthEq(3, 'abc'), true);
+    eq(RA.lengthEq(3, 'abcd'), false);
+    eq(RA.lengthEq(3, 'ab'), false);
+    eq(RA.lengthEq(0, ''), true);
+  });
+
+  it(`should return false for values without a length property`, function() {
```
char0n:
```js
context('when there are values without a length property', function() {
    ...
  });
});

char0n: Do the same for `lensNotEq` tests.

char0n:
@Undistraction to late, I already did it ;] It looks nice, we need just need to do some improvements here. I need `length*` function next week (have a heavy usecases for them) so I want to release today or tomorrow. I think we can get this last PR in easily.

Undistraction:
I've also added @see refs to each of the other `length…` functions to each length function.

Undistraction:
Ah sorry. I use literals for everything as it means I can interpolate into stings without changing the deliverers which is quicker and better for seeing changes in diffs so I'm doing it out of habit. Obviously I'll try and remember to use normal deliverers, but there should be a way of longing for it.

char0n:
@Undistraction https://github.com/char0n/ramda-adjunct/issues/553

This PR can go inside. We'll deal with the string literals in #553.

PR #557, opened 2018-05-15
Update signature for curryRight

mimol91:
I think that signature for `curryRight` currently is not valid, so I've update it to match.

P.S. 
I am just learning FP, so I may be wrong.


Undistraction:
@mimol91 Thanks very much. That does indeed look like a mistake.

char0n:
@mimol91 yep thank you for the fix. It seems that I just copied the signature from curryRightN. I'm re-running the CI, after it passes we can merge.

PR #568, opened 2018-05-21
fix types for RA.contained

ipanasenko:


Undistraction:
@ipanasenko Thank you.

PR #572, opened 2018-05-24
feat: add allEqual

char0n:
Continuation of https://github.com/char0n/ramda-adjunct/pull/513

### DIFF
--- a/src/allEqual.js
+++ b/src/allEqual.js
@@ -6,7 +6,7 @@ import lengthLte from './lengthLte';
 // in https://github.com/char0n/ramda-adjunct/pull/513.
 
 /**
- * Checks that all elements in a list are equal.
+ * Returns true if all items in the list are equivalent using `R.equals` for equality comparisons.
  *
  * @func allEqual
  * @memberOf RA
--- a/src/allEqual.js
+++ b/src/allEqual.js
@@ -1,0 +1,28 @@
+import { pipe, uniq } from 'ramda';
+
+import lengthLte from './lengthLte';
+
+// Original idea for this function was conceived by https://github.com/jackmellis
+// in https://github.com/char0n/ramda-adjunct/pull/513.
+
+/**
+ * Checks that all elements in a list are equal.
+ *
+ * @func allEqual
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.9.0|v2.9.0}
+ * @category List
+ * @sig [a] -> Boolean
+ * @param {Array} list The list of values
+ * @return {boolean}
+ * @see {@link https://ramdajs.com/docs/#equals|equals}
+ * @example
+ *
+ * RA.allEqual([ 1, 2, 3, 4 ]); //=> false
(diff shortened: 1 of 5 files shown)


test/allEqual.js
```diff
@@ -0,0 +1,29 @@
+import { assert } from 'chai';
+
+import * as RA from '../src/index';
+
```
Undistraction: Might be worth adding a test for an array of `[1, 1, '1']` or is that actually testing `equal`?

test/allEqual.js
```diff
@@ -0,0 +1,29 @@
...
+  context('when items are not equal', function() {
+    specify('should return false', function() {
+      assert.isFalse(RA.allEqual([1, 1, 2, 1, 1]));
+    });
+  });
+
+  context('when empty list provided', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.allEqual([]));
+    });
```
Undistraction: What happens if no value is passed: `allEqual()`?
What happens if the same function is passed multiple times?
```
const f = () => {}
RA.allEqual([f, f, f]))
```

src/allEqual.js
```diff
@@ -0,0 +1,28 @@
+import { pipe, uniq } from 'ramda';
+
+import lengthLte from './lengthLte';
+
+// Original idea for this function was conceived by https://github.com/jackmellis
+// in https://github.com/char0n/ramda-adjunct/pull/513.
+
+/**
+ * Checks that all elements in a list are equal.
```
Undistraction: I think you need to note what 'equal' means in this context i.e. 'Checks that all elements in a list are equal using `R. uniq` which uses `R.equals` for comparison'.

test/allEqual.js
char0n: Yeah it's testing equal but we can have explicit test for this to demonstrate usecase for different types. I'll add it.

test/allEqual.js
char0n: Can add test for testing functions. 

> What happens if no value is passed: allEqual()?
I think we don't care. The signature is clear and everything outside of the signature
is unspecified.

src/allEqual.js
char0n: Will explain it explicitly.

char0n:
@rockymadden will investigate the algorithm, thx

char0n:
All CRC addressed. Requesting for re-review ;]

Undistraction:
Changes look good, but failing in CI.

rockymadden:
Can’t write our code now, away from laptop. The idea, however, is this: If you have an array of a million items and you use uniq on it, how many times will have equals have been run? Now if you use reduceWhile where you short circuit on the first non equality how many? The worst case scenario in the latter is the best case in the former right?

char0n:
The CI failured was caused by https://github.com/ramda/ramda/pull/1992; trying to compensate

char0n:
I'll go for 2.9.0 release. We have some new keywords in package.json that needs to go out ASAP.

PR #657, opened 2018-08-04
chore(LICENSE): use uptodate BSD 3-Clause License template

### LINKED ISSUE #645: License type
Chech if our license in actually not bsd 2 clause license

char0n:
https://opensource.org/licenses/BSD-3-Clause

Closes #645

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,29 +1,25 @@
 BSD 3-Clause License
 
-Copyright (c) 2017-2018, Vladimír Gorej
-All rights reserved.
+Copyright 2017-2018 Vladimír Gorej
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions are met:
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
 
-* Redistributions of source code must retain the above copyright notice, this
-  list of conditions and the following disclaimer.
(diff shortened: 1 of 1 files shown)


srghma:
```
But I’d like to ask one thing of you: please don’t use the BSD licence or the MIT licence. Use the Apache Licence instead.

MIT and BSD are probably perfectly good licences, but this is based on the assumption that a lot of the stuff that’s implied by the language of these licences actually holds in a court of law. The Apache Licence, on the other hand, works in almost exactly the same way but tends to have explicit language for all this stuff, and is thus much more likely to hold up in court.

An example of this is that MIT contains an implied patent grant, which matters because users of your software wouldn’t want you to be able to sue them for patent infringement if they use it. MIT basically says “you’re free to use this software without restrictions,” which implies that you won’t have to buy separate licences for patents held by the author. Apache has specific language for these patent grants.

So please consider the Apache Licence for your next project, or, even better, consider the Mozilla Licence or the GPL.

```

From https://github.com/bodil/spectre-of-free-software/blob/master/script.org

char0n:
Yeah lately I'm dealing with licenses a lot (by working in Oracle). It is not easy to change the license of existing project.  As contributors contributed all their stuff under this license. When setting up a new project new time, I'll think hard which license I choose.

char0n:
Need som explicit approvals here from core commiters to double check that nothing in the license changed.

Issue #667, opened 2018-08-17
padEnd, repeatStr


trustedtomato:
Hey! Firstly, thanks for all the effort you put into the library!
I've got some functions which might be suitable for the library too. I would be happy to do a PR!

---

`padEnd` & friends:
```javascript
const padEnd = curry((n, str) => str.padEnd(n));
const padEndWith = curry((n, str, padString) => str.padEnd(n, padString));
const padStart = curry((n, str) => str.padStart(n));
const padStartWith = curry((n, str, padString) => str.padStart(n, padString));
```

---

`repeatStr`:
```javascript
const repeatStr = curry((n, str) => str.repeat(n));
```

char0n:
Hi @trustedtomato, no problem we hope it helps people build better software.

Regarding your feature set, there is already an issue for it #400  and a PR #418. We've discontinued the efforts there. You can start fresh here from the start. I'd suggest starting with this: https://github.com/char0n/ramda-adjunct/pull/418

https://github.com/char0n/ramda-adjunct/issues/400#issuecomment-368325029

Why using ponyfill ? We want RA to work even on environments where String.prototype.repeat and other functions are not present

char0n:
These seems to be requested features so I'll start implementing this.

char0n:
repeatStr is on the way

PR #714, opened 2018-10-15
feat: add repeatStr

char0n:
Ref #667

I made this consistent with https://ramdajs.com/docs/#repeat. The data seems to be the first argument and I suppose there is a reason for this. To make it truly functional without consulting ramda, I'd switch the order of the arguments. In FP the data should be always last.

### DIFF
--- a/src/internal/polyfills/String.repeat.js
+++ b/src/internal/polyfills/String.repeat.js
@@ -1,0 +1,49 @@
+import isNotFinite from '../../isNotFinite';
+import isNegative from '../../isNegative';
+
+const repeat = (value, count) => {
+  let validCount = Number(count);
+
+  if (validCount !== count) {
+    validCount = 0;
+  }
+
+  if (isNegative(validCount)) {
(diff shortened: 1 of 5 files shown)


src/internal/polyfills/String.repeat.js
```diff
@@ -0,0 +1,49 @@
...
+    return '';
+  }
+
+  // Ensuring validCount is a 31-bit integer allows us to heavily optimize the
+  // main part. But anyway, most current (August 2014) browsers can't handle
+  // strings 1 << 28 chars or longer, so:
+  // eslint-disable-next-line no-bitwise
+  if (value.length * validCount >= 1 << 28) {
+    throw new RangeError('repeat count must not overflow maximum string size');
+  }
```
srghma: wow

char0n:
Putting this in without explicit PR approval. Pls provide ex-post CR.

char0n:
@srghma thanks

PR #715, opened 2018-10-15
feat: add allIdentical

### LINKED ISSUE #578: allIdentical
Same as `RA.allEqual` but using `R.identical` as a comparator.

char0n:
Closes #578

### DIFF
--- a/src/allIdentical.js
+++ b/src/allIdentical.js
@@ -1,0 +1,32 @@
+import { uniqWith, identical, pipe } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+import lengthLte from './lengthLte';
+
+/**
+ * Returns true if all items in the list are equivalent using `R.identical` for equality comparisons.
+ *
+ * @func allIdentical
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.11.0|v2.11.0}
+ * @category List
+ * @sig [a] -> Boolean
+ * @param {Array} list The list of values
+ * @return {boolean}
+ * @see {@link https://ramdajs.com/docs/#identical|identical}
+ * @example
+ *
+ * RA.allIdentical([ 1, 2, 3, 4 ]); //=> false
+ * RA.allIdentical([ 1, 1, 1, 1 ]); //=> true
+ * RA.allIdentical([]); //=> true
+ * RA.allIdentical([ {}, {} ]); //=> false
+ * RA.allIdentical([ () => {}, () => {} ]); //=> false
+ */
+const allIdentical = curry1(
+  pipe(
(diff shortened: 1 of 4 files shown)


test/allIdentical.js
```diff
@@ -0,0 +1,49 @@
...
+
+describe('allIdentical', function() {
+  context('given all items are identical', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.allIdentical([4, 4, 4, 4]));
+    });
+  });
+
+  context('given all items are deeply equal', function() {
+    specify('should return true', function() {
```
guillaumearm:
```suggestion
    specify('should return false', function() {
```

test/allIdentical.js
```diff
@@ -0,0 +1,49 @@
...
+    });
+  });
+
+  context('given items have different type', function() {
+    specify('should return false', function() {
+      assert.isFalse(RA.allIdentical([1, 1, '1', 1]));
+    });
+  });
+
+  context('given items are reference to function', function() {
```
guillaumearm: you can add : 
```js
  context('given items are same functions without the same reference', function() {
      const getF = () => () => {};
      assert.isFalse(RA.allIdentical([getF(), getF(), getF()]));
    });
  });
```

src/allIdentical.js
```diff
@@ -0,0 +1,32 @@
...
+ * @sig [a] -> Boolean
+ * @param {Array} list The list of values
+ * @return {boolean}
+ * @see {@link https://ramdajs.com/docs/#identical|identical}
+ * @example
+ *
+ * RA.allIdentical([ 1, 2, 3, 4 ]); //=> false
+ * RA.allIdentical([ 1, 1, 1, 1 ]); //=> true
+ * RA.allIdentical([]); //=> true
+ * RA.allIdentical([ {}, {} ]); //=> false
```
guillaumearm:
```js
RA.allIdentical([ () => {}, () => {} ]); //=> false
```

src/allIdentical.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+const allIdentical = curry1(
```
guillaumearm: Maybe we can make a more optimized version using `R.reduce` and `R.reduced` ?

On the other hand, the implementation is less beautiful.

char0n: We could as  a performance optimization after it is in master.

src/allIdentical.js
char0n: will add

test/allIdentical.js
char0n: make sense will add

char0n:
Everything addressed. Thank you for review

PR #716, opened 2018-10-15
feat: add allEqualTo

### LINKED ISSUE #525: allEqualTo
We talked about `allEqualTo` in PR #513 

Implementation : 
```js
const allEqualTo = R.curry((x, arr) => (
  R.all(R.equals(x), arr)
));

allEqualTo(42, [42, 42, 42]) // => true
allEqualTo(21, [42, 42, 42]) // => false
```

char0n:
Closes #525

### DIFF
--- a/src/allEqualTo.js
+++ b/src/allEqualTo.js
@@ -1,0 +1,25 @@
+import { all, equals, curry } from 'ramda';
+
+/**
+ * Returns true if all items in the list are equivalent to user provided value using `R.equals` for equality comparisons.
+ *
+ * @func allEqualTo
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.11.0|v2.11.0}
+ * @category List
+ * @sig a -> [b] -> Boolean
+ * @param {*} val User provided value to check the `list` against
+ * @param {Array} list The list of values
+ * @return {boolean}
+ * @see {@link RA.allEqual|allEqual}, {@link https://ramdajs.com/docs/#equals|equals}
+ * @example
+ *
+ * RA.allEqualTo(1, [ 1, 2, 3, 4 ]); //=> false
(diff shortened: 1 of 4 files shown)


src/allEqualTo.js
```diff
@@ -0,0 +1,24 @@
...
+ * @sig a -> [b] -> Boolean
+ * @param {*} val User provided value to check the `list` against
+ * @param {Array} list The list of values
+ * @return {boolean}
+ * @see {@link RA.allEqual|allEqual}
+ * @example
+ *
+ * RA.allEqualTo(1, [ 1, 2, 3, 4 ]); //=> false
+ * RA.allEqualTo(1, [ 1, 1, 1, 1 ]); //=> true
+ * RA.allEqualTo(1, []); //=> true
```
guillaumearm: IMO, even if `R.all(R.always(false), []) === true`, I think `RA.allEqualTo(1, [])` should return `false`.

Semantically, there is no value equal to 1 in this case, so this should return `false`.

Anyway, I can understand we don't want to update this behavior since this is the default behavior for `R.all`.

char0n: I was thinking about that too. `allEqualTo` is just a specialization of `R.all(R.always(false), [])` and in my mind it should have consistent behavior. It only makes sense in logical perspective. 

All members of empty group equal to constant value === true. It's consistent with how ramda and even lodash approach this problem in consistent pattern. @guillaumearm can we agree to go further with this and in the future be consistent with another functions ?

char0n: @guillaumearm @Undistraction any more thoughts on this ?

char0n: After CI finishes, I'm putting this in. If there is a feeling in the future that the behavior should be different, then we'll change it and bump the major version.

PR #782, opened 2018-12-30
Ramda placeholder support + tests consistency

char0n:


test/isFunction.js
```diff
@@ -1,12 +1,14 @@
...
 import eq from './shared/eq';
 import Symbol from './shared/Symbol';
 import args from './shared/arguments';
 import genFunc from './shared/genFunc';
 import asyncFunc from './shared/asyncFunc';
 
 describe('isFunction', function() {
-  it('tests a value for `Function`', function() {
+  it('should test value for a `Function`', function() {
     eq(RA.isFunction(genFunc), typeof genFunc === 'function');
```
srghma: `typeof genFunc === 'function'` should return false sometimes?

char0n: Yes, for envs where genFunc is not supported. Notice how the `genFunc` is created.

Issue #106, opened 2017-07-28
round, roundTo


char0n:
https://github.com/lodash/lodash/issues/2310

char0n:
RA.round is already implemented. Transforming this issue into R.roundTo.

PR #805, opened 2019-01-27
feat: add round

char0n:
Ref #106

### DIFF
--- a/src/round.js
+++ b/src/round.js
@@ -1,0 +1,27 @@
+import { bind } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+/**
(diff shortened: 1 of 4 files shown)


src/round.js
```diff
@@ -0,0 +1,26 @@
...
+ * RA.round(0.9); //=> 1
+ * RA.round(5.95); //=> 6
+ * RA.round(5.5); //=> 6
+ * RA.round(5.05); //=> 5
+ * RA.round(-5.05); //=> -5
+ * RA.round(-5.5); //=> -5
+ * RA.round(-5.95); //=> -6
+ */
+
+const round = curry1(Math.round.bind(Math));
```
char0n: use ramda bind here

PR #831, opened 2019-02-14
feat(async): add initial implementation

### LINKED ISSUE #830: async
Creates async interface from generator. https://www.promisejs.org/generators/

```js
const responsesP = async(function *() {
  const response1 = yield request('http://localhost:8000');
  const response2 = yield request('http://localhost:8000/2');

  return { response1, response2 };
});

responsesP().then(console.dir);
```
Yield delegation and recursion delegatin must be tested too to guarantee correctness of the implementation.

char0n:
Closes #830

### DIFF
--- a/src/async.js
+++ b/src/async.js
@@ -1,0 +1,61 @@
+import { curryN, bind } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+import resolveP from './resolveP';
+import rejectP from './rejectP';
+
+/**
+ * Takes a generator function and returns an async function.
+ * The async function returned is a curried function whose arity matches that of the generator function.
+ *
+ * Note: This function is handy for environments that does support generators but doesn't support async/await.
+ *
+ * @func async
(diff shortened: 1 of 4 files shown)


src/async.js
```diff
@@ -0,0 +1,28 @@
+import { curryN, bind } from 'ramda';
+
+import resolveP from './resolveP';
+import rejectP from './rejectP';
+
+const async = generatorFn =>
+  curryN(generatorFn.length, function asyncWrapper(...args) {
+    const iterator = bind(generatorFn, this)(...args);
+
+    const handle = result => {
```
char0n: Can be extracted for reuse

src/async.js
```diff
@@ -0,0 +1,28 @@
+import { curryN, bind } from 'ramda';
+
+import resolveP from './resolveP';
+import rejectP from './rejectP';
+
+const async = generatorFn =>
```
char0n: Curry1

PR #928, opened 2019-05-14
fix(mergePath) Fix TS typing for mergePath

### LINKED ISSUE #925: mergePath typing is wrong
TypeScript method definition for mergePath function seems to be wrong.

Current code:
```
mergePath(path: Array<Array<string | number>>, source: object, obj: object): object;
```

According to documentation, the first argument is array of strings or numbers, but not an array of arrays.

This example currently throws TS error:
```
RA.mergePath(
   ['outer', 'inner'],
   { foo: 3, bar: 4 },
   { outer: { inner: { foo: 2 } } }
); => { outer: { inner: { foo: 3, bar: 4 } }
```
-> `Type string is not assignable to to type (string | number)[]`

borys-kupar:
Closes #925

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -516,9 +516,9 @@ declare namespace RamdaAdjunct {
          * merged with the own properties of the provided `source`.
          * If a key exists in both objects, the value from the `source` object will be used.
          */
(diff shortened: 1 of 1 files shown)


char0n:
Thank you very much for contributing this!

PR #935, opened 2019-05-21
feat: add isIterable

### LINKED ISSUE #351: isIterable/isNotIterable
Can we add a predicate to check if a value is or isn't iterable?

A basic check:
```
either(isArray, isPlainObj)
```
Note: the `isPlainObj` as opposed to `isObj` is needed to avoid false positives for things like RegExp objects. 

But I'm not sure exactly what else Ramda classifies as an Iterable, so this will probably need to be fleshed out.

Note discussion and usecase in #350

dimitropoulos:
closes #351

I'm not sure why, but at least on my end I'm getting an error that `IterableIterator` is unknown, but it should be as far as I can tell:
https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2015.iterable.d.ts

either way, let me know what you think.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -47,6 +47,11 @@ declare namespace RamdaAdjunct {
         isArray(val: any): val is any[];
 
         /**
+         * Checks whether the passed value is iterable.
+         */
+        isIterable<T>(val: any): val is IterableIterator<T>;
+
+        /**
          * Checks if input value is an empty `Array`.
          */
         isEmptyArray(val: any): val is any[];
--- a/src/isIterable.js
+++ b/src/isIterable.js
@@ -1,0 +1,38 @@
+import { hasIn } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+import isFunction from './isFunction';
+
+/**
+ * Checks whether the passed value is iterable.
+ *
+ * @func isIterable
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.18.0|v2.18.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol}
+ * @return {boolean}
+ * @example
+ *
+ * RA.isIterable(['arrays', 'are', 'iterable']); //=> true
+ * RA.isIterable('strings are iterable, too'); //=> true
+ * RA.isIterable((function* () {})()); //=> true (generator objects are both iterable and iterators)
+ *
+ * RA.isIterable({}); //=> false
+ * RA.isIterable(-0); //=> false
+ * RA.isIterable(null); //=> false
+ * RA.isIterable(undefined); //=> false
+ */
+const isIterable = curry1(val => {
+  if (typeof Symbol === 'undefined') {
+    return false;
+  }
+
+  return (
+    hasIn(Symbol.iterator, Object(val)) && isFunction(val[Symbol.iterator])
+  );
+});
+
+export default isIterable;
--- a/test/isIterable.js
+++ b/test/isIterable.js
@@ -1,0 +1,212 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+// Notice: BigInt64Array and BigUint64Array should be added to this list once they are supported by RA's node version
+
+describe('isIterable', function() {
+  context('given the item is an array with items', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(['arrays', 'are', 'iterable']));
+    });
+  });
+
+  context('given the item is a non-empty string', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable('strings are iterable, too'));
+    });
+  });
+
+  context('given the item is an empty array', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable([]));
+    });
+  });
+
+  context('given the item is an empty string', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(''));
+    });
+  });
+
+  context('given the item is a Map', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Map()));
+    });
+  });
+
+  context('given the items is a Set', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Set()));
+    });
+  });
+
+  context('given the item is an Int8Array', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Int8Array()));
+    });
+  });
+
+  context('given the item is a Uint8Array', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Uint8Array()));
+    });
+  });
+
+  context('given the item is a Uint8ClampedArray', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Uint8ClampedArray()));
+    });
+  });
+
+  context('given the item is an Int16Array', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Int16Array()));
+    });
+  });
+
+  context('given the item is a Uint16Array', function() {
+    specify('returns true', function() {
+      assert.isTrue(RA.isIterable(new Uint16Array()));
+    });
+  });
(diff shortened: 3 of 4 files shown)


src/index.d.ts
```diff
@@ -46,6 +46,11 @@ declare namespace RamdaAdjunct {
          */
         isArray(val: any): val is any[];
 
+        /**
+         * Checks if input value is iterable.
```
char0n: Checks whether the passed value is iterable.

src/isIterable.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Checks whether the passed value is iterable.
+ *
+ * @func isIterable
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.18.0|v2.18.0}
+ * @category Type
+ * @sig * -> Iterable
```
char0n: @sig * -> Boolean

char0n: Pls add `@see` tag:
```js
@see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols|Iteration protocol}

File: src/isIterable.js [hunk 3]
```diff
@@ -0,0 +1,53 @@
...
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @example
+ *
+ * RA.isIterable(['arrays', 'are', 'iterable']); //=> true
```
char0n: Let's limit example to 6 values max here. Having all variations in example is not necessary.

src/isIterable.js
```diff
@@ -0,0 +1,53 @@
...
+ * RA.isIterable('strings are iterable, too');   //=> true
+ * RA.isIterable([]);                //=> true
```
char0n: Pls don't use empty spaces to pad-end the resulting values.
Use following formatting from e.g. `isNan` function:
```js
RA.isNaN('NaN'); // => false
RA.isNaN(undefined); // => false
RA.isNaN({}); // => false
RA.isNaN('blabla'); // => false

File: src/isIterable.js (hunk 3 above)
Comment (dimitropoulos) [reply src/isIterable.js]: I certainly understand why you would request that.  Please tell me which 6 you would like to keep.  I did remove a few of them. 

To the uninitiated, some people may be surprised that you can iterate over a string in a for loop, or that you can iterate over an empty string (which is otherwise sometimes falsey).  Similarly, I think it's useful to know that functions are not iterable but the return of a generator _is_.   Just the same, I think some people might expect that a WeakMap or WeakSet (without deeply understanding the collections themselves and how they work) would be iterable.  Not to mention, I left out from the example that you can create your own iterable (although it's in the tests) and I think that might be something worth including.  This is a somewhat tricky topic and I feel like it's best to be clear with how the function works.

Furthermore, I was wondering if you would have any opinions about how I handled checking for the presence of `Symbol`.  As it relates to this topic: perhaps we might put a mention in the example or elsewhere as to what happens if your browser doesn't support it.

All up to you :)

File: src/isIterable.js [hunk 5]
```diff
@@ -0,0 +1,51 @@
+/**
+ * Checks whether the passed value is iterable.
+ *
+ * @func isIterable
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.18.0|v2.18.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols|Iteration protocol}
```
char0n: Even more appropriate would be this here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol

We're checking for iterable protocol not the iteration protocol. There is a difference between those. Sorry ;]

src/isIterable.js
```diff
@@ -0,0 +1,51 @@
...
+ * RA.isIterable(new Promise((resolve, reject) => {})); //=> false
+ * RA.isIterable({'not': 'plain objects'}); //=> false
+ * RA.isIterable({}); //=> false
+ * RA.isIterable(new WeakMap); //=> false (new in ES2015)
+ * RA.isIterable(new WeakSet); //=> false (new in ES2015)
+ * RA.isIterable(NaN); //=> false
+ * RA.isIterable(-0); //=> false
+ * RA.isIterable(true); //=> false
+ */
+const isIterable = val => {
```
char0n: What about the following implementation ?
```js
import { hasIn } from 'ramda';
import curry1 from 'ramda/src/internal/_curry1';
import isFunction from './isFunction';

const isIterable = curry1((val) => {
  if (typeof Symbol === 'undefined') { return false; } 

  return hasIn(Symbol.iterator, Object(val)) && isFunction(val[Symbol.iterator]);
});
```
1.) supports currying and Ramda placeholder
2.) properly checks for Symbol support with `typeof` operator
3.) doesn't check for type of the iterator return value

char0n: I've created https://github.com/char0n/ramda-adjunct/issues/937 to define another `isAsyncIterable` predicate. It's questionable is `isIterable` should also return `true` for async iterables. I'd say YES. But this can be added later.

src/isIterable.js
char0n: > Furthermore, I was wondering if you would have any opinions about how I handled checking for the presence of Symbol. As it relates to this topic: perhaps we might put a mention in the example or elsewhere as to what happens if your browser doesn't support it.

I provided my opinion about implementation in another comment. IMHO we should return false when there is no support for iterables, because non of the objects in such environment is iterable.

char0n: > To the uninitiated, some people may be surprised that you can iterate over a string in a for loop, or that you can iterate over an empty string (which is otherwise sometimes falsey). Similarly, I think it's useful to know that functions are not iterable but the return of a generator is. Just the same, I think some people might expect that a WeakMap or WeakSet (without deeply understanding the collections themselves and how they work) would be iterable. Not to mention, I left out from the example that you can create your own iterable (although it's in the tests) and I think that might be something worth including. This is a somewhat tricky topic and I feel like it's best to be clear with how the function works.

That is why we have tests or people can open the browser console on our [documentation](https://char0n.github.io/ramda-adjunct/2.17.0/) and write `RA.isIterable(3)` to see the result. IMHO 6 cases suffice, like showing array, string and generator returns true; null, undefined and numbers returns false...

test/isIterable.js
```diff
@@ -0,0 +1,47 @@
...
+const myCustomIterable = {
+  *[Symbol.iterator]() {
+    yield 1;
+    yield 2;
+    yield 3;
+  },
+};
+
+describe('isIterable', function() {
+  it('tests a value to determine if it is iterable', function() {
```
char0n: Can we structure tests in following way ? Pls also use assertions from `chai` and not our custom `eq` function. We're trying to migrate  to more modern test structure lately; example here: https://github.com/char0n/ramda-adjunct/blob/master/test/allEqual.js

```js
describe('isIterable', function() {
  context('given value with iterable protocol', function() {
    specify('should return true', function() { 
      // all cases showing iterable protocol
    }); 
  });
  
  context('given value without iterable protocol', function() {
      // all cases showing non iterable protocol
    });
  });

  context('given custom iterable', function() {
    // define custom iterable here
    specify('should return true', function() {
      // test it
    });
  });

    const isIterable = RA.isIterable(R.__);

    assert.isTrue(isIterable([]));
  });
});

char0n: You can also add case for `asyncIterable` and show that it's now failing.

src/isIterable.js
dimitropoulos: ok. updating now.

src/isIterable.js
dimitropoulos: sounds good to me! updating now

test/isIterable.js
dimitropoulos: ahh, I was copying from somewhere else - must have picked the wrong thing to copy.  updating now.

dimitropoulos: ok, I updated everything but I'm not quite sure exactly what you mean about adding a case `asyncIterable` failing.  Would you like me to add a test for that in the `isIterable` tests?  Seems like that should be a separate thing if/when that function is made.

char0n: Lets avoid ssynciterable for now

dimitropoulos: ok, so we're good to go on this PR then?

char0n:
Thank you. Merging. I'll make follow-up commit to make some minor changes

char0n:
@dimitropoulos https://github.com/char0n/ramda-adjunct/pull/939

Issue #237, opened 2017-12-09
nand


char0n:
```javascript
export const nand = complement(and);

**Comment by char0n** [comment]
@Undistraction

`and` in ramda is defined with Hindley-Milner signature `a → b → a | b`
This means that `and` can process any types, not just Booleans and can return any types.

Example of what I mean
```javascript
R.and(1, 2); // => 2
R.and("test", "test1"); // => "test1"
```
Keeping this in mind, if we implement `nand` as complement of `and` it will always return Boolean.
The signature be then  `a → b → Boolean`

Are we okay with that ? Does any usecase for this exist ?

@BjornMelgaard any opinion on this ?

BjornMelgaard:
@char0n 
What we should return otherwise? first parameter? there is `or` for this
I think its fine to return boolean and be closer to ramda as possible

char0n:
@BjornMelgaard the question was rather, do we need something like this in the lib ? ... rather then what it should return. It can return Boolean, and Boolean only,

BjornMelgaard:
@char0n is someone wants to, I've never used complement for and, I would rather add complement of path and prop

rockymadden:
Tagging on to this discussion: I am curious if there might be interest in adding more logical operations/logic gates/truth functions (e.g. AND, OR, XOR, NOT, NAND, NOR and XNOR) that might also align/dispatch with scalar/not-scalar types alike. I know ramada already has many of them (i.e. the relation functions). More of a placeholder comment for me for now.

Undistraction:
Complement of Ramda's `and`.

- T T = F
- T F = T
- F T = T
- F F = T

See: #234

dawehner:
This is implemented in https://github.com/char0n/ramda-adjunct/pull/952

PR #952, opened 2019-06-09
Implement nand

dawehner:
Tries to solve https://github.com/char0n/ramda-adjunct/issues/237

Note: While committing it, eslint-plugin-ramda had a bug, see https://github.com/ramda/eslint-plugin-ramda/pull/26, so I disabled this specific eslint rule in the configuration. Maybe I'm missing something.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -980,6 +980,11 @@ declare namespace RamdaAdjunct {
         neither(firstPredicate: Function, secondPredicate: Function): Function;
 
         /**
+         * Returns `true` if both arguments are falsy, otherwise `false`.
+         */
+        nand(a: any, b: any): Boolean;
+
+        /**
          * Takes a list of predicates and returns a predicate that returns true for a given list of
          * arguments if one or more of the provided predicates is not satisfied by those arguments.
          * It is the complement of Ramda's allPass.
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -980,9 +980,10 @@ declare namespace RamdaAdjunct {
         neither(firstPredicate: Function, secondPredicate: Function): Function;
 
         /**
-         * Returns `true` if both arguments are falsy, otherwise `false`.
+         * Returns true if both arguments are falsy; false otherwise.
          */
         nand(a: any, b: any): Boolean;
+        nand(a: any): (b: any) => Boolean;
 
         /**
          * Takes a list of predicates and returns a predicate that returns true for a given list of
--- a/src/nand.js
+++ b/src/nand.js
@@ -1,0 +1,28 @@
+import { and, complement } from 'ramda';
+
+/**
+ * Returns `true` if both arguments are falsy, otherwise `false`.
+ *
+ * @func nand
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.19.0|v2.19.0}
+ * @category Logic
+ * @sig a -> b -> Boolean
+ * @param {Boolean} a
+ * @param {Boolean} b
+ * @return {Boolean} True if both arguments are false.
+ * @example
+ *
+ * RA.nand(True, True); //=> False
+ * RA.nand(False, True); //=> True
+ * RA.nand(True, False); //=> True
+ * RA.nand(False, False); //=> True
+ * RA.nand(1.0, 1.0); //=> False
+ * RA.nand(1.0, 0); //=> True
+ * RA.nand(0, 1.0); //=> True
+ * RA.nand(0, 0); //=> True
+ */
+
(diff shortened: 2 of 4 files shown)


.eslintrc
```diff
@@ -31,6 +31,7 @@
     "quotes": ["error", "single", { "avoidEscape": true }],
     "ramda/always-simplification": ["error"],
     "ramda/compose-simplification": ["error"],
+    "ramda/complement-simplification": "off",
```
char0n: Can we turn this off just for the case of `nand` and  not globally ? 

```js
const nand = complement(and); // eslint-disable-line ramda/complement-simplification

File: src/index.d.ts [hunk 2]
```diff
@@ -979,6 +979,11 @@ declare namespace RamdaAdjunct {
          */
         neither(firstPredicate: Function, secondPredicate: Function): Function;
 
+        /**
+         * Returns `true` if both arguments are falsy, otherwise `false`.
```
char0n: `Returns true if both arguments are false; false otherwise.` - just to make it more consistent with `ramda` `and` description

src/nand.js
```diff
@@ -0,0 +1,28 @@
+import { and, complement } from 'ramda';
+
+/**
+ * Returns `true` if both arguments are falsy, otherwise `false`.
+ *
+ * @func nand
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.19.0|v2.19.0}
+ * @category Logic
+ * @sig a -> b -> Boolean
```
char0n:
```js
@sig a -> b -> a & b

File: src/nand.js [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+ * @param {Boolean} a
+ * @param {Boolean} b
+ * @return {Boolean} True if both arguments are false.
```
char0n: Pls drop the dot from the end of the sentense

src/nand.js
```diff
@@ -0,0 +1,28 @@
...
+ * @example
+ *
+ * RA.nand(True, True); //=> False
```
char0n: True, False are lowercase in `javascript`.

src/nand.js
```diff
@@ -0,0 +1,28 @@
...
+ * RA.nand(False, True); //=> True
+ * RA.nand(True, False); //=> True
+ * RA.nand(False, False); //=> True
+ * RA.nand(1.0, 1.0); //=> False
+ * RA.nand(1.0, 0); //=> True
+ * RA.nand(0, 1.0); //=> True
+ * RA.nand(0, 0); //=> True
+ */
+
```
char0n: drop this empty line pls

test/nand.js
```diff
@@ -0,0 +1,15 @@
+import eq from './shared/eq';
+import { nand } from '../src';
+
+describe('nand', function() {
+  it('compared two values with nand', function() {
```
char0n: Can you pls write test in the style more resembling this ? https://github.com/char0n/ramda-adjunct/blob/4f5068566fd34ccd0d60a2efb15732386cf2a963/test/hasPath.js#L10-L9

Divide the cases + add explicit case for currying.

Thank you

.eslintrc
dawehner: That makes way more sense, thank you!

src/index.d.ts
dawehner: 👍

src/nand.js
dawehner: Too much python recently I suppose

dawehner:
I hopefully addressed all your feedback now! Thank you for the super quick response

src/index.d.ts
char0n: This description is not sync with description in nand.js

test/nand.js
```diff
@@ -0,0 +1,37 @@
...
+    eq(nand(true, 1.0), false);
+    eq(nand(null, true), true);
+    eq(nand(1.0, undefined), true);
+    eq(nand(0.0, false), true);
+    eq(nand(null, null), true);
+    eq(nand(null, undefined), true);
+  });
+
+  it('should support currying', function() {
+    const trueNand = nand(true);
```
char0n: Pls import RA in following way `import * as RA from '../src';`

can we simplify into ?
```js
it('should support currying', function() {
  eq(RA.nand(true, true), false);
  eq(RA.nand(true)(true), false);   
});

dawehner:
No worries, I totally hear you about opensource projects. Keeping things consistent can be really hard.

dawehner:
I've pushed up a change which hopefully addresses all your feedback. I hope rebasing the PRs is okay.

PR #1079, opened 2019-10-01
feat(symbol): add isSymbol

### LINKED ISSUE #806: isSymbol
Checks if value is classified as a Symbol primitive or object.

**Possible implementation**:
```js
const isSymbol =  curry1((val) => {
    return typeof val === 'symbol'
        || typeof val === 'object' && type(val) === 'Symbol';
});
```
[type](https://ramdajs.com/docs/#type) function is from ramda.

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `isSymbol` function.

peaonunes:
Closes #806 .

## Tests
- [x] Running tests `$ npm run test`
- [x] Running tests in browser `$ npm run test:web`
- [x] Running compatibility tests for supported ramda versions `$ npm run test:ramda`
- [x] Running code coverage numbers `$ npm run coverage`
- [x] Running linter `$ npm run lint`

### DIFF
--- a/src/isSymbol.js
+++ b/src/isSymbol.js
@@ -1,0 +1,30 @@
+import { type } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+/**
+ * Checks if input value is a Symbol.
+ *
+ * @func isSymbol
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol|MDN Symbol}
+ * @example
+ *
+ * RA.isSymbol(Symbol('1')); //=> true
+ * RA.isSymbol(Symbol(1)); //=> true
+ * RA.isSymbol('string'); //=> false
+ * RA.isSymbol(undefined); //=> false
+ * RA.isSymbol(null); //=> false
+ */
+const isSymbol = curry1(val => {
+  return (
+    typeof val === 'symbol' ||
+    (typeof val === 'object' && type(val) === 'Symbol')
+  );
+});
+
+export default isSymbol;
--- a/test/isSymbol.js
+++ b/test/isSymbol.js
@@ -1,0 +1,34 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import * as RA from '../src';
+
+describe('isSymbol', function() {
+  context('given a Symbol as value', function() {
+    it('should return true', function() {
+      assert.isTrue(RA.isSymbol(Symbol('any')));
+      assert.isTrue(RA.isSymbol(Symbol(1)));
+      assert.isTrue(RA.isSymbol(Symbol.iterator));
+    });
+  });
+
+  context('given anything else as value', function() {
+    it('should return false', function() {
+      assert.isFalse(RA.isSymbol(undefined));
+      assert.isFalse(RA.isSymbol(null));
+      assert.isFalse(RA.isSymbol(() => {}));
+      assert.isFalse(RA.isSymbol([]));
+      assert.isFalse(RA.isSymbol({}));
+      assert.isFalse(RA.isSymbol(NaN));
+      assert.isFalse(RA.isSymbol(true));
+      assert.isFalse(RA.isSymbol('string'));
+      assert.isFalse(RA.isSymbol(1));
+    });
+  });
+
+  it('should support placeholder to specify "gaps"', function() {
+    const isSymbol = RA.isSymbol(R.__);
+
+    assert.isFalse(isSymbol('abc'));
+  });
+});
--- a/src/index.js
+++ b/src/index.js
@@ -75,6 +75,7 @@ export { default as isFalsy } from './isFalsy';
 export { default as isRegExp } from './isRegExp';
 export { default as isNotRegExp } from './isNotRegExp';
 export { default as isSparseArray } from './isSparseArray';
+export { default as isSymbol } from './isSymbol';
 // Function
 export { default as stubUndefined } from './stubUndefined';
 export { default as stubNull } from './stubNull';
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -116,6 +116,11 @@ declare namespace RamdaAdjunct {
         isNotUndefined(val: any): boolean;
 
         /**
+         * Checks if input value is `Symbol`.
+         */
+        isSymbol(val: any): val is Symbol;
+
+        /**
          * Checks if input value is `null`.
          */
         isNull(val: any): val is null;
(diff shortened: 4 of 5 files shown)


peaonunes:
I was running into an issue with one Chrome test locally. I don't know how to proceed with this one:
```
not ok 3169 Headless Chrome - [undefined ms] - error
    ---
        message: >
            Error: Browser failed to connect within 30s. testem.js not loaded?
            Stderr:

            DevTools listening on ws://127.0.0.1:58623/devtools/browser/41e86244-023c-44ef-917f-4d2dad004660
            [1001/225324.853425:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
            [1001/225324.879568:WARNING:gpu_process_host.cc(1207)] The GPU process has crashed 1 time(s)
            [1001/225325.188327:WARNING:gpu_process_host.cc(980)] Reinitialized the GPU process after a crash. The reported initialization time was 0 ms
            [1001/225325.300001:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
            [1001/225325.313586:WARNING:gpu_process_host.cc(1207)] The GPU process has crashed 2 time(s)
            [1001/225325.576972:WARNING:gpu_process_host.cc(980)] Reinitialized the GPU process after a crash. The reported initialization time was 0 ms
            [1001/225325.700206:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)
            [1001/225325.716367:WARNING:gpu_process_host.cc(1207)] The GPU process has crashed 3 time(s)
            [1001/225325.716672:FATAL:gpu_data_manager_impl_private.cc(919)] The display compositor is frequently crashing. Goodbye.
            [1001/225325.824733:WARNING:system_snapshot_mac.cc(42)] sysctlbyname kern.nx: No such file or directory (2)


        browser log: |
            [object Object]
            [object Object]
    ...
```

src/isSymbol.js
```diff
@@ -0,0 +1,32 @@
...
+ * RA.isSymbol('string'); //=> false
+ * RA.isSymbol(undefined); //=> false
+ * RA.isSymbol(null); //=> false
+ */
+const isSymbol = curry1(val => {
+  return (
+    !!val &&
+    (typeof val === 'symbol' ||
+      (typeof val === 'object' && type(val) === 'Symbol') ||
+      (typeof val === 'object' || val.constructor.name === 'Symbol'))
```
peaonunes: I have added some checks on top of the suggested in the issue, for example, `!!val` and checking constructor name since some polyfills might define `Symbol` as objects. What do you think?

src/isSymbol.js
```diff
@@ -0,0 +1,32 @@
+import { type } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+/**
+ * Checks if input value is a Symbol.
+ *
+ * @func isSymbol
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
```
peaonunes: I had to guess the version based on current version.

src/isSymbol.js
```diff
@@ -0,0 +1,32 @@
...
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol|MDN Symbol}
+ * @example
+ *
+ * RA.isSymbol(Symbol('1')); //=> true
+ * RA.isSymbol(Symbol(1)); //=> true
+ * RA.isSymbol('string'); //=> false
+ * RA.isSymbol(undefined); //=> false
+ * RA.isSymbol(null); //=> false
+ */
+const isSymbol = curry1(val => {
```
peaonunes: The function is returning `true` when `val` is `{}` or `[]` or `new Error()` and I believe it should not return `true`. I did not understand the reason, I guess is something do to with `curry1`. Do you have any suggestions?

char0n:
Nice job, I'll process all your comments separately. Regarding chrome tests failing it seem to be somehow related to testem library and you system. I wouldn't worry about it anyway, this PR is run on CircleCI and everything is check automatically.

src/isSymbol.js
char0n: Version is correct, sorry for this. This will be automated in future.

src/isSymbol.js
char0n: I'd remove `!!val`, we are concerned here by happy path only, we don't need to eliminate negatives.

char0n: Also having check for constructor name is  a **no-go**. Everything with the constructor name symbol will be matched  by this function and we don't want that. In modern javascript there is [toStringTag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) symbol which solves this problem.

src/isSymbol.js
char0n: The problem is caused by bug in your code `typeof val === 'object' || val.constructor.name === 'Symbol'`. This should be joined by `&&`. But this code I suggested to remove anyway in another comment, so no big deal. When this matching statement is removed everything should be working as expected

test/isSymbol.js
```diff
@@ -0,0 +1,30 @@
...
+      assert.isFalse(RA.isSymbol(undefined));
+      assert.isFalse(RA.isSymbol(null));
+      assert.isFalse(RA.isSymbol(() => {}));
+      assert.isFalse(RA.isSymbol(NaN));
+      assert.isFalse(RA.isSymbol(true));
+      assert.isFalse(RA.isSymbol('string'));
+      assert.isFalse(RA.isSymbol(1));
+    });
+  });
+});
```
char0n: We need to add one additional test which is testing that the function is curried:
```js
    const isSymbol = RA.isSymbol(R.__);

    assert.isFalse(isSymbol('abc'));
  });

File: src/isSymbol.js (hunk 3 above)
Comment (peaonunes) [reply src/isSymbol.js]: You are right! Thanks.

File: src/isSymbol.js (hunk 1 above)
Comment (peaonunes) [reply src/isSymbol.js]: I understand. I have seen the usage of `toStringTag` in some articles, that is indeed correct. I should have checked the details on `type` before!

Comment (peaonunes) [reply src/isSymbol.js]: Done dc6d287c652e63f8ac31c3511ac97eb238bcf7a4 😄

File: test/isSymbol.js (hunk 4 above)
Comment (peaonunes) [reply test/isSymbol.js]: Done dc6d287c652e63f8ac31c3511ac97eb238bcf7a4

File: src/isSymbol.js [hunk 5]
```diff
@@ -0,0 +1,30 @@
...
```
char0n: This is just a suggestion. This function can be rewritten in more concise syntax without explicit return statement.
```js
const isSymbol = curry1(val =>
  typeof val === 'symbol' ||
  (typeof val === 'object' && type(val) === 'Symbol')
);

char0n:
And it's in! Thank you ;]

src/isSymbol.js
peaonunes: It can indeed! However, I believe the prettier + linter rules were automatically formatting this way.

PR #1080, opened 2019-10-01
feat(promise): add anyP

### LINKED ISSUE #807: anyP
Make sense to wait untill stage-2 is reached. https://github.com/tc39/proposal-promise-any

**Update:**
Already present in other Promise libraries. No need to wait for stage-2.

tynopet:
Closes #807.

### DIFF
--- a/src/anyP.js
+++ b/src/anyP.js
@@ -1,0 +1,33 @@
+import { bind } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+import isFunction from './isFunction';
+import polyfill, {AggregatedError} from './internal/polyfills/Promise.any';
+
+export const anyPPolyfill = curry1(polyfill);
+export {AggregatedError};
+
+/**
+ * returns a promise that is fulfilled by the first given promise to be fulfilled,
+ * or rejected with an array of rejection reasons if all of the given promises are rejected.
+ *
+ * @func anyP
+ * @memberOf RA
+ * @category Function
+ * @sig [Promise a] -> Promise [a]
+ * @param {Iterable.<*>} iterable An iterable object such as an Array or String
+ * @return {Promise} Returns a promise that is fulfilled by the first given promise to be fulfilled,
+ * or rejected with an array of rejection reasons if all of the given promises are rejected.
+ * @example
+ *
+ * RA.anyP([
+ *   Promise.resolve(1),
+ *   2,
+ *   Promise.reject(3),
+ * ]); //=> Promise(2)
+ */
+const anyP = isFunction(Promise.any)
+  ? curry1(bind(Promise.any, Promise))
+  : anyPPolyfill;
+
+export default anyP;
--- a/src/anyP.js
+++ b/src/anyP.js
@@ -2,10 +2,10 @@ import { bind } from 'ramda';
 import curry1 from 'ramda/src/internal/_curry1';
 
 import isFunction from './isFunction';
-import polyfill, {AggregatedError} from './internal/polyfills/Promise.any';
+import polyfill, { AggregatedError } from './internal/polyfills/Promise.any';
 
 export const anyPPolyfill = curry1(polyfill);
-export {AggregatedError};
+export { AggregatedError };
 
 /**
  * returns a promise that is fulfilled by the first given promise to be fulfilled,
--- a/src/anyP.js
+++ b/src/anyP.js
@@ -8,13 +8,13 @@ export const anyPPolyfill = curry1(polyfill);
 export { AggregatedError };
 
 /**
- * returns a promise that is fulfilled by the first given promise to be fulfilled,
+ * Returns a promise that is fulfilled by the first given promise to be fulfilled,
  * or rejected with an array of rejection reasons if all of the given promises are rejected.
  *
  * @func anyP
  * @memberOf RA
  * @category Function
- * @sig [Promise a] -> Promise [a]
+ * @sig [Promise a] -> Promise a
  * @param {Iterable.<*>} iterable An iterable object such as an Array or String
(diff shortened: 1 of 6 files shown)


src/anyP.js
```diff
@@ -0,0 +1,33 @@
...
+import curry1 from 'ramda/src/internal/_curry1';
+
+import isFunction from './isFunction';
+import polyfill, { AggregatedError } from './internal/polyfills/Promise.any';
+
+export const anyPPolyfill = curry1(polyfill);
+export { AggregatedError };
+
+/**
+ * returns a promise that is fulfilled by the first given promise to be fulfilled,
```
char0n: returns -> Returns

src/internal/polyfills/Promise.any.js
```diff
@@ -0,0 +1,36 @@
...
+  return new Promise((resolve, reject) => {
+    const onReject = e => {
+      exceptions.push(e);
+
+      if (exceptions.length === iterable.length) {
+        reject(new AggregatedError(exceptions));
+      }
+    };
+
+    map(
```
char0n: Suggestion without epxlicit `{}`. Formatting will be handled by prettier.
```js
map(p => resolveP(p).then(resolve).catch(onReject), [...iterable])

File: types/index.d.ts [hunk 3]
```diff
@@ -784,6 +784,12 @@ declare namespace RamdaAdjunct {
         */
         allSettledP<T>(iterable: Iterable<T>): Promise<Array<SettledPromise<T>>>;
 
+        /**
+         * anyP accepts an iterable of promises and returns a promise that is fulfilled by the first given promise to be fulfilled,
```
char0n: Doesn't match the description in the source: https://github.com/char0n/ramda-adjunct/pull/1080/files#diff-4cb076891bd6a121df6d7ea58b2d06b7R11. We need to have this identical.

test/anyP.js
```diff
@@ -0,0 +1,191 @@
...
+      const v3 = 3;
+      const actual = [v1, v2, v3];
+      const expected = 1;
+
+      assert.strictEqual(await RA.anyP(actual), expected);
+    });
+  });
+
+  context('given list of values and promises', function() {
+    specify('should return first of fulfilled values', async function() {
```
char0n: This test doesn't correspond with example here: https://github.com/char0n/ramda-adjunct/pull/1080/files#diff-4cb076891bd6a121df6d7ea58b2d06b7R27

test/anyP.js
```diff
@@ -0,0 +1,191 @@
...
+    specify(
+      'should fail with AggregatedError that contains a list of rejected reasons',
+      async function() {
+        const p1 = RA.rejectP(1);
+        const p2 = RA.rejectP(2);
+        const p3 = RA.rejectP(3);
+        const actual = [p1, p2, p3];
+        const expected = [1, 2, 3];
+
+        try {
```
char0n: Checkout the [assert.throw](https://www.chaijs.com/api/assert/#method_throws). It's more idiomatic to the problem you're trying to solve here.

types/index.d.ts
```diff
@@ -784,6 +784,12 @@ declare namespace RamdaAdjunct {
...
+         * or rejected with an array of rejection reasons if all of the given promises are rejected.
+         */
+        anyP<T>(iterable: Iterable<T>): Promise<T>;
```
char0n: We're missing tests for typings. You can learn more about how to do them here: https://github.com/char0n/ramda-adjunct/tree/master/types/test

test/anyP.js
tynopet: I tried this, but I couldn't use it. Assert.throw has an argument that is a function but it calls it without arguments and doesn't work with async functions.

src/anyP.js
tynopet: Oh, I'm sorry, I fixed it

src/internal/polyfills/Promise.any.js
tynopet: I removed the braces, but prettier saved the formating

types/index.d.ts
tynopet: Thank you, I fixed it

test/anyP.js
tynopet: Thank you, I fixed it

types/index.d.ts
tynopet: Thank you, I added tests for typings

test/anyP.js
```diff
@@ -0,0 +1,191 @@
...
+          await RA.anyP(actual);
+          throw new Error('resolving should fail');
+        } catch (e) {
+          assert.notStrictEqual(e.message, 'resolving should fail');
+          assert.strictEqual(e instanceof AggregatedError, true);
```
char0n:
```js
assert.instanceOf(e, AggreatedError)
```

char0n: This is just a suggestion with more concise way of doing the same

types/index.d.ts
```diff
@@ -784,6 +784,12 @@ declare namespace RamdaAdjunct {
         */
         allSettledP<T>(iterable: Iterable<T>): Promise<Array<SettledPromise<T>>>;
 
+        /**
+         * anyP returns a promise that is fulfilled by the first given promise to be fulfilled,
```
char0n: This is still different then the description in `anyP.js`. We have a convention to keep the description exactly the same.

```js
/**
 * Returns a promise that is fulfilled by the first given promise to be fulfilled,
 * or rejected with an array of rejection reasons if all of the given promises are rejected.
 *
....

char0n:
Nicely done. If you're interested I exposed more promise related functions with `Hacktoberfest` label: https://github.com/char0n/ramda-adjunct/issues?q=is%3Aissue+is%3Aopen+label%3AHacktoberfest

Issue #482, opened 2018-03-31
divideNum


Undistraction:
Divides two numbers. Equivalent to `b / a`.

**Preferred implementation**
```js
const divideNum = flip(divide);
```
`flip` and `divide` functions are already part of [ramda](https://ramdajs.com/docs).

If the person that is implementing this function comes up with the better name for the function, all suggestions are welcome.

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `divideNum` function.

guillaumearm:
+1 for divideBy

But multiply is commutative, why do you want to flip it ?

Undistraction:
@guillaumearm Of course. I meant `subtract` not `multiply` which isn't commutative.

guillaumearm:
You means #483

char0n:
**Original issue description by @Undistraction** 
I'm forever doing `flip(divide)`, so how about:

```
const divideBy = flip(divide)

const percentToRatio = divideBy(100)
percentToRatio(75) // => 0.75
```

Issue #653, opened 2018-08-02
renameKeysWith vs renameKeysBy


lgy87:
i note that the signature of renameKeysWith is `(a -> b) -> {a: *} -> {b: *}`
the first argument is unary，so is it better rename it to `renameKeysBy` according to lodash and ramda?

char0n:
There was an issue in ramda what is the distinction between the *By and *With. I'll try to find it. Can you pls provide some formalization of the rules if you are aware of it ?

lgy87:
```
// nope
R.startsWith
R.endsWith

// special
R.useWith
R.pickBy            first argument is binary
R.reduceBy       first arg is binary

// normal
// first argument is binary
R.differenceWith
R.dropRepeatsWith
R.groupWith
R.memoizeWith
R.mergeDeepWith
R.symmetricDifferenceWith
R.unionWith
R.uniqWith
R.zipWith

// first argument is unary
R.countBy
R.eqBy
R.groupBy
R.indexBy
R.maxBy
R.minBy
R.sortBy
R.uniqBy
```

as far as i know, all `*with`(except **startsWith**, **endsWith**) and `*By` obey the regularity in lodash.

please correct me if i am wrong.

char0n:
Your argumentation seems valid. What we can do to create alias for `renameKeysWith` and call it `renameKeysBy`. Or we can extend the current implementatin and allow a value to be passed inside the function along with the key.

lgy87:
I prefer the first solution to the second. but if `renameKeysWith`(take a binary callback) is also need, there will be a broken update. 😢

char0n:
You're right. Second solution would be a breaking change. Callback would consume value as the first argument and key as second. I tend to like the second option, which extend the functionality of the function. @Undistraction, @guillaumearm, @srghma, @rockymadden opinions on this ?

rockymadden:
First path with a caveat:

- A new binary version of `renameKeysWith` is made a la something like https://ramdajs.com/docs/#mapObjIndexed
- `renameKeysWith` is implemented with the above function, whatever it is called. It essentially is implemented ignoring the value parameter in its implementation.
- `renameKeysWith` has an alias/deprecation to `renameKeysBy` and signature remains the same.

char0n:
[...]

PR #1081, opened 2019-10-02
feat: add divideNum

willcaul:
https://github.com/char0n/ramda-adjunct/issues/482


src/divideBy.js
```diff
@@ -0,0 +1,20 @@
+import { divide, flip } from 'ramda';
+
+/**
+ * Divides two numbers, where the second number is divided by the first number.
+ *
+ * @func divideBy
```
char0n: I'm struggling with the name of the function. In original ticket I've suggested to use `divideNum`. The thing is that `By` and `With` suffixes on functions are reserved for special behavior described in [this issue](https://github.com/char0n/ramda-adjunct/issues/653). By looking at the name of the function I'd expect that it consumes unary function as a first argument. I'd prefer `divideNum` even though it's "uglier" then `divideBy`.

test/divideBy.js
```diff
@@ -0,0 +1,20 @@
...
+describe('divideBy', function() {
+  it('should divide the second number by the first number', function() {
+    eq(RA.divideBy(2, 1), 0.5);
+    eq(RA.divideBy(2.0, 1.0), 0.5);
+    eq(RA.divideBy(0, 2.0), Infinity);
+    eq(RA.divideBy(0.0, 2.0), Infinity);
+    eq(RA.divideBy(-0.0, 2.0), -Infinity);
+  });
+
+  it('should curry', function() {
```
char0n: To satisfy test for currying we need to test for function call combinations. For binary function, it's two cases:
```js
it('should curry', function() {
  eq(RA.divideBy(2)(1), 0.5);
  eq(RA.divideBy(2, 1), 0.5);
});

src/divideBy.js
willcaul: That makes sense... Thanks for the explanation.

willcaul:
@char0n - thanks for the PR comments... updated.

Regarding the name, I changed it from `divideBy` to `divideNum` as requested. But I was wondering... you think `divideByNum` would be clearer, or do you think it is clearer as-is?

char0n:
@willcaul not sure at this point, I think what matters is that `divideNum` and `subtractNum` are named by same convention and you can expect the same behavior. In future if we come up with better names, we can still alias the functions with new names and leave the old names for backward compatibility.

Issue #483, opened 2018-03-31
subtractNum


Undistraction:
Subtracts its first argument from its second argument.

**Preferred implementation**
```js
const subtractNum = flip(subtract);
```
`flip` and `subtract` functions are already part of [ramda](https://ramdajs.com/docs).

If the person that is implementing this function comes up with the better name for the function, all suggestions are welcome.

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `subtractNum` function.

guillaumearm:
@char0n: are you ok to put a `feature` label on  this and #482 ?
This functions are very useful (even in real life use cases)

char0n:
**Original issue description by @Undistraction :**
I'm forever doing  `flip(subtract)`, so how about:

```
const subtractValue = flip(subtract)

const subtract5 = subtractValue(5)
subtract5(6) // => 1
```

PR #1082, opened 2019-10-02
feat: add subtractNum

willcaul:
https://github.com/char0n/ramda-adjunct/issues/483


src/subtractValue.js
```diff
@@ -0,0 +1,20 @@
+import { flip, subtract } from 'ramda';
+
+/**
+ * Subtracts its first argument from its second argument.
+ *
+ * @func subtractValue
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Math
+ * @sig Number → Number → Number
```
char0n: Can we please replace with: `Number -> Number -> Number`. Although ramda is using `→` char we use `->` chars, something we can find on the keyboard ;]

src/subtractValue.js
```diff
@@ -0,0 +1,20 @@
+import { flip, subtract } from 'ramda';
+
+/**
+ * Subtracts its first argument from its second argument.
+ *
+ * @func subtractValue
```
char0n: I'd opt for a name `subtractNum` to be consistent with future `divideNum` naming.

test/subtractValue.js
```diff
@@ -0,0 +1,16 @@
...
+import eq from './shared/eq';
+
+describe('subtractValue', function() {
+  it('should subtract the first number from the second number', function() {
+    eq(RA.subtractValue(3, 5), 2);
+    eq(RA.subtractValue(5, 3), -2);
+    eq(RA.subtractValue(3, 'foo'), NaN);
+  });
+
+  it('should curry', function() {
```
char0n:
```js
it('should curry', function() {
  eq(RA.subtractValue(3)(5), 2);
  eq(RA.subtractValue(3, 5), 2);
});

willcaul:
@char0n - thanks for the PR comments... updated.

char0n:
This is good to go

Issue #342, opened 2018-02-08
toArray


char0n:
This functions should convert value to an Array. Following rules are required to be satisfied by the implementation and by the tests:

**Should return shallow copy of an array**
```js
RA.toArray([1, 2]); // [1, 2]
```

**Should return values of object**
```js
RA.toArray({'foo': 1, 'bar': 2}); //=> [1, 2]
```

**Should convert string to array**
```js
RA.toArray('foo'); //=> ['f', 'o', 'o']
```

**Should convert any Iterable to array**
```js
const map = new Map();
map.set('0', 'foo');
map.set(1, 'bar');

RA.toArray(map); //=> [['0', 'foo'], [1, 'bar']]
```

```js
const set = new Set([1, 2, 3]);

RA.toArray(set); //=> [1, 2, 3]
```

```js
const object = { '0': 'a', 'length': 1 };
object[Symbol.iterator] = Array.prototype[Symbol.iterator];

RA.toArray(object); //=> ['a']
```

Technically this function is very similar to function [Array.from](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from), except that `Array.from` returns empty array if plain javascript objects is provided. For the implementation it is possible to use `Array.from` internally (if available) and handle the the differences manually. 

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `toArray` function.

guillaumearm:
Can you be more precise about what is `RA.toArray` is supposed to do ?

char0n:
@guillaumearm should do exactly what https://lodash.com/docs/4.17.5#toArray does

char0n:
@rafaelpadovezi go ahead! It's all yours ! ;]

PR #1085, opened 2019-10-02
feat: add toArray #342

rafaelpadovezi:
#342

### DIFF
--- a/src/internal/polyfills/Array.from.js
+++ b/src/internal/polyfills/Array.from.js
@@ -11,10 +11,10 @@ const copyArray = (items, mapFn, thisArg) => {
   return isNotUndefined(mapFn) ? [...items].map(boundMapFn) : [...items];
 };
 
-const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
+const fromArray = (items, mapFn, thisArg) => {
   if (items == null) {
     throw new TypeError(
-      'Array.from: when provided, the second argument must be a function'
+      'Array.from requires an array-like object - not null or undefined'
     );
   }
 
--- a/src/internal/polyfills/Array.from.js
+++ b/src/internal/polyfills/Array.from.js
@@ -1,14 +1,24 @@
+import { bind } from 'ramda';
+
 import isIterable from '../../isIterable';
-import isUndefined from '../../isUndefined';
+import isNotUndefined from '../../isNotUndefined';
+import isNotNil from '../../isNotNil';
+import isNotFunction from '../../isNotFunction';
 
 const copyArray = (items, mapFn, thisArg) => {
-  if (isUndefined(mapFn)) return [...items];
-  if (isUndefined(thisArg)) return [...items].map(mapFn);
-  return [...items].map(mapFn).bind(thisArg);
+  const boundMapFn = isNotUndefined(thisArg) ? bind(mapFn, thisArg) : mapFn;
+
+  return isNotUndefined(mapFn) ? [...items].map(boundMapFn) : [...items];
 };
 
 const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
   if (items == null) {
+    throw new TypeError(
+      'Array.from: when provided, the second argument must be a function'
+    );
+  }
+
+  if (isNotNil(mapFn) && isNotFunction(mapFn)) {
     throw new TypeError(
       'Array.from: when provided, the second argument must be a function'
     );
--- a/src/internal/polyfills/Array.from.js
+++ b/src/internal/polyfills/Array.from.js
@@ -1,0 +1,20 @@
+import isIterable from '../../isIterable';
+import isUndefined from '../../isUndefined';
+
+const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
+  if (items == null) {
+    throw new TypeError(
+      'Array.from: when provided, the second argument must be a function'
+    );
+  }
+
+  if (isIterable(items)) {
+    if (isUndefined(mapFn)) return [...items];
+    if (isUndefined(thisArg)) return [...items].map(mapFn);
+    return [...items].map(mapFn).bind(thisArg);
+  }
+
+  return [];
+};
+
+export default fromArray;
--- a/src/internal/polyfills/Array.from.js
+++ b/src/internal/polyfills/Array.from.js
@@ -1,6 +1,12 @@
 import isIterable from '../../isIterable';
 import isUndefined from '../../isUndefined';
 
+const copyArray = (items, mapFn, thisArg) => {
+  if (isUndefined(mapFn)) return [...items];
+  if (isUndefined(thisArg)) return [...items].map(mapFn);
+  return [...items].map(mapFn).bind(thisArg);
+};
+
 const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
   if (items == null) {
     throw new TypeError(
@@ -9,9 +15,7 @@ const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
   }
 
   if (isIterable(items)) {
-    if (isUndefined(mapFn)) return [...items];
-    if (isUndefined(thisArg)) return [...items].map(mapFn);
-    return [...items].map(mapFn).bind(thisArg);
+    return copyArray(items, mapFn, thisArg);
   }
 
   return [];
--- a/src/toArray.js
+++ b/src/toArray.js
@@ -1,0 +1,23 @@
+import { ifElse, values } from 'ramda';
+
+import isIterable from './isIterable';
+
+/**
+ * Converts input to an array.
+ *
+ * @func toArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.21.0|v2.21.0}
+ * @category List
+ * @sig * -> [a]
+ * @param {*} val The value to convert
+ * @return {Array}
+ * @example
+ *
+ * RA.toArray([1, 2]); //=> [1, 2]
+ * RA.toArray({'foo': 1, 'bar': 2}); //=> [1, 2]
+ */
+
+const toArray = ifElse(isIterable, Array.from, values);
+
+export default toArray;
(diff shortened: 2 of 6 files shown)


rafaelpadovezi:
Hey @char0n, you wrote:

> For the implementation it is possible to use Array.from internally **(if available)**

Should I provide a polyfill for `Array.from`? If yes, should I base my code on [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill)?

char0n:
@rafaelpadovezi yes I think basing the implementation on `Array.from` is the best choice currently. And yes we need to also provide a polyfill if `Array.from` is not available. We can base the polyfill on MDN polyfill but it cannot be exact poly because of the licensing reasons. Replicate the algorytm but try to use already available function in ramda and ramda-adjunct. RA polyfill are maintained here: https://github.com/char0n/ramda-adjunct/tree/1e56c50929c7d88c3052c656230a6e96bb58a21b/src/internal/polyfills

src/internal/polyfills/Array.from.js
```diff
@@ -0,0 +1,24 @@
+import isIterable from '../../isIterable';
+import isUndefined from '../../isUndefined';
+
+const copyArray = (items, mapFn, thisArg) => {
+  if (isUndefined(mapFn)) return [...items];
+  if (isUndefined(thisArg)) return [...items].map(mapFn);
+  return [...items].map(mapFn).bind(thisArg);
```
char0n: I don't think you're correctly binding the `thisArg`. It should be more like:
```js
const copyArray = (items, mapFn, thisArg) => {
  const boundMapFn = isNotUndefined(mapFn) ? bind(mapFn, thisArg) : mapFn;

  return isUndefined(boundMapFn) ? [...items]  :  [...items].map(boundMapFn);
};
```
`bind` function is from ramda

src/internal/polyfills/Array.from.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
+  if (items == null) {
```
char0n: This logic is incorrect. The original [polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) makes assumption that if the `mapFn` is provided, it must be a function. We're assuming following: If `items` is `nil` we're throwing error about that fact that second argument must be a function. It should be:

```js
if (isNotNil(mapFn) && isNotFunction(mapFn) {
  throw new TypeError(
     'Array.from: when provided, the second argument must be a function'
  );
}

File: types/index.d.ts [hunk 3]
```diff
@@ -1156,6 +1156,11 @@ declare namespace RamdaAdjunct {
          * Identity type.
          */
         Identity: Function;
+
+        /**
+         * Converts input to an array.
+         */
+        toArray<T>(iterable: Iterable<T> | T): any[];
```
char0n: @willcaul can you pls look into this signature and ts tests ? Could it be more specific ?

src/toArray.js
```diff
@@ -0,0 +1,29 @@
...
+import polyfill from './internal/polyfills/Array.from';
+
+export const fromPolyfill = curry(polyfill);
+
+/**
+ * Converts input to an array.
+ *
+ * @func toArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.21.0|v2.21.0}
```
char0n: 2.22.0

src/toArray.js
```diff
@@ -0,0 +1,29 @@
+import { ifElse, values, curry } from 'ramda';
+
+import isIterable from './isIterable';
+import isFunction from './isFunction';
+import polyfill from './internal/polyfills/Array.from';
+
+export const fromPolyfill = curry(polyfill);
+
+/**
+ * Converts input to an array.
```
char0n: Converts value to an array.

src/internal/polyfills/Array.from.js
```diff
@@ -0,0 +1,24 @@
+import isIterable from '../../isIterable';
+import isUndefined from '../../isUndefined';
+
+const copyArray = (items, mapFn, thisArg) => {
+  if (isUndefined(mapFn)) return [...items];
+  if (isUndefined(thisArg)) return [...items].map(mapFn);
+  return [...items].map(mapFn).bind(thisArg);
+};
+
+const fromArray = (items, mapFn = undefined, thisArg = undefined) => {
```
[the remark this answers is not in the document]
rafaelpadovezi: The reason to set the defaults is imitate the behavior of `Array.from` when using `R.curry` for example:

`R.curry(Array.from)`
`> ƒ r(e){return 0===arguments.length||n(e)?r:t.apply(this,arguments)}`

src/internal/polyfills/Array.from.js
rafaelpadovezi: I hope I understand it correctly. The second argument check was added

src/internal/polyfills/Array.from.js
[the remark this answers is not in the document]
char0n: @rafaelpadovezi if you curry the function which has default param values you need to explicitly pass the `undefined` value to trigger usage of default param value. In our case `undefined` is default value so that's why I'm saying it doesn't make sense to define them explicitly.

more on the subject of currying and default params in this article: https://www.linkedin.com/pulse/function-currying-default-params-javascript-vladim%C3%ADr-gorej/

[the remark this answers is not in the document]
rafaelpadovezi: Thanks for the clarification @char0n. I removed the default values and fixed the other comments.

src/toArray.js
```diff
@@ -0,0 +1,30 @@
...
+ * @func toArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category List
+ * @sig * -> [a]
+ * @param {*} val The value to convert
+ * @return {Array}
+ * @example
+ *
+ * RA.toArray([1, 2]); //=> [1, 2]
```
char0n: Let's add some additional examples to demonstrate the power of this function:
```js
RA.toArray([1, 2]); //=> [1, 2]
RA.toArray({'foo': 1, 'bar': 2}); //=> [1, 2]
RA.toArray('abc'); //=> ['a', 'b', 'c']
RA.toArray(1); //=> []
RA.toArray(null); //=> []

File: test/toArray.js [hunk 8]
```diff
@@ -0,0 +1,129 @@
...
+
+      assert.sameOrderedMembers(actual, ['a']);
+    });
+
+    specify('should return shallow copy of array', function() {
+      const array = [1, 2];
+      const actual = RA.toArray(array);
+
+      assert.sameOrderedMembers(actual, [1, 2]);
+      assert.notEqual(actual, array);
```
char0n: You need to use `notStrictEqual` here to verify the shallow copy.

test/toArray.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+  });
+
+  context('given non object types', function() {
```
char0n:
```js 
 context('given non object types', function() {
    const assertIsEmptyArray = val => {
       assert.isArray(val);
       assert.isEmpty(val);
    }

    specify('should return an empty array', function() {
      assertIsEmptyArray(RA.toArray(1));
      assertIsEmptyArray(RA.toArray(undefined));
      assertIsEmptyArray(RA.toArray(null));
      assertIsEmptyArray(RA.toArray(true));
    });
  });

File: test/toArray.js [hunk 10]
```diff
@@ -0,0 +1,129 @@
...
+    specify('should return an empty array', function() {
+      assert.isArray(RA.toArray(1));
+      assert.isArray(RA.toArray(undefined));
+      assert.isArray(RA.toArray(null));
+      assert.isArray(RA.toArray(true));
+    });
+  });
+
```
char0n: We're missing tests for currying
```js
    const toArray = RA.toArray(R.__);

    assert.sameOrderedMembers(toArray([1, 2, 3]), [1, 2, 3]);
  });

File: test/toArray.js [hunk 11]
```diff
@@ -0,0 +1,129 @@
...
+  context('fromPolyfill', function() {
```
char0n: This should have exactly the same tests as `toArray`. What we're testing here is if `fromPolyfill` have the same behavior as `toArray`. We're not concerned with `mapFn` or `thisArgs` which we're not using anyway.

src/toArray.js
```diff
@@ -0,0 +1,30 @@
...
+ * RA.toArray({'foo': 1, 'bar': 2}); //=> [1, 2]
+ */
+
+const fromArray = isFunction(Array.from) ? curry1(Array.from) : fromPolyfill;
```
char0n: Move from array just after the `fromPolyfill` definition so that jsdoc comment applies to our `toArray` constant instead `fromPolyfill`

rafaelpadovezi:
@char0n Thanks for the comments. I'm learning a lot.

src/toArray.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Converts value to an array.
+ *
+ * @func toArray
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category List
+ * @sig * -> [a]
```
char0n: @sig * -> [*]

test/toArray.js
```diff
@@ -0,0 +1,122 @@
...
+
+      assert.sameOrderedMembers(actual, ['a']);
+    });
+
+    specify('should return a shallow copy of array', function() {
+      const array = [1, 2];
+      const actual = fromPolyfill(array);
+
+      assert.sameOrderedMembers(actual, [1, 2]);
+      assert.notEqual(actual, array);
```
char0n: notEqual -> notStringEqual

char0n: Ok this is the last think, although you're solution is correct here, `notStringEqual` captures you original intention here.

rafaelpadovezi: Do you mean `notStrictEqual`, right? I agree with you, this assertion is better for this case.

Issue #1084, opened 2019-10-02
pathOrLazy


char0n:
**Is your feature request related to a problem? Please describe.**

Sometimes creating a default value for [pathOr](https://ramdajs.com/docs/#pathOr) is not cheap.

```js
const key = pathOr(new Date().getTime() + Math.random(), ['options', 'key'], notification);
```
**Describe the solution you'd like**

Whould be nice to have `pathOrLazy` that accepts a function instead of value as the first argument.

```js
const key = pathOrLazy(() => new Date().getTime() + Math.random(), ['options', 'key'], notification);
```
This way computing a default value only happens when we really need it.

**Additional context**
[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `pathOrLazy` function.

willcaul:
Could I take this one?

char0n:
@willcaul sure, go ahead

PR #1086, opened 2019-10-03
feat(object): add pathOrLazy

willcaul:
https://github.com/char0n/ramda-adjunct/issues/1084

### DIFF
--- a/src/pathOrLazy.js
+++ b/src/pathOrLazy.js
@@ -1,0 +1,26 @@
+import { curryN, isNil, nAry, path, when } from 'ramda';
+
+/**
+ * If the given, non-null object has a value at the given path, returns the value at that path.
+ * Otherwise returns the result of invoking the provided function.
+ *
+ * @func pathOrLazy
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Object
+ * @typedefn Idx = String | Int
+ * @sig (() -> a) -> [Idx] -> {a} -> a
+ * @param {Function} fd The function that will return the default value.
+ * @param {Array} p The path to use.
+ * @param {Object} obj The object to retrieve the nested property from.
+ * @return {*} The data at `path` of the supplied object or the default value.
+ * @example
+ *
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {a: {b: 2}}); //=> 2
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {c: {b: 2}}); //=> "N/A"\
+ */
+const pathOrLazy = curryN(3, function pathOrLazy(fd, p, obj) {
+  return when(isNil, nAry(0, fd), path(p, obj));
+});
+
+export default pathOrLazy;
--- a/src/pathOrLazy.js
+++ b/src/pathOrLazy.js
@@ -8,8 +8,8 @@ import { curryN, identical, partial, pathOr, unary, when } from 'ramda';
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
  * @category Object
- * @typedefn Idx = String | Int
- * @sig (() -> a) -> [Idx] -> {a} -> a
+ * @typedef Idx = String | Int
+ * @sig ({a} -> a) -> [Idx] -> {a} -> a
  * @param {Function} defaultFn The function that will return the default value.
  * @param {Array} path The path to use.
  * @param {Object} obj The object to retrieve the nested property from.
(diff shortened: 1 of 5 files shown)


src/pathOrLazy.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param {Function} fd The function that will return the default value.
+ * @param {Array} p The path to use.
+ * @param {Object} obj The object to retrieve the nested property from.
+ * @return {*} The data at `path` of the supplied object or the default value.
+ * @example
+ *
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {a: {b: 2}}); //=> 2
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {c: {b: 2}}); //=> "N/A"\
+ */
+const pathOrLazy = curryN(3, function pathOrLazy(fd, p, obj) {
```
char0n: What do say to the following implementation ? It's simpler to reason about (for me sugjectively) and more explicit what is going on.
Also let's not deal with the arity problem - there is nothing we can do about it and it may be expected behavior to return function with first parameter applied by running it. Also it would be beneficial to pass the `obj`  into the default function to be able to compute the state from object in pure manner. We leave `pathOr` to define to return `defaultFn` and if it  is returns we call it with `obj`. Otherwise we return original value return by `pathOr`.
```js
const pathOrLazy = curryN(3, (defaultFn, path, obj) => when(
  identical(defaultFn), 
  partial(defaultFn, [obj]), 
  pathOr(defaultFn, path, obj)
))
```

willcaul: @char0n - thanks for the suggestion. I like your approach better :)

I updated the PR to include your suggestion but I also wrapped the `defaultFn` in `unary` so that `when` doesn't pass the `defaultFn` to itself as the second parameter after `obj`.  If you do not like this change then let me know and I can remove `unary`.  What are your thoughts?

char0n: @willcaul fine with me, if unary is solving some case i didn't see before let's leave it as you suggest.

src/pathOrLazy.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * If the given, non-null object has a value at the given path, returns the value at that path.
+ * Otherwise returns the result of invoking the provided function with the object.
+ *
+ * @func pathOrLazy
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Object
+ * @typedefn Idx = String | Int
```
char0n: IMHO this should be `@typedef` to be properly rendered in docs

src/pathOrLazy.js
```diff
@@ -0,0 +1,30 @@
...
+ * @sig (() -> a) -> [Idx] -> {a} -> a
```
char0n: `@sig ({a} -> a) -> [Idx] -> {a} -> a` - this should be more correct reflecting on the fact we provide the original object back to defaultFn

src/pathOrLazy.js
```diff
@@ -0,0 +1,30 @@
...
+ * @param {Function} defaultFn The function that will return the default value.
+ * @param {Array} path The path to use.
+ * @param {Object} obj The object to retrieve the nested property from.
+ * @return {*} The data at `path` of the supplied object or the default value.
+ * @example
+ *
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {a: {b: 2}}); //=> 2
+ * RA.pathOrLazy(() => 'N/A', ['a', 'b'], {c: {b: 2}}); //=> "N/A"\
```
char0n: `"N/A"\` -> `"N/A"`

test/pathOrLazy.js
```diff
@@ -0,0 +1,53 @@
...
+
+describe('pathOrLazy', function() {
+  const obj = {
+    a: { b: { c: 1, d: ['x', 'y', 'z'] } },
+    e: NaN,
+    f: null,
+    g: undefined,
+  };
+
+  it('returns value for requested path', function() {
```
char0n: We're using more BDD approach of describing our tests. Every test should start with `should`. In this case `should return value for requested path`.

test/pathOrLazy.js
```diff
@@ -0,0 +1,53 @@
...
+    eq(RA.pathOrLazy(() => {}, ['a', 'b', 'c'], obj), 1);
+  });
+
+  it('returns default value when requested path does not exist', function() {
```
char0n: If the test description contains conditional logic, they're harder to read. To eliminate complexity we're using [gherkin] (https://www.guru99.com/gherkin-test-cucumber.html) language and nested contexts to eliminate the complexity. Example of this can be found here: https://github.com/char0n/ramda-adjunct/blob/e440acdfa5c7f59475251226e6b0cd3e13d0b59c/test/allEqual.js#L25

The rules are: nested `describe`  will become `context`. Every `it` in `context` will become `specify`.

```js
context('given requested path does not exist', function() {
  specify('should return default value', function() {
    eq(RA.pathOrLazy(() => 7, ['nonexistent'], obj), 7);
  });
});

**Comment by willcaul** [comment]
Thanks @char0n. I addressed your PR comments.  Let me know if I should change anything else.

File: types/index.d.ts [hunk 7]
```diff
@@ -445,6 +445,26 @@ declare namespace RamdaAdjunct {
...
+        /**
+         * If the given, non-null object has a value at the given path, returns the value at that path.
+         * Otherwise returns the result of invoking the provided function with the object.
+         */
+        pathOrLazy<T>(
+            defaultValueFn: () => T,
+            path: Array<number | string>,
+            obj: object
+          ): T;
+          pathOrLazy<T>(
```
char0n: The indentation here looks off. Do you see it too ?

willcaul: yes you were right... I just updated it. good catch! 👍

char0n:
Let's wait for tests and we're good to go! Thank you for your contribution.

PR #1087, opened 2019-10-05
feat(promise): add firstP

### LINKED ISSUE #815: firstP/headP
This pattern is like a race with [anyP([ .. ])](https://github.com/char0n/ramda-adjunct/issues/807), which means that it ignores any rejections and fulfills as soon as the first Promise fulfills.

Promise patterns explained in [YDKN series](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/async%20%26%20performance/ch3.md#variations-on-all---and-race--)

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `firstP/headP` function.

tynopet:
Closes #815


test/firstP.js
```diff
@@ -0,0 +1,100 @@
...
+      const p4 = new Promise(resolve => setTimeout(resolve, 40, 'four'));
+      const actual = [p1, p2, p3, p4];
+      const expected = 'one';
+
+      assert.strictEqual(await RA.firstP(actual), expected);
+    });
+  });
+
+  context('given list of rejected promises', function() {
+    specify('should fail with a list of rejected reasons', async function() {
```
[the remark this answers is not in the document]
char0n: And I think that is correct behavior which is identical to Promise.race

src/firstP.js
```diff
@@ -0,0 +1,53 @@
...
+      p =>
+        resolveP(p)
+          .then(resolve)
+          .catch(onReject),
+      iterable
+    );
+  });
+};
+
+const curriedFirstP = curry1(firstP);
```
char0n: Pls rename to `curriedFirstP` to  `firstP`

src/firstP.js
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * RA.firstP([
+ *   Promise.resolve(1),
+ *   2,
+ *   Promise.reject(3),
+ * ]); //=> Promise(1)
+ */
+const firstP = iterable => {
+  const errors = [];
+  const isEqualsIterableLength = equals(length(iterable));
```
char0n: `hasEqualIterableLength` is more proper variable name here.

char0n: If the iterable is custom iterable here, the R.length will return always `NaN`. You cannot assume length of iterable without iterating it.
```js
const customIterable = {
    *[Symbol.iterator]() {
        yield 1;
        yield 2;
        yield 3;
    },
};

R.length(customIterable);

File: src/firstP.js [hunk 4]
```diff
@@ -0,0 +1,53 @@
...
+ * @see {@link RA.anyP|anyP}
+ * @example
+ *
+ * RA.firstP([
+ *   Promise.resolve(1),
+ *   2,
+ *   Promise.reject(3),
+ * ]); //=> Promise(1)
+ */
+const firstP = iterable => {
```
char0n: To get rid of `curriedFirstP`
```js
const firstP = curry(iterable => {
});

**Comment by char0n** [comment]
@tynopet after going through this couple of times, this `firstP` looks exactly the same in behavior as the `anyP`. I can see it now after the implementation. Original suggested behavior of `anyP` was:
```js
anyP([resolveP(1), resolveP(2), rejectP(3)); //=> [1 ,2]
```
Only the resolved promises gets into the resulting array.

-


Then it changed by [TC39 proposal](https://github.com/tc39/proposal-promise-any) to following:
```js
anyP([resolveP(1), resolveP(2), rejectP(3)); //=> 1
```
The second implementation is the implementation that you already correctly implemented. This makes `firstP` just alias of `anyP` because there is now not any divergence in behavior.

Would you agree with this assumptions ? If yes, then I'm sorry we sped unnecessary time on this `firstP` function, because I did not realize that we have it already implemented in `anyP`.

tynopet:
Yes, you are right

char0n:
@tynopet cool that you agree. I'm sorry I did not spot this earlier, but things like this sometimes happen in our line of work. Can you evolve this PR into simple aliasing of `anyP` with `headP` and `firstP` aliases ?

char0n:
Closing this in favor of https://github.com/char0n/ramda-adjunct/pull/1128. Thank you for your work!

Issue #814, opened 2019-02-05
noneP


char0n:
This pattern is like [allP([ .. ])](https://char0n.github.io/ramda-adjunct/2.21.0/RA.html#.allP), but fulfillments and rejections are transposed. All Promises need to be rejected—rejections become the fulfillment values and vice versa.

Promise patterns explained in [YDKN series](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/async%20%26%20performance/ch3.md#variations-on-all---and-race--)

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `noneP` function.

PR #1095, opened 2019-10-06
feat(function): add noneP

willcaul:
Closes https://github.com/char0n/ramda-adjunct/issues/814

### DIFF
--- a/src/noneP.js
+++ b/src/noneP.js
@@ -1,0 +1,29 @@
+import { compose, map } from 'ramda';
+
+import allP from './allP';
+import rejectP from './rejectP';
+import resolveP from './resolveP';
+
+/**
+ * Returns a Promise that is resolved with an array of reasons when all of the provided Promises reject, or rejected when any Promise is resolved.
+ *
+ * @func noneP
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Function
+ * @sig [Promise a] -> Promise [a]
+ * @param {Iterable.<*>} iterable An iterable object such as an Array or String
+ * @return {Promise} A Promise that is resolved with a list of rejection reasons if all Promises are rejected, or a Promise that is rejected with the fulfillment value of the first Promise that resolves.
+ * @example
+ *
+ * RA.noneP([Promise.reject('hello'), Promise.reject('world')]); //=> Promise(['hello', 'world'])
+ * RA.noneP([]); //=> Promise([])
+ * RA.noneP([Promise.reject(), Promise.resolve('hello world')]); //=> Promise('hello world')
+ * RA.noneP([Promise.reject(), 'hello world']); //=> Promise('hello world')
+ */
+const noneP = compose(
+  allP,
+  map(pr => resolveP(pr).then(rejectP, resolveP))
+);
+
+export default noneP;
--- a/test/noneP.js
+++ b/test/noneP.js
@@ -1,0 +1,72 @@
+import * as RA from '../src';
+import eq from './shared/eq';
+
+const wait = (ms = 0) => new Promise(res => setTimeout(res, ms));
+
+describe('noneP', function() {
+  context('given all promises reject', function() {
+    specify('should resolve with the reasons', async function() {
+      const reasons = await RA.noneP([
+        RA.rejectP(1),
+        RA.rejectP('b'),
+        RA.rejectP(),
+      ]);
+      eq(reasons, [1, 'b', undefined]);
+    });
+  });
+
+  context('given one promise resolves', function() {
+    specify('should reject with the fulfillment value', async function() {
+      const value = await RA.noneP([
+        RA.rejectP(),
+        RA.rejectP(),
+        RA.resolveP('oops'),
+      ]).catch(RA.resolveP);
+      eq(value, 'oops');
+    });
+  });
+
+  context('given multiple promises resolve', function() {
+    specify(
+      'should reject with the fulfillment value of the first resolved promise',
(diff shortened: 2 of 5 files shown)


test/noneP.js
```diff
@@ -0,0 +1,72 @@
...
+      const value = await RA.noneP([RA.rejectP(1), 2, RA.rejectP(3)]).catch(
+        RA.resolveP
+      );
+      eq(value, 2);
+    });
+  });
+
+  context('given an empty list', function() {
+    specify('should resolve with empty reasons', async function() {
+      const value = await RA.noneP([]).catch(RA.resolveP);
```
char0n: You need to use R.identity instead of RA.resolveP here

test/noneP.js
```diff
@@ -0,0 +1,72 @@
...
+      const reasons = await RA.noneP([
+        RA.rejectP(1),
+        RA.rejectP('b'),
+        RA.rejectP(),
+      ]);
+      eq(reasons, [1, 'b', undefined]);
+    });
+  });
+
+  context('given one promise resolves', function() {
```
char0n: Thinking what is more readable:
```js
  context('given one promise resolves', function() {
    specify('should reject with the fulfillment value', async function() {
      const value = await RA.noneP([
        RA.rejectP(),
        RA.rejectP(),
        RA.resolveP('oops'),
      ]).catch(RA.resolveP);
      eq(value, 'oops');
    });
  });
```
vs
```js
  context('given one promise resolves', function() {
    specify('should reject with the fulfillment value', async function() {
      try {
        const value = await RA.noneP([
          RA.rejectP(),
          RA.rejectP(),
          RA.resolveP('oops'),
        ]);
        throw new Error('should reject');
      }
      catch(e) {
         assert.strictEqual(e.message, 'oops');
      }
    });
  });
```
I'll leave the decision up to you, but if you do go for your original version pls divide the the definition of value const and assertion with one empty line. 

```js
  context('given one promise resolves', function() {
    specify('should reject with the fulfillment value', async function() {
      const value = await RA.noneP([
        RA.rejectP(),
        RA.rejectP(),
        RA.resolveP('oops'),
      ]).catch(RA.resolveP);

      eq(value, 'oops');
    });
  });
```

test/noneP.js
```diff
@@ -0,0 +1,72 @@
...
+      eq(value, 2);
+    });
+  });
+
+  context('given an empty list', function() {
+    specify('should resolve with empty reasons', async function() {
+      const value = await RA.noneP([]).catch(RA.resolveP);
+      eq(value, []);
+    });
+  });
```
char0n: Missing test for currying, see [anyP test](https://github.com/char0n/ramda-adjunct/blob/f8e87a8c71fcf0023d00d87c8fe8c04944bf2d0d/test/anyP.js#L180)

src/noneP.js
```diff
@@ -0,0 +1,29 @@
...
+ * @param {Iterable.<*>} iterable An iterable object such as an Array or String
+ * @return {Promise} A Promise that is resolved with a list of rejection reasons if all Promises are rejected, or a Promise that is rejected with the fulfillment value of the first Promise that resolves.
+ * @example
+ *
+ * RA.noneP([Promise.reject('hello'), Promise.reject('world')]); //=> Promise(['hello', 'world'])
+ * RA.noneP([]); //=> Promise([])
+ * RA.noneP([Promise.reject(), Promise.resolve('hello world')]); //=> Promise('hello world')
+ * RA.noneP([Promise.reject(), 'hello world']); //=> Promise('hello world')
+ */
+const noneP = compose(
```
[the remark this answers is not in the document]
willcaul: @char0n - I had gone with only 1 map for a slight efficiency improvement, but I agree yours is more readable so I have updated it to match yours. However, in the updated version, I continued using `.then` instead of `thenP` because unfortunately from what I can tell, `thenP` only accepts an `onFulfilled` callback (and not a `onRejected` callback)

willcaul:
@char0n - thanks for the help. let me know if there is anything else you want me to change

types/index.d.ts
```diff
@@ -804,6 +804,12 @@ declare namespace RamdaAdjunct {
          */
         allP<T>(iterable: Iterable<T>): Promise<T[]>;
 
+        /**
+         * Returns a Promise that is resolved with an array of reasons when all of the provided Promises reject, or rejected when any Promise is resolved.
+         * This pattern is like allP, but fulfillments and rejections are transposed - rejections become the fulfillment values and vice versa.
+         */
+        noneP<T>(iterable: Iterable<T | Promise<T>>): Promise<T[]>;
```
willcaul: @char0n - actually do you think it would make more sense if the return type was `Promise<any[]>` instead of `Promise<T[]>`? I am thinking that the rejection reasons could be anything (i.e. `any`) and the promise will resolve with the rejection reasons since the fulfillments and rejections are transposed 🤔

src/noneP.js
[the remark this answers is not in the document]
char0n: Yep, make sense. Thanks

types/index.d.ts
char0n: Can we have definitions ? One for rejection one for resolution case. Notice that `any` is not an array here. This will be verified in typings tests I guess.
```typescript
noneP<T>(iterable: Iterable<T | Promise<T>>): Promise<T[]>;
noneP<T>(iterable: Iterable<T | Promise<T>>): Promise<any>;
```

char0n:
I've create a [delayP issue](https://github.com/char0n/ramda-adjunct/issues/1099) inspired by your original `wait` function. Merging this PR, thank you for contributing!

PR #1107, opened 2019-10-11
feat(promise): add delayP

PranjalAgni:
# feat: add delayP

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 /lib
 /dist
 /ramda-adjunct-*.*.*.tgz
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,3 @@
 /lib
 /dist
 /ramda-adjunct-*.*.*.tgz
-.vscode
--- a/src/delayP.js
+++ b/src/delayP.js
@@ -1,0 +1,43 @@
+/**
+ *
+ * @func delayP
+ * @memberOf RA
+ * @category Function
+ * @sig a -> Promise a
+ * @param {Number|Object} Argument to be resolved after a certain timeout and returns value if provided.
+ * @return {Promise} A Promise that is resolved with the given value if provided after certain delay.
+ * @see {@link RA.delayP|delayP}
+ * @example
+ *
+ * RA.delayP(200); //=> Returns Promise(undefined) after 200ms
+ * RA.delayP({ timeout: 1000, value: 'hello world' }); //=> Returns Promise('hello world') after 1000ms
+ * RA.delayP.reject(100); //=> rejects after 100 milliseconds with `undefined` value
+ * RA.delayP.reject({ timeout: 100, value: new Error('error') }); //=> rejects after 100 milliseconds with `Error('error')` value
+ */
+
+const makeDelay = (value, delayInMS, promiseType) => {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      if (promiseType === 'resolve') {
+        return resolve(value);
+      }
+      return reject(value);
+    }, delayInMS);
+  });
+};
+
+const delayP = param => {
+  if (typeof param === 'number') {
+    return makeDelay(undefined, param, 'resolve');
+  }
+  return makeDelay(param.value, param.timeout, 'resolve');
+};
+
+delayP.reject = param => {
+  if (typeof param === 'number') {
+    return makeDelay(undefined, param, 'reject');
+  }
+  return makeDelay(param.value, param.timeout, 'reject');
+};
+
+export default delayP;
--- a/src/delayP.js
+++ b/src/delayP.js
@@ -1,3 +1,6 @@
+import { curry, propOr, partial, nth } from 'ramda';
+
+import isPlainObj from './isPlainObj';
 /**
  *
  * @func delayP
@@ -15,29 +18,22 @@
  * RA.delayP.reject({ timeout: 100, value: new Error('error') }); //=> rejects after 100 milliseconds with `Error('error')` value
  */
 
-const makeDelay = (value, delayInMS, promiseType) => {
-  return new Promise((resolve, reject) => {
-    setTimeout(() => {
-      if (promiseType === 'resolve') {
-        return resolve(value);
-      }
-      return reject(value);
-    }, delayInMS);
-  });
-};
+const makeDelay = curry((settleFnPicker, opts) => {
+  let timeout = opts;
+  let value;
 
-const delayPTemplate = (param, promiseType) => {
-  if (typeof param === 'number') {
-    return makeDelay(undefined, param, promiseType);
+  if (isPlainObj(opts)) {
+    timeout = propOr(0, 'timeout', opts);
+    value = propOr(value, 'value', opts);
   }
-  return makeDelay(param.value, param.timeout, promiseType);
-};
-const delayP = param => {
-  return delayPTemplate(param, 'resolve');
-};
 
-delayP.reject = param => {
-  return delayPTemplate(param, 'reject');
-};
+  return new Promise((...args) => {
+    const settleFn = settleFnPicker(args);
 
+    setTimeout(partial(settleFn, [value]), timeout);
+  });
+});
+
+const delayP = makeDelay(nth(0));
+delayP.reject = makeDelay(nth(1));
 export default delayP;
--- a/src/delayP.js
+++ b/src/delayP.js
@@ -2,20 +2,22 @@ import { curry, propOr, partial, nth } from 'ramda';
 
 import isPlainObj from './isPlainObj';
 /**
+ * Create a promise which resolves/rejects after
+ * the specified milliseconds.
  *
  * @func delayP
  * @memberOf RA
  * @category Function
  * @sig a -> Promise a
- * @param {Number|Object} Argument to be resolved after a certain timeout and returns value if provided.
- * @return {Promise} A Promise that is resolved with the given value if provided after certain delay.
- * @see {@link RA.delayP|delayP}
+ * @sig () -> Promise Undefined
+ * @param {number|Object}  milliseconds number of milliseconds or options object.
+ * @return {Promise} A Promise that is resolved/rejected with the given value if provided after certain delay.
  * @example
  *
- * RA.delayP(200); //=> Returns Promise(undefined) after 200ms
- * RA.delayP({ timeout: 1000, value: 'hello world' }); //=> Returns Promise('hello world') after 1000ms
- * RA.delayP.reject(100); //=> rejects after 100 milliseconds with `undefined` value
- * RA.delayP.reject({ timeout: 100, value: new Error('error') }); //=> rejects after 100 milliseconds with `Error('error')` value
+ * RA.delayP(200); //=> Promise(undefined)
+ * RA.delayP({ timeout: 1000, value: 'hello world' }); //=> Promise('hello world')
+ * RA.delayP.reject(100); //=> Promise(undefined)
+ * RA.delayP.reject({ timeout: 100, value: new Error('error') }); //=> Promise(Error('error'))
  */
 
 const makeDelay = curry((settleFnPicker, opts) => {
@@ -36,4 +38,5 @@ const makeDelay = curry((settleFnPicker, opts) => {
 
 const delayP = makeDelay(nth(0));
 delayP.reject = makeDelay(nth(1));
+
 export default delayP;
--- a/src/delayP.js
+++ b/src/delayP.js
@@ -26,18 +26,18 @@ const makeDelay = (value, delayInMS, promiseType) => {
   });
 };
 
-const delayP = param => {
+const delayPTemplate = (param, promiseType) => {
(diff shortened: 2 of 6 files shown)


src/delayP.js
```diff
@@ -0,0 +1,43 @@
...
+ * @see {@link RA.delayP|delayP}
+ * @example
+ *
+ * RA.delayP(200); //=> Returns Promise(undefined) after 200ms
+ * RA.delayP({ timeout: 1000, value: 'hello world' }); //=> Returns Promise('hello world') after 1000ms
+ * RA.delayP.reject(100); //=> rejects after 100 milliseconds with `undefined` value
+ * RA.delayP.reject({ timeout: 100, value: new Error('error') }); //=> rejects after 100 milliseconds with `Error('error')` value
+ */
+
+const makeDelay = (value, delayInMS, promiseType) => {
```
char0n: Cool, nice job for a first run! Not take a look at this implementation. It's a psedo code (I never run it), but the code make use of the tools we have already available in ramda (R) + ramda-adjnuct (RA) and make your implementation a little bit more concise and readable. It's also settings some sane defaults when objects with no properties that we expect is provided.
```js
const makeDelay = R.curry((settleFnPicker, opts) => {
  let timeout = opts;
  let value = undefined;

  if (RA.isPlainObject(opts)) {
    timeout = R.propOr(0, 'timeout', opts);
    value = R.propOr(value, 'value', opts);
  }

  return new Promise((...args) => {
    const settleFn = settleFnPicker(args);
    setTimeout(R.partial(settleFn, [value]), timeout);
  });
});

const delayP = makeDelay(R.nth(0));
delayP.reject = makeDelay(R.nth(1));
```

PranjalAgni: > ```js
> const delayP = makeDelay(R.nth(0));
> delayP.reject = makeDelay(R.nth(1));
> ```

Can you explain what this piece of code is doing.

PranjalAgni: Also can you please review how was my implementation.

char0n: > Also can you please review how was my implementation.

It was doing fine, until you got error from CI about repeating code. You eliminated the problem by increasing the complexity of the solution by introducing another helper called `delayPTemplate`. At that point the solution become unnecessarily complex. You switched the eliminating of the repetition for the complexity.

char0n: > Can you explain what this piece of code is doing.

This code is using concept called currying, one the most powerfull tools of FP.
```js
const add = curry((a, b) => a + b);
add(1)(2); //=> 3

const add1 = add(1);
add1(2); //=> 3
```
Now if you understand this look at the code I provided again and try to understand it.

PranjalAgni: Yup I got the concept of `currying` in FP. 
I had this doubt as you are passing `R.nth(0)` in  `makeDelay(R.nth(0));` you are not specifying any list or string as per the ramda docs `nth()` returns the nth element of given list or string.

char0n: `R.nth` is of arity two, which means.
```js
const pickZeroIndexValue = nth(0);
pickZeroIndexValue([1, 2]); //=> 1;
```
Then in our case `pickZeroIndexValue` is called `settleFnPicker` and is passed into `makeDelay` as first argument. When `makeDelay` is executed, `settleFnPicker` is used to consume arguments of the promise lamda `...args` which is an array and will pick the first value from it, which is the `resolve` function.

Observe:
```js
const argsFn = (...args) => console.dir(args); 
argsFn(1, 2, 3); //=> outputs [1, 2, 3]

Comment (PranjalAgni) [reply src/delayP.js]: Also can you explain this line `const settleFn = settleFnPicker(args);`
How `makeDelay()` knows about when to use `resolve()` and `reject()`.

Comment (char0n) [reply src/delayP.js]: `makeDelay` doesn't know anything about that. We configure it it by passing `settleFnPicker` as an argument. And `settleFnPicker` knows if it should to pick `resolve` (nth(0)) or `reject` (nth(1)).
```js
const settleFn = settleFnPicker(args); // settleFn at this point is either resolve or reject function
```

PranjalAgni:
Thanks @char0n for helping me out through the process I learned a lot about functional programming through this challenge, this is one of the best maintained code-base I have ever worked on I loved how you have used all the good tooling like `husky` , `eslint` , `code-climate` , `jsdoc` and `code coverage`. It was very systematic from starting to work on the feature to making the pull request. Once again thanks for helping me out through the very basics of FP

char0n:
@PranjalAgni glad you like the codebase and that you learned something. That's why we do it ;] I'll start working on code review here. Nice job!

src/delayP.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @func delayP
+ * @memberOf RA
+ * @category Function
+ * @sig a -> Promise a
+ * @param {Number|Object} Argument to be resolved after a certain timeout and returns value if provided.
+ * @return {Promise} A Promise that is resolved with the given value if provided after certain delay.
+ * @see {@link RA.delayP|delayP}
+ * @example
+ *
```
char0n: In examples we're trying to describe the result value, not how we got to that value

src/delayP.js
```diff
@@ -0,0 +1,39 @@
...
+
+  return new Promise((...args) => {
+    const settleFn = settleFnPicker(args);
+
+    setTimeout(partial(settleFn, [value]), timeout);
+  });
+});
+
+const delayP = makeDelay(nth(0));
+delayP.reject = makeDelay(nth(1));
```
char0n: insert one empty line before `export default delayP` to increase readability.

.gitignore
```diff
@@ -11,3 +11,4 @@
 /lib
 /dist
 /ramda-adjunct-*.*.*.tgz
+.vscode
```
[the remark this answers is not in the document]
PranjalAgni: Sure :)

src/delayP.js
```diff
@@ -0,0 +1,39 @@
+import { curry, propOr, partial, nth } from 'ramda';
+
+import isPlainObj from './isPlainObj';
+/**
+ *
+ * @func delayP
+ * @memberOf RA
+ * @category Function
+ * @sig a -> Promise a
```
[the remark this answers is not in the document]
PranjalAgni: Sure

src/delayP.js
```diff
@@ -0,0 +1,39 @@
+import { curry, propOr, partial, nth } from 'ramda';
+
+import isPlainObj from './isPlainObj';
+/**
```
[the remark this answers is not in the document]
PranjalAgni: Done

src/delayP.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @func delayP
+ * @memberOf RA
+ * @category Function
+ * @sig a -> Promise a
+ * @param {Number|Object} Argument to be resolved after a certain timeout and returns value if provided.
```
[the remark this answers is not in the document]
PranjalAgni: Done

src/delayP.js
```diff
@@ -0,0 +1,39 @@
...
+ * @return {Promise} A Promise that is resolved with the given value if provided after certain delay.
+ * @see {@link RA.delayP|delayP}
```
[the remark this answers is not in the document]
PranjalAgni: Done

src/delayP.js
```diff
@@ -0,0 +1,39 @@
...
+ * @example
+ *
+ * RA.delayP(200); //=> Returns Promise(undefined) after 200ms
```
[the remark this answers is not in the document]
PranjalAgni: Done

src/delayP.js
PranjalAgni: Okay :)

src/delayP.js
PranjalAgni: Done

.gitignore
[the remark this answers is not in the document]
PranjalAgni: > We're completely missing typescript types + tests

What is the use of `index.d.ts` file?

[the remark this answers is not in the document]
PranjalAgni: Also I cannot find test for all the functions inside `types/test` directory.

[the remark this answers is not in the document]
char0n: I guess you did find it eventually ;]

types/index.d.ts
```diff
@@ -839,6 +839,16 @@ declare namespace RamdaAdjunct {
          */
         rejectP<T>(value?: T): Promise<T>;
 
+        /**
+         * Composable shortcut for `setTimeout(Promise)`.
```
char0n: The description here needs to match the to the original function description here: https://github.com/char0n/ramda-adjunct/pull/1107/files#diff-c90c9d832e4d7336f62bd494cc38e07cR5

Pls drop the `Composable shortcut for setTimeout(Promise).` sentence

PranjalAgni: Sure

types/index.d.ts
```diff
@@ -839,6 +839,17 @@ declare namespace RamdaAdjunct {
          */
         rejectP<T>(value?: T): Promise<T>;
 
+        /**
```
```ts
/**
 * Create a promise which resolves/rejects after
 * the specified milliseconds.
 */
delayP: {
  <T>(value?: T): Promise<any>,
  reject<T>(value?: T): Promise<any>
};

File: test/delayP.js [hunk 12]
```diff
@@ -0,0 +1,41 @@
+import * as RA from '../src';
+import eq from './shared/eq';
+
+describe('delayP', function() {
```
char0n: Cool, let's now deal with the tests. First think we need to do is to aligned the style of the tests to new conventions. These conventions can be seen for example here: https://github.com/char0n/ramda-adjunct/blob/a56484defec103cd1bcff2f55099b9699d876892/test/allP.js#L5-L4. We're working hard on migrating all our tests to this style. Another thing is to use explic asserts instead of internal `eq` function.
```js
describe('delayP', function() {
  context('given an number as an argument', function() {
    specify('should delay the resolution of the returned promise', async function() {
       const delayedValue = await RA.delayP(10);
       
       assert.isUndefined(delayedValue);
    });
  });
});
```
Notice we can use power or mocha framework to write asynchronous tests directly usgin `await`. This is example how to write async test for rejections. Also don't use big numbers for delays - big numbers makes our entire suite slower by this delay.
```js
describe('delayP', function() {
  context('reject', function() {
    context('given an number as an argument', function() {
      specify('should delay the rejected of the returned promise', async function() {
        try {
          const delayedValue = await RA.delayP.reject(10);
          throw new Error('rejection should fail');
        } catch (e) {       
          assert.isUndefined(delayedValue);
        }
      });
    });
  });
});
```

`test delay when passing object with different properties as argument` - given object with unknown properties -> should use defaults.

We're also completely missing tests for currying. 

Let me know if you have any questions.

char0n:
@PranjalAgni how are we looking here ? I can take over this PR and finalize it. Let me know. Thank you for contributing!

PranjalAgni:
Hey @char0n just been very busy this week with my internship work, I will definitely get back to in a day or two and finish the PR.

char0n:
@PranjalAgni ok understood. No worries, take your time, I was just probing.

test/delayP.js
PranjalAgni: What test for currying we need to work on?

char0n: Here is an example of unary functional currying test: https://github.com/char0n/ramda-adjunct/blob/22f4ec2f3b9917391959c96a3ffb36381c82ae17/test/isNilOrEmpty.js#L97-L96

PranjalAgni: Can you quickly tell me how to test for equality in error objects. I am currently trying something like this but it seems `chai` does not support `deepEqual()` on error objects:
```js
    context('passing object as argument', function() {
      specify(
        'should delay the rejected of the returned promise',
        async function() {
         try {
            const delayedValue = RA.delayP.reject({
              timeout: 900,
              value: new Error('Gracefull Error'),
            });
            throw new Error('rejection should fail');
          } catch (e) {
            assert.deepEqual(new Error('Gracefull Error'), delayedValue);
          }
        }
      );
    });
```

PranjalAgni: Can you help me with the above one

char0n: Sorry for delay in answer, I took some time off. Ok so problem is assert of errors and equality. In your case it's sufficient to do strict equality comparison:
```js
    context('passing object as argument', function() {
      specify(
        'should delay rejection of the returned promise',
        async function() {
        const error = new Error('error');

         try {
            const actual = await RA.delayP.reject({
              timeout: 90,
              value: error,
            });
            throw new Error('rejection should fail');
          } catch (e) {
            assert.strictEqual(actual, error);
          }
        }
      );
    });

char0n:
@PranjalAgni is there something I can still help you with ?

PranjalAgni:
Just wrapping the work, will finish it soon.

PranjalAgni:
Please check CI build getting 'Access Denied' in Node v9(test_node_9)

PranjalAgni:
> I think we're good here. Great job and thank you for contributing!

Thank you once again, I enjoyed this contribution and the project. Opensource is awesome ❤

PR #1109, opened 2019-10-12
feat(type): add isNegativeZero

### LINKED ISSUE #784: isNegativeZero
Object.is can be used to provide the check under hood

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `isNegativeZero` function.

miguelcarvalho13:
Closes #784

### DIFF
--- a/src/isNegativeZero.js
+++ b/src/isNegativeZero.js
@@ -1,0 +1,22 @@
+import { identical } from 'ramda';
+
+/**
+ * Checks if value is a negative zero (-0).
+ *
+ * @func isNegativeZero
+ * @memberof RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see @see {@link RA.isPositiveZero|isPositiveZero}
+ * @example
+ *
+ * RA.isNegativeZero(-0); //=> true
+ * RA.isNegativeZero(+0); //=> false
+ * RA.isNegativeZero(null); //=> false
+ */
+const isNegativeZero = identical(-0);
+
+export default isNegativeZero;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -279,6 +279,11 @@ declare namespace RamdaAdjunct {
         isNegative(val: any): val is number;
 
         /**
+         * Checks if value is a negative zero (-0).
+         */
+        isNegativeZero(val: any): boolean;
+
+        /**
          * Checks if value is a non-positive `Number` primitive or object. This includes all
          * negative numbers and zero.
          */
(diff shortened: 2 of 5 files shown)


src/isNegativeZero.js
```diff
@@ -0,0 +1,22 @@
+import { identical } from 'ramda';
+
+/**
+ * Checks if value is a negative zero (-0).
+ *
+ * @func isNegativeZero
+ * @memberof RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.21.0|v2.21.0}
```
char0n: `@since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}`

types/index.d.ts
```diff
@@ -278,6 +278,11 @@ declare namespace RamdaAdjunct {
          */
         isNegative(val: any): val is number;
 
+        /**
+         * Checks if value is a negative zero (-0).
+         */
+        isNegativeZero(val: any): val is number;
```
char0n: IMHO this is incorrect. 
```ts
isNegativeZero(val: any): boolean;
```
..would be more appropriate. This function can return `false` for number values.

src/isNegativeZero.js
```diff
@@ -0,0 +1,22 @@
...
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link https://ramdajs.com/docs/#identical|R.identical}
```
char0n: We already have a PR for positive zero. Would be good to link to it.
`@see {@link https://ramdajs.com/docs/#identical|R.identical}`

I'd drop identical here, it's more an implementation detail that something that can really help you navigating the library.

src/isNegativeZero.js
miguelcarvalho13: Done ✅

types/index.d.ts
miguelcarvalho13: Totally agree. Tbh I'm not that well familiar with Typescript but seeing that now it totally makes sense!

Done ✅

src/isNegativeZero.js
miguelcarvalho13: Agreed!
Done ✅

miguelcarvalho13:
> Very nice work! You uncovered very important fact to me and that is R.identical is using `Object.is` polyfill under the hood. Thanks. Pls address couple of comments that I had we can go for a merge.

I saw your suggestion to use `Object.is` and just to be sure I thought to search for that in the Ramda website and then I found that in the docs.

Btw, I'm very impressed with this `ramda-adjunct` repo its very well maintained and organized 👏

char0n:
> I saw your suggestion to use Object.is and just to be sure I thought to search for that in the Ramda website and then I found that in the docs.

Yep, nice reasoning here, we avoided unnecessary polyfill.

> Btw, I'm very impressed with this ramda-adjunct repo its very well maintained and organized 

Thank you! With addition of your code it's going be even better! ;]

char0n:
Ok I just resolved the code conflicts with master, let's wait for tests and we can go for a merge then. Thank you for contributing!

PR #1111, opened 2019-10-12
Feat: isPositiveZero

js636f:


types/index.d.ts
```diff
@@ -278,6 +278,11 @@ declare namespace RamdaAdjunct {
          */
         isNegative(val: any): val is number;
 
+        /**
+         * Checks if value is a positive zero (+0).
+         */
+        isPositiveZero(val: any): val is boolean;
```
```ts
isPositiveZero(val: any): val is boolean;
```
must be
```ts
isPositiveZero(val: any): boolean;

File: src/isPositiveZero.js [hunk 2]
```diff
@@ -0,0 +1,23 @@
...
+ * Checks if value is a positive zero (+0).
+ *
+ * @func isPositiveZero
+ * @memberof RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link https://ramdajs.com/docs/#identical|R.identical}
```
char0n: `@see {@link RA.isNegativeZero|isNegativeZero}`

test/isPositiveZero.js
```diff
@@ -0,0 +1,38 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import * as RA from '../src';
+
+describe('isPositiveZero', function() {
+  context('given a positive zero (+0)', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isPositiveZero(+0));
```
char0n: Last thing: we are missing test for implicit positive zero
assert.isTrue(RA.isPositiveZero(0));

js636f: I think I did something wrong with Git console when sending this update. So, when I made first edit of the PR, I edited files, then issued `git add -A` in isPositiveZero branch, then `git commit -m "something"`, then `git rebase master -i`, then `git push -f origin isPositiveZero` and everything looks right. Then I tried to make the same and that "git" file appeared (I deleted it here). I think it's because I was in wrong branch or something like this (I don't fully understand the rebase command). How should I make a commit after the first editing and rebasing?

js636f: I think the issue is resolved, but any additional info on the workflow is highly appreciated.

char0n:
Nice work on this and thank you for contributing !

js636f:
@char0n, thank you!

test/isPositiveZero.js
js636f: OK, thanks!

Issue #1110, opened 2019-10-12
isSet


char0n:
**Is your feature request related to a problem? Please describe.**

Predicate for determining if a provided value is an instance of a Set.

**Describe the solution you'd like**

```js
RA.isSet(new Set([1, 2])); //=> true
RA.isSet([1, 2]); //=> false
```

**Additional context**
[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `isSet` function. Issue original conceived in https://github.com/char0n/ramda-adjunct/issues/66

realSaddy:
Could I handle this one?

Issue #724, opened 2018-10-25
isDate/isValidDate returns false when it should be true


vorticalbox:
Valid dates are being returned as false.

"ramda": "^0.25.0"
"ramda-adjunct": "^2.10.0"

node 8.12.0 && 9.7.1

```
const ramda = require('ramda')
const RA = require('ramda-adjunct')
console.log(RA.isDate(new Date()))
console.log(RA.isValidDate(new Date()))
console.log(RA.isNotDate(new Date()))
console.log(RA.isNotValidDate(new Date()))
```

```
false
false
true
true
```

https://repl.it/@vorticalbox/GoldenrodLuxuriousValue

char0n:
Thank you for report @vorticalbox. I'll look into this ASAP.

char0n:
Currently not able to reproduce in any browser - chrome, firefox, edge.

Also this [test](https://github.com/char0n/ramda-adjunct/blob/a3109fc1020c6f4a8686d032cb7d539a230af258/test/isDate.js#L8) is running on node 6,7,8 and 9 successfully and without problems.

Trying to investigate further why it's not working in your case.

char0n:
Cannot reproduce the issue on node 8.12.0 || 9.7.1. If I use your code and save in file named `test.js` and  run the file with `node test.js` everything is fine.

IMHO I think that the problem my be in REPL itself. `RA.isDate` is using [R.is](https://ramdajs.com/docs/#is) to check for date instances. JavaScript have prototypal nature and concept of realms. For example if you have two tabs in the browser opened and send a date object from one to another, even though the transferred object is of `Date` instance, if you check it with `transferedDate intanceof Date` it will return false. Both tabs have their own realms and `instanceof` operator only works for one realm, it cannot work cross realms. 

In Ramda-Adjunct we don't try to solve the realm problem, because we consider it a corner case not worth handling.

vorticalbox:
Thanks, the repl.it  I posted has the same issue, also the application I am using it in is in node lts running inside docker.

What's weird is when I just use ramda is function is completely fine.

srghma:
@char0n https://stackoverflow.com/a/643827/3574379

```
A work-around for this is to check the object's class via

Object.prototype.toString.call(date) === '[object Date]'

```

char0n:
Yes exactly, that is what we already do wjth other predicates

char0n:
Fixes are comming in next couple of days in 2.12.0 release

[...]

PR #1114, opened 2019-10-12
feat(type): add isSet

realSaddy:
#1110

### DIFF
--- a/src/isSet.js
+++ b/src/isSet.js
@@ -1,0 +1,25 @@
+import { is } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+/**
+ * Predicate for determining if a provided value is an instance of a Set.
+ *
+ * @func isSet
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link RA.isSet}
+ * @example
+ *
+ * RA.isSet(new Map()); //=> false
+ * RA.isSet(new Set()); //=> true
+ * RA.isSet(new Set([1,2]); //=> true
+ * RA.isSet(new Object()); //=> false
+ */
+
+const isSet = curry1(val => is(Set, val));
+
+export default isSet;
--- a/src/isSet.js
+++ b/src/isSet.js
@@ -1,4 +1,4 @@
-import { is } from 'ramda';
+import { type, identical, pipe } from 'ramda';
 import curry1 from 'ramda/src/internal/_curry1';
 
 /**
@@ -11,7 +11,7 @@ import curry1 from 'ramda/src/internal/_curry1';
  * @sig * -> Boolean
  * @param {*} val The value to test
  * @return {boolean}
- * @see {@link RA.isSet}
+ * @see {@link RA.isMap|isMap}}
  * @example
  *
  * RA.isSet(new Map()); //=> false
@@ -20,6 +20,11 @@ import curry1 from 'ramda/src/internal/_curry1';
  * RA.isSet(new Object()); //=> false
  */
 
-const isSet = curry1(val => is(Set, val));
+const isSet = curry1(
+  pipe(
+    type,
+    identical('Set')
+  )
+);
 
 export default isSet;
--- a/test/isSet.js
+++ b/test/isSet.js
@@ -1,24 +1,34 @@
+import * as R from 'ramda';
+import { assert } from 'chai';
+
 import * as RA from '../src';
-import eq from './shared/eq';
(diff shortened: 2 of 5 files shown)


src/isSet.js
```diff
@@ -0,0 +1,25 @@
...
+ * Predicate for determining if a provided value is an instance of a Set.
+ *
+ * @func isSet
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link RA.isSet}
```
char0n: `@see {@link RA.isMap|isMap}` - it doesn't make sense to link to itself. Somebody else is working on `isMap` PR, so let's link to it.

types/index.d.ts
```diff
@@ -362,6 +362,11 @@ declare namespace RamdaAdjunct {
          */
         isRegExp(val: any): boolean;
 
+        /**
+         * Checks if value is `Set`.
+         */
+        isSet(val: any): boolean;
```
```ts
isSet(val: any): val is Set

File: types/test/isSet.ts [hunk 3]
```diff
@@ -0,0 +1,16 @@
...
+RA.isSet(NaN); // $ExpectType boolean
+RA.isSet({}); // $ExpectType boolean
+RA.isSet([]); // $ExpectType boolean
+RA.isSet(() => {}); // $ExpectType boolean
+RA.isSet('string'); // $ExpectType boolean
+RA.isSet(1); // $ExpectType boolean
+RA.isSet(Object()); // $ExpectType boolean
+RA.isSet(Object(5)); // $ExpectType boolean
+RA.isSet(new Map()); // $ExpectType boolean
+RA.isSet(new Set()); // $ExpectType boolean
```
char0n: Move doing assertions about set on top. We always want to have positive stuff on top, and then continue with all the negative ;]

test/isSet.js
```diff
@@ -0,0 +1,24 @@
+import * as RA from '../src';
+import eq from './shared/eq';
+import Symbol from './shared/Symbol';
+import args from './shared/arguments';
+
+describe('isSet', function() {
```
char0n: Can you pls aligned your test with the new conventions we have for writing tests demonstrated here? https://github.com/char0n/ramda-adjunct/blob/9cd86e7d06de667c14f4fdc8661d0fd4a4c99dc3/test/isNotBoolean.js#L5

We also missing the test for currying completely.

src/isSet.js
```diff
@@ -0,0 +1,25 @@
...
+ * @example
+ *
+ * RA.isSet(new Map()); //=> false
+ * RA.isSet(new Set()); //=> true
+ * RA.isSet(new Set([1,2]); //=> true
+ * RA.isSet(new Object()); //=> false
+ */
+
+const isSet = curry1(val => is(Set, val));
```
char0n: We're not using `R.is` in this library anymore. It's has a lot of problems. One of the main problem is described here: https://github.com/char0n/ramda-adjunct/issues/724

instead we prefer to use [tagChecking](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) which solves the problem. 

Pls aligned your implementation with `isDate` function - https://github.com/char0n/ramda-adjunct/blob/96d0772005ddf262ef2871ca5484dc156eb49805/src/isDate.js#L21

src/isSet.js
realSaddy: I think I have done so now

types/index.d.ts
realSaddy: I think I have done so now. I had to do `val is Set<any>` due to it requiring a type

types/test/isSet.ts
realSaddy: I think I have done so now

test/isSet.js
realSaddy: Moved it to the new standard. I am confused about testing for currying. I tried to check other examples where it was used, for example isNotBoolean, however, their tests don't have anything that jumps out as different from what I wrote. Could you elaborate if I got it incorrect? Thank you

src/isSet.js
realSaddy: Moved over to the pipe, type, identical standard

realSaddy:
Made changes to fix these issues; thank you for hosting this opportunity! If anything more needs to be changed, please tell me, and I'll get on it

types/index.d.ts
char0n: Yeah, that's good enought.

test/isSet.js
char0n: It's possible that I overlook this test `should support placeholder to specify "gaps"`. Now i can see it, so marking this as resolved.

char0n:
Nice job. Thank you for contributing!

PR #1116, opened 2019-10-13
feat(string): add trimStart #808

### LINKED ISSUE #808: trimStart/trimEnd
https://github.com/tc39/proposal-string-left-right-trim

Don't forget appropriate polyfills + aliases (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimStart)

rafaelpadovezi:
Closes #808


src/trimStart.js
```diff
@@ -0,0 +1,29 @@
...
+
+import polyfill from './internal/polyfills/String.trimStart';
+import isFunction from './isFunction';
+
+export const trimStartPolyfill = polyfill;
+
+export const trimStartInvoker = invoker(0, 'trimStart');
+
+/**
+ * Removes whitespace from the beginning of a string
```
char0n: string -> string. (dot at the end)

src/trimStart.js
```diff
@@ -0,0 +1,29 @@
+import { invoker } from 'ramda';
+
+import polyfill from './internal/polyfills/String.trimStart';
+import isFunction from './isFunction';
+
+export const trimStartPolyfill = polyfill;
+
+export const trimStartInvoker = invoker(0, 'trimStart');
```
char0n: No need to export invoker here

test/trimStart.js
```diff
@@ -0,0 +1,99 @@
...
+      });
+    });
+
+    context('given an string with whitespaces only', function() {
+      specify('should return an empty string', function() {
+        eq(trimStartPolyfill('   '), '');
+      });
+    });
+  });
+});
```
char0n: We're missings tests in for currying in every top level context

types/test/trimStart.ts
```diff
@@ -0,0 +1,10 @@
+import * as RA from 'ramda-adjunct';
+
+RA.trimStart('abc'); // $ExpectType string
```
char0n: Missing tests for `trimLeft` here. We can put them in this file directly.

src/trimStart.js
char0n: Taking this back. This is consistent with other part of the codebase (replaceAll)

rafaelpadovezi:
Thanks for the comments!

char0n:
Thanks for contributing !

Issue #1113, opened 2019-10-12
isMap


char0n:
**Is your feature request related to a problem? Please describe.**

Predicate for determining if a provided value is an instance of a Map.

**Describe the solution you'd like**

```js
RA.isMap(new Map()); //=> true
RA.isMap({ a: 1 }); //=> false
```

**Additional context**
[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `isMap` function. Issue original conceived in https://github.com/char0n/ramda-adjunct/issues/66

realSaddy:
Could I handle this?

char0n:
@realSaddy sure. Assigned to you

realSaddy:
Would you also like isNotMap or isNotSet?

char0n:
@realSaddy I can create issues for them, but not sure if they would be any value. Let's the those issues for vote and we'll find out if there are real world usecases for them.

char0n:
isNotSet: https://github.com/char0n/ramda-adjunct/issues/1123
isNotMap: https://github.com/char0n/ramda-adjunct/issues/1122

I've put them for voting to see a real value in them. Let's see.

PR #1120, opened 2019-10-14
feat(type): add isMap

realSaddy:
#1113


types/test/isMap.ts
```diff
@@ -0,0 +1,17 @@
+import * as RA from 'ramda-adjunct';
+
+RA.isSet(new Map()); // $ExpectType boolean
```
char0n: Tests here are testing `isSet` instead of `isMap`.

realSaddy: Oops; forgot to change that, sorry! I've updated it

PR #1127, opened 2019-10-19
feat(promise): add lastP

### LINKED ISSUE #816: lastP
This pattern is like [firstP([ .. ])](https://github.com/char0n/ramda-adjunct/issues/815), but only the latest fulfillment wins.

Promise patterns explained in [YDKN series](https://github.com/getify/You-Dont-Know-JS/blob/1st-ed/async%20%26%20performance/ch3.md#variations-on-all---and-race--)

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `lastP` function.

thomasmichaelwallace:
Closes #816

I really struggled to find a more "functional" approach to implementing `lastP` and would be greatful for the feedback.

I also made the following behaviour assumptions:
 * Failure responds with an array of rejection reasons
 * `lastP([])` should return `undefined`

That `lastP([]) === undefined` fell out of my implementation, which would return `rejectP([])`. I notice that this edge-case isn't often catered for (`anyP([])` seems to hang, for example?), so I wasn't sure how you wanted to handle it.

### DIFF
--- a/src/lastP.js
+++ b/src/lastP.js
@@ -1,0 +1,52 @@
+import { bind, curryN, ifElse, last, map, pipe, then } from 'ramda';
+
+import allP from './allP';
+import rejectP from './rejectP';
+import resolveP from './resolveP';
+import isEmptyArray from './isEmptyArray';
+
+/**
+ * Returns a promise that is fulfilled by the last given promise to be fulfilled,
+ * or rejected with an array of rejection reasons if all of the given promises are rejected.
+ *
+ * @func lastP
+ * @memberOf RA
+ * @category Function
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
+ * @sig [Promise a] -> Promise a
+ * @param {Iterable.<*>} iterable An iterable object such as an Array or String
+ * @return {Promise} Returns a promise that is fulfilled by the last given promise to be fulfilled,
+ * or rejected with an array of rejection reasons if all of the given promises are rejected.
+ * @see {@link RA.resolveP|resolveP}, {@link RA.rejectP|rejectP}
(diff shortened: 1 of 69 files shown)


thomasmichaelwallace:
Famous last words but it worked for me! I'll take a look at why those test cases are failing!

char0n:
Started to working on reviewing this PR.

src/lastP.js
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * RA.lastP([
+ *   1,
+ *   Promise.resolve('2-slow-promise-result'),
+ *   Promise.reject('3-very-slow-promise-result'),
+ *   Promise.resolve('4-fast-promise-result'),
+ * ]); //=> Promise('2-slow-promise-result')
+ */
+const lastP = curryN(1, iterable => {
+  const resolved = [];
```
char0n: I went ahead and tried to simplify your implementation. What would you say about following code ?
```js
const lastP = iterable => {
  const fulfilled = [];
  const rejected = [];
  const onFulfill = bind(fulfilled.push, fulfilled);
  const onReject = bind(rejected.push, rejected);

  const listOfPromises = map(
    p =>
      resolveP(p)
        .then(onFulfill)
        .catch(onReject),
    [...iterable]
  );

  return allP(listOfPromises).then(() => {
    if (lengthEq(0, fulfilled) && lengthEq(0, rejected)) {
      return undefined;
    }
    if (lengthGte(1, fulfilled)) {
      return last(fulfilled);
    }

    return rejectP(rejected);
  });
};
```
PS: it's a pseudocode, I did not run it.

thomasmichaelwallace: Much better than mine :)

PR #1131, opened 2019-10-20
feat(string): add padCharsEnd

### LINKED ISSUE #1104: padCharsEnd

This function pads the current string with a given string (repeated, if needed) so that the resulting string reaches a given length. The padding is applied from the end of the current string.

```js
RA.padCharsEnd('-', 3, 'a'); // => 'a--'
```

The function is of arity=3 and is autocurried.

**Additional context**

The functions should use [String.prototype.padEnd()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd) if available. If not polyfill like [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd#Polyfill) should be provided to compensate.

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `padCharsEnd` function.

dshubhadeep:
Closes #1104

### DIFF
--- a/src/internal/polyfills/String.padCharsEnd.js
+++ b/src/internal/polyfills/String.padCharsEnd.js
@@ -1,0 +1,21 @@
+const padCharsEnd = (padString, targetLength, value) => {
+  // eslint-disable-next-line no-bitwise
+  let finalLength = targetLength >> 0;
+  let finalPadString = String(
+    typeof padString !== 'undefined' ? padString : ' '
+  );
+
+  if (value.length > finalLength) {
+    return String(value);
+  }
+  finalLength -= value.length;
+  if (finalLength > finalPadString.length) {
+    finalPadString += finalPadString.repeat(
+      finalLength / finalPadString.length
+    );
+  }
+
+  return String(value) + finalPadString.slice(0, finalLength);
+};
+
+export default padCharsEnd;
--- a/src/internal/polyfills/String.padCharsEnd.js
+++ b/src/internal/polyfills/String.padCharsEnd.js
@@ -1,4 +1,7 @@
-const padCharsEnd = (padString, targetLength, value) => {
+import isFunction from '../../isFunction';
+import repeat from './String.repeat';
+
+const padCharsEndPolyfill = (padString, targetLength, value) => {
   // eslint-disable-next-line no-bitwise
   let finalLength = targetLength >> 0;
   let finalPadString = String(
@@ -10,12 +13,13 @@ const padCharsEnd = (padString, targetLength, value) => {
   }
   finalLength -= value.length;
   if (finalLength > finalPadString.length) {
-    finalPadString += finalPadString.repeat(
-      finalLength / finalPadString.length
-    );
+    const remainingLength = finalLength / finalPadString.length;
+    finalPadString += isFunction(String.prototype.repeat)
+      ? finalPadString.repeat(remainingLength)
+      : repeat(finalPadString, remainingLength);
   }
 
   return String(value) + finalPadString.slice(0, finalLength);
 };
 
-export default padCharsEnd;
+export default padCharsEndPolyfill;
--- a/src/padCharsEnd.js
+++ b/src/padCharsEnd.js
@@ -1,0 +1,36 @@
+import { curry, invoker, flip } from 'ramda';
+
+import polyfill from './internal/polyfills/String.padCharsEnd';
+import isFunction from './isFunction';
+
+export const padCharsEndStrPolyfill = curry(polyfill);
+
+export const padCharsEndStrInvoker = flip(invoker(2, 'padEnd'));
+
+/**
+ * The function pads the current string with a given string
+ * (repeated, if needed) so that the resulting string reaches a given length.
+ * The padding is applied from the end of the current string.
+ *
+ * @func padCharsEnd
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.11.0|v2.11.0}
+ * @category String
+ * @sig String -> Number -> String -> String
+ * @param {string} padString The string to pad the current string with
+ * @param {number} targetLength The length of the resulting string once
+ * the current string has been padded
+ * @param {string} value String value to be padded
+ * @return {string} A new string of the specified length with the pad string
+ * applied at the end of the current string
+ * @example
+ *
+ * RA.padCharsEnd('-', 3, 'a'); // => 'a--'
+ * RA.padCharsEnd('foo', 10, 'abc'); // => 'abcfoofoof'
+ * RA.padCharsEnd('123456', 6, 'abc'); // => 'abc123'
+ */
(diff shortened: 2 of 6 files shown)


src/internal/polyfills/String.padCharsEnd.js
```diff
@@ -0,0 +1,21 @@
+const padCharsEnd = (padString, targetLength, value) => {
```
char0n: padCharsEnd -> padCharsEndPolyfill

src/internal/polyfills/String.padCharsEnd.js
```diff
@@ -0,0 +1,21 @@
...
+  let finalPadString = String(
+    typeof padString !== 'undefined' ? padString : ' '
+  );
+
+  if (value.length > finalLength) {
+    return String(value);
+  }
+  finalLength -= value.length;
+  if (finalLength > finalPadString.length) {
+    finalPadString += finalPadString.repeat(
```
char0n: String.repeat may not be available, so we have two options here:

**1.** use always polyfill of [String.prototype.repeat](https://github.com/char0n/ramda-adjunct/blob/f1c5dfaac3b38a5599a6bb71a570205b025c7112/src/internal/polyfills/String.repeat.js#L50-L49) here that we already have in codebase

**2.** use polyfill of [String.prototype.repeat](https://github.com/char0n/ramda-adjunct/blob/f1c5dfaac3b38a5599a6bb71a570205b025c7112/src/internal/polyfills/String.repeat.js#L50-L49) only when `repeat` method on string is not available. 

I'll leave the decission and implementation up to you. IMHO best approach is second option

src/padCharsEnd.js
```diff
@@ -0,0 +1,36 @@
...
+export const padCharsEndStrInvoker = flip(invoker(2, 'padEnd'));
+
+/**
+ * The function pads the current string with a given string
+ * (repeated, if needed) so that the resulting string reaches a given length.
+ * The padding is applied from the end of the current string.
+ *
+ * @func padCharsEnd
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.11.0|v2.11.0}
```
char0n:
```js
@since {@link https://char0n.github.io/ramda-adjunct/2.22.0|v2.22.0}
```

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+  padCharsEndStrPolyfill,
+} from '../src/padCharsEnd';
+import eq from './shared/eq';
+
+describe('padCharsEnd', function() {
+  specify('should pad string with given padString', function() {
+    eq(RA.padCharsEnd('-', 3, 'a'), 'a--');
+  });
+
+  specify('should be curried', function() {
```
char0n: Usually we cant currying tests as last tests in the block. This is just a convention that we use.

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+      if (RA.isNotFunction(String.prototype.padEnd)) {
+        this.skip();
+      }
+    });
+
+    specify('should pad string with given padString', function() {
+      eq(padCharsEndStrInvoker('-', 3, 'a'), 'a--');
+    });
+
+    specify('should be curried', function() {
```
char0n: Usually we cant currying tests as last tests in the block. This is just a convention that we use.

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+  context('given targetLength less than equal to padString length', function() {
+    specify('should trim padString', function() {
+      eq(RA.padCharsEnd('123456', 6, 'abc'), 'abc123');
+      eq(RA.padCharsEnd('123456', 5, 'abc'), 'abc12');
+    });
+  });
+
+  context('given targetLength supplied as float', function() {
+    specify('should convert targetLength to integer', function() {
+      eq(RA.padCharsEnd('$', 7, 'abc'), 'abc$$$$');
```
char0n: We're passing an integer here instead of float

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+    });
+  });
+
+  context('given targetLength is not a number', function() {
+    specify('should return padded string', function() {
+      eq(RA.padCharsEnd('*', '5', 'abc'), 'abc**');
```
char0n: This look more like `given targetLength is a value coercible to valid integer number`

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+    });
+  });
+
+  context('given empty padString', function() {
```
char0n: `given padString set to undefined`

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+      }
+    );
+
+    context('given targetLength supplied as float', function() {
+      specify('should convert targetLength to integer', function() {
+        eq(padCharsEndStrInvoker('$', 7, 'abc'), 'abc$$$$');
+      });
+    });
+
+    context('given targetLength is not a number', function() {
```
char0n: This look more like `given targetLength is a value coercible to valid integer number`

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+      specify('should return padded string', function() {
+        eq(padCharsEndStrInvoker('*', '5', 'abc'), 'abc**');
+      });
+    });
+
+    context('given empty padString', function() {
```
char0n: `given padString set to undefined`

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
...
+      });
+    });
+  });
+
+  context('padCharsEndStrPolyfill', function() {
+    specify('should pad string with given padString', function() {
+      eq(padCharsEndStrPolyfill('-', 3, 'a'), 'a--');
+    });
+
+    specify('should be curried', function() {
```
char0n: Usually we cant currying tests as last tests in the block. This is just a convention that we use.

test/padCharsEnd.js
```diff
@@ -0,0 +1,185 @@
+import * as RA from '../src';
+import {
+  padCharsEndStrInvoker,
+  padCharsEndStrPolyfill,
+} from '../src/padCharsEnd';
+import eq from './shared/eq';
+
+describe('padCharsEnd', function() {
```
char0n: Can we add tests that demonstrates what happens when `targetLength` is a negative number ?

types/index.d.ts
```diff
@@ -1243,6 +1243,14 @@ declare namespace RamdaAdjunct {
          */
         allUnique<T>(list: T[]): boolean;
 
+        /**
```
char0n: The description should exactly match the description provided in `padCharsEnd.js`.
```js
* The function pads the current string with a given string
 * (repeated, if needed) so that the resulting string reaches a given length.
 * The padding is applied from the end of the current string.

Issue #878, opened 2019-04-13
replaceAll


char0n:
https://github.com/tc39/proposal-string-replaceall

Currently in stage 2. Should be fairly safe to implement.

char0n:
This has reached stage 3. Possibly to replace our custom implementation with future polyfill. One things was added there and that is handling of empty strings: https://github.com/tc39/proposal-string-replaceall#what-happens-if-searchvalue-is-the-empty-string

We need to make sure our implementation compensate for that.

js636f:
@char0n, so, we need this:

```
'xxx'.replaceAll('', '_');
// → '_x_x_x_'
```

now we have this:

```
RA.replaceAll('', '_', 'xxx');
// → 'x_x_x'
```

I think we may change this:

```
const replaceAll = curryN(3, (searchValue, replaceValue, str) => {
  if (isRegExp(searchValue)) {
    throw new Error('searchValue must be a String, not a RegExp.');
  }

  return pipe(
    split(String(searchValue)),
    join(String(replaceValue))
  )(str);
});

```
to something like this:

```
const replaceAll = curryN(3, (searchValue, replaceValue, str) => {
  if (isRegExp(searchValue)) {
    throw new Error('searchValue must be a String, not a RegExp.');
  }

  const callIfsearchValueIsEmpty = () => {
    return String(str).replace(/(?:)/g, String(replaceValue));
  }

  if (searchValue==='') 
    return callIfsearchValueIsEmpty();

  return pipe(
    split(String(searchValue)),
    join(String(replaceValue))
  )(str);

});
```


and add this to the test cases:

```
it('should correctly handle empty searchValue case', function() {
    const value = 'xxx';
    const actual = RA.replaceAll('', '_', value);
    const expected = '_x_x_x_';

    assert.strictEqual(actual, expected);
  });
```

If these ideas look reasonable - I'm ready to prepare and make a PR.

char0n:
@js636f your assumptions are correct. But we need a little bit different strategy. If we get this right we might be the first reference implementation. So let's do it in the following way:

**1.) Ponyfill**

 Create a ponyfil in our ponyfill directory. This ponyfill must not use anything from ramda or ramda-adjunct bust mu be pure javascript, so that it's framework agnostic. 

**2.) Implementation**

Then as usually, we check for String.prototype.replaceAll and use it if present, otherwise use ponyfill. 

[...]

PR #1141, opened 2019-10-23
test(replaceAll): make tests consistent with other string functions

char0n:
Ref #878


char0n:
@js636f FYI just made some consistency cleanup in tests.

js636f:
@char0n, OK, thanks for the info.

PR #1149, opened 2019-10-29
test(files): Use chai directly instead of internal eq function in test asserts, B#2

js636f:
Batch 2

test/

- contained.js
- curryRight.js
- curryRightN.js
- defaultWhen.js
- dispatch.js


js636f:
Thank you!

PR #1156, opened 2019-11-04
test(files): Use chai directly instead of internal eq function in test asserts, B#4

js636f:
Batch 4

test/
- floor.js
- hasPath.js
- inRange.js
- isArray.js
- isArrayLike.js


js636f:
@char0n, I hope I'll edit them all :-)

PR #1163, opened 2019-11-07
test(files): Use chai directly instead of internal eq function in test asserts, B#7

js636f:
Batch 7

test/

- isNotGeneratorFunction.js
- isNotNil.js
- isNotNull.js
- isNotNumber.js
- isNotObj.js
- isNotObjLike.js
- isNotPair.js

### DIFF
--- a/test/isNotObj.js
+++ b/test/isNotObj.js
@@ -1,33 +1,34 @@
+import { assert } from 'chai';
+
 import * as RA from '../src';
-import eq from './shared/eq';
 import args from './shared/arguments';
 import Symbol from './shared/Symbol';
 
 describe('isNotObj', function() {
   it('tests a value for complement of language type of `Object`', function() {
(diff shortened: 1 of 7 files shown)


test/isNotObj.js
```diff
@@ -1,33 +1,34 @@
...
-    eq(RA.isNotObj(undefined), true);
+    assert.isTrue(RA.isNotObj(null));
+    assert.isTrue(RA.isNotObj(undefined));
   });
 });
 
 describe('isNotObject', function() {
   it('tests an alias', function() {
-    eq(RA.isNotObj === RA.isNotObject, true);
+    assert.isTrue(RA.isNotObj === RA.isNotObject);
```
char0n: This should be assert.strictEqual

js636f: OK

PR #1177, opened 2019-11-15
test(files): Use chai directly instead of internal eq function in test asserts, B#13

js636f:
Batch 13

test/
- liftFN.js
- list.js
- mapIndexed.js
- mergePath.js
- mergePaths.js
- mergeProp.js
- mergeProps.js

### DIFF
--- a/test/liftFN.js
+++ b/test/liftFN.js
@@ -3,7 +3,6 @@ import { assert } from 'chai';
 import { Maybe } from 'monet';
 
 import * as RA from '../src';
-import eq from './shared/eq';
 import Identity from '../src/fantasy-land/Identity';
 
 const addN = (...args) => R.sum(args);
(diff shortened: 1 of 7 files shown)


char0n:
We're getting our self a little bit deeper into monadic waters here. We should not check deep equality on monads but rather use [Setoid](https://github.com/fantasyland/fantasy-land#setoid) laws of the monad. How to do that ? `R.equals` can help us with that. It's documentations sais: `Dispatches symmetrically to the equals methods of both arguments, if present.`. This means it know Setoid spec. I've created the POC how to compare monads here: https://github.com/char0n/ramda-adjunct/commit/1e9201d00797171d0e1fcb522ac1d4d06ca8bbcb. Check it out and align your code according to this.

```js
R.equals(Maybe.Some(1), Maybe.Some(1);
```
is equivalent to
```js
Maybe.Some(1).equals(Maybe.Some(1));
```

js636f:
@char0n, I edited my code. I hope I got everything right.

char0n:
Thanks, I'll check it out

char0n:
Thank you!

PR #1182, opened 2019-11-19
test(files): Use chai directly instead of internal eq function in test asserts, B#15

js636f:
Batch 15

test/
- noop.js
- notAllPass.js
- notBoth.js
- omitBy.js
- omitInIndexes.js

### DIFF
--- a/test/notAllPass.js
+++ b/test/notAllPass.js
@@ -23,7 +23,7 @@ describe('notAllPass', function() {
   });
 
   it('returns a curried function whose arity matches that of the highest-arity predicate', function() {
-    assert.strictEqual(RA.notAllPass([odd, gt5, plusEq]).length, 4);
+    assert.lengthOf(RA.notAllPass([odd, gt5, plusEq]), 4);
     assert.isTrue(RA.notAllPass([odd, gt5, plusEq])(9, 9, 9, 10));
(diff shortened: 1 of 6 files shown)


test/notAllPass.js
```diff
@@ -9,21 +10,21 @@ describe('notAllPass', function() {
...
   it('returns false on empty predicate list', function() {
-    eq(RA.notAllPass([])(3), false);
+    assert.isFalse(RA.notAllPass([])(3));
   });
 
   it('returns a curried function whose arity matches that of the highest-arity predicate', function() {
-    eq(RA.notAllPass([odd, gt5, plusEq]).length, 4);
-    eq(RA.notAllPass([odd, gt5, plusEq])(9, 9, 9, 10), true);
-    eq(RA.notAllPass([odd, gt5, plusEq])(9)(9)(9)(10), true);
+    assert.strictEqual(RA.notAllPass([odd, gt5, plusEq]).length, 4);
```
char0n: Checkout the [lengthOf](https://www.chaijs.com/api/assert/#method_lengthof) method of chai. More idiomatic to use it here IMHO

js636f:
@char0n, OK. I changed that file and a similar assertion in nonePass.js from Batch 14

PR #1195, opened 2019-11-27
test(files): Use chai directly instead of internal eq function in test asserts, B#21

js636f:
Batch 21

test/
- stubUndefined.js
- subtractNum.js
- trimEnd.js
- trimStart.js
- trunc.js
- viewOr.js
- weave.js
- weaveLazy.js
- Y.js

### DIFF
--- a/test/Y.js
+++ b/test/Y.js
@@ -1,5 +1,6 @@
+import { assert } from 'chai';
+
 import * as RA from '../src';
-import eq from './shared/eq';
 
 describe('Y', function() {
   it('tests for making a factorial function', function() {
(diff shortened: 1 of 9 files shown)


char0n:
is it the last one ?

js636f:
Yes, this is the last one. As I wrote, all files in the main folder are ready.

js636f:
Do I need to refactor tests in test/fantasy-land/ and test/internal/?

char0n:
Yep `internal` and `fantasy-land` are required also. The goal is to delete the `eq` function from the codebase.

js636f:
OK, I'll refactor that files too.

PR #1196, opened 2019-11-28
test(files): Use chai directly instead of internal eq function in test asserts, B#22

js636f:
Batch 22
- test/fantasy-land/Identity.js
- test/internal/ap.js

It seems to me eq is eliminated from all files.


char0n:
Can you pls (as part of this PR) remove the `eq` function from the codebase ? (I mean physically deleting it). Thank you

js636f:
OK, I deleted the eq.js file

Issue #624, opened 2018-06-24
Make all tests consistent + support for Ramda placeholder


char0n:
Currently our support for ramda placeholder is limited. Almost all binary functions are lacking the support for it. We need to continually provide full support for it. Every binary function needs to have the following test in it's test suite
```js
  it('should support placeholder to specify "gaps"', function() {
    const trimStart = RA.trimStart(R.__);

    assert.strictEqual(trimStart('  abc'), 'abc');
  });
```
All other functions needs to have following curring test which tests the combination of params invocation:
```js
  it('should curry', function() {
    assert.isTrue(RA.allEqualTo(1, [1, 1]));
    assert.isTrue(RA.allEqualTo(1)([1, 1]));
  });
```

As part of this, we should also refactor all tests to be consistent with gherkin style of testing. Here is an [example PR](https://github.com/char0n/ramda-adjunct/commit/8b8f8dff4d76d8512959b1142e3416c1bd4f5c4b) which demonstrates what we want. I already invested tens of hours into this effort, but need help to rectify these things in the rest of our tests.

char0n:
@js636f would would you say to help me with this effort ? Is it something you would be interested in ?

char0n:
@js636f thank you for doing great job! Last 3 days I've spent 40 hours so far to increase quality of our issues as part of https://github.com/char0n/ramda-adjunct/issues/1069. I've already transformed half of our issues. Feel free to take any one of them If you feel to and there is no need for you to always just work on this one. So...feel free to grab any issue with `Feature` and `Help wanted` label.

js636f:
@char0n, thank you! I think I'll close this issue first.

char0n:
@js636f how are we looking here? Are we done here?

char0n:
Closing this as we have last PR approved. @js636f great job here!

PR #1209, opened 2019-12-12
test(isNotObj): add ramda placeholder and make tests consitent

js636f:
Ref #624

### DIFF
--- a/src/isObj.js
+++ b/src/isObj.js
@@ -1,4 +1,5 @@
 import { both, either } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
 
 import isNotNull from './isNotNull';
 import isFunction from './isFunction';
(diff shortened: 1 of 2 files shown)


js636f:
@char0n, this PR is non ready yet. As I said in the comments,   we have a bug here. RA.isNotObj(R.__) returns false. I think the problem is in src/isObj.js because src/isNotObj.js is a complement(isObj) and RA.isObj(R.__) just returns true, so I think we need to refactor src/isObj.js to eliminate the bug. Do you have any ideas about possible refactoring?

char0n:
Implementing currying support for `isObj` should resolve it:

```js
import curry1 from 'ramda/src/internal/_curry1';

const isObj = curry1(both(isNotNull, either(isOfTypeObject, isFunction)));

js636f:
OK, thank you!

js636f:
Thank you!

PR #1215, opened 2019-12-17
test(isNotObjLike): add ramda placeholder and make tests consistent

js636f:
Ref #624

Batch 3

### DIFF
--- a/test/isNotObjLike.js
+++ b/test/isNotObjLike.js
@@ -43,7 +43,7 @@ describe('isNotObjLike', function() {
 });
 
 describe('isNotObjectLike', function() {
-  it('tests an alias', function() {
+  it('should be an alias for isNotObjLike', function() {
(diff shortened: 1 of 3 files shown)


test/isNotObjLike.js
```diff
@@ -1,33 +1,49 @@
...
+
+  it('should support placeholder to specify "gaps"', function() {
+    const isNotObjLike = RA.isNotObjLike(R.__);
+
+    assert.isTrue(isNotObjLike(null));
   });
 });
 
 describe('isNotObjectLike', function() {
   it('tests an alias', function() {
```
char0n: tests an alias -> should be alias for isNotObjLike

js636f: OK. And I found the same pattern in isNotObj.js and edited it too.

PR #1220, opened 2019-12-19
test(isNotPlainObj): add ramda placeholder and make tests consistent

js636f:
Batch 5

@char0n, as I know the function must be curried to support Ramda placeholders. The isNotPlainObj function is a complement(isPlainObj). [isPlainObj](https://github.com/char0n/ramda-adjunct/blob/master/src/isPlainObj.js) looks like a normal arrow function, but placeholder works with it. I can't find a piece of code where isPlainObj was curried (or perhaps I missed something). May you please explain this to me?

### DIFF
--- a/test/isNotPlainObj.js
+++ b/test/isNotPlainObj.js
@@ -1,4 +1,5 @@
 import { assert } from 'chai';
+import * as R from 'ramda';
 
 import * as RA from '../src';
 import element from './shared/element';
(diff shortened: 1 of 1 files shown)


char0n:
@js636f it's the `R.complement` which does the magic. It inspects the function that you're complementing and making it autocurried.

```js
const lift = _curry1(function lift(fn) {
  return liftN(fn.length, fn);
});

complement = lift(not);
```
Key part is `liftN` part. Did my explanation did the trick or should I collaborate further ?

js636f:
@char0n, thank you! I get it, but I have another question. Why R.complement didn't help us [here](https://github.com/char0n/ramda-adjunct/pull/1209), in isNotObj?

PR #1252, opened 2020-01-04

js636f:
Batch 9

Files:
test/
- isPair.js
- isPlainObj.js
- isPositive.js

src/
- isPlainObj.js


js636f:
@char0n, it seems to me all tests are OK, but test_node_9 failed because of some internal system error. Should I merge this PR or you'll restart test first?

char0n:
@js636f I've restarted the test. Yes this sometimes happens, I did not find a cause for this yes, but seems that it's a CircleCI issue with node. When tests fail, usualy restarting them on CI is a primary action you should seek for.

js636f:
@char0n, OK, thank you!

Issue #1103, opened 2019-10-11
padStart


char0n:
**Is your feature request related to a problem? Please describe.**

This function pads the current string with another string (multiple times, if needed) until the resulting string reaches the given length. The padding is applied from the start of the current string.

**Describe the solution you'd like**

```js
RA.padStart(3, 'a'); // => '  a'
```
The function is of arity=2 and is autocurried.

**Additional context**

This function is just a specialization of [padCharsStart](https://github.com/char0n/ramda-adjunct/issues/1102) and should/must use this function to implement it's behavior.

[This PR](https://github.com/char0n/ramda-adjunct/pull/1067) should serve as a model template of PR for `padStart` function.

jrga03:
can i have this issue?

char0n:
@jrga03 you would first need to implement https://github.com/char0n/ramda-adjunct/issues/1102 to process this. Write a comment in https://github.com/char0n/ramda-adjunct/issues/1102 so that I can assigned you there.

char0n:
@jrga03 already taken by different developer, it would be possible to execute this issue when #1102 is landed on master.

jrga03:
@char0n alright, will wait for #1102

char0n:
@jrga03 meanwhile I can offer you [RA.isSet](https://github.com/char0n/ramda-adjunct/issues/1110) ? Would it be acceptable for you ?

char0n:
@jrga03 go for #1102 if you wish you. There has been no activity from original issue requester.

char0n:
@jrga03 issue is in master. Go ahead with the implementation if you're still interested.

PR #1287, opened 2020-01-26
feat: add padStart

kitasuna:
#1103

Add padStart / types / tests

### DIFF
--- a/src/padStart.js
+++ b/src/padStart.js
@@ -1,9 +1,7 @@
 import padCharsStart from './padCharsStart';
 
 /**
- * The function pads the current string with an empty string
- * so that the resulting string reaches a given length.
- * The padding is applied to the beginning of the current string.
+ * Pads string on the left side if it's shorter than length.
  *
  * @func padStart
  * @memberOf RA
--- a/src/padStart.js
+++ b/src/padStart.js
@@ -1,0 +1,25 @@
+import padCharsStart from './padCharsStart';
+
+/**
+ * The function pads the current string with an empty string
+ * so that the resulting string reaches a given length.
+ * The padding is applied to the beginning of the current string.
+ *
+ * @func padStart
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.25.0|v2.25.0}
+ * @category String
+ * @sig Number -> String -> String
+ * @param {number} targetLength The length of the resulting string once
+ * the current string has been padded
+ * @param {string} value String value to be padded
+ * @return {string} A new string of the specified length with the empty string
+ * applied to the beginning of the current string
+ * @see {@link RA.padCharsEnd|padCharsEnd}, {@link RA.padCharsStart|padCharsStart}, {@link RA.padEnd|padEnd}
+ * @example
+ *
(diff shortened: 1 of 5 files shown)


src/padStart.js
```diff
@@ -0,0 +1,25 @@
+import padCharsStart from './padCharsStart';
+
+/**
+ * The function pads the current string with an empty string
```
char0n: What do you say to the following: `Pads string on the left side if it's shorter than length. Padding characters are truncated if they exceed length.`

test/padStart.js
```diff
@@ -0,0 +1,47 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe('padStart', function() {
+  specify('should pad string with spaces', function() {
```
char0n: specify -> it

test/padStart.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+  );
+
+  context('given targeLength is a negative number', function() {
+    specify('should return original string', function() {
+      assert.strictEqual(RA.padStart(-10, 'abc'), 'abc');
+    });
+  });
+
+  specify('should curry', function() {
```
char0n: specify -> it

types/index.d.ts
```diff
@@ -1331,6 +1331,14 @@ declare namespace RamdaAdjunct {
         padCharsStart(padString: string, targetLength: number, value: string): string;
         padCharsStart(padString: string, targetLength: number): (value: string) => string;
         padCharsStart(padString: string): (targetLength: number, value: string) => string;
+
+        /**
```
char0n: Please align this description with the final description suggested above: https://github.com/char0n/ramda-adjunct/pull/1287/files#diff-8322d64eab92ea5ea1ca5d559ab14f97R4

src/padStart.js
kitasuna: Thanks for the feedback! One clarification on this bit:

> Padding characters are truncated if they exceed length.

What would be an example of this happening?

char0n: You're right! We can drop this last sentence, because that is specific to `padCharsStart` and not to `padStart`

test/padStart.js
char0n: We only use `specify` inside `context`. In other cases `it` is used inside `describe`.

src/padStart.js
kitasuna: Ok, cool. I was concerned the implementation was missing something after reading your initial comment 😅. Will update shortly. Thanks again!

kitasuna:
Comments addressed! btw, originally I had stolen the descriptions from the `padEnd` function and modified as necessary. If you want, I'm happy to align the desc on `padEnd` as well, to something like `Pads string on the right side if it's shorter than length.` Just let me know 👍

char0n:
> Comments addressed! btw, originally I had stolen the descriptions from the padEnd function and modified as necessary. If you want, I'm happy to align the desc on padEnd as well, to something like Pads string on the right side if it's shorter than length. Just let me know +1

That a great catch. We should definitely make them consistent.

PR #1291, opened 2020-01-28

js636f:
Batch 16

Files:

test/
- liftF.js
- liftFN.js
- list.js

src/
- liftF.js

### DIFF
--- a/test/liftFN.js
+++ b/test/liftFN.js
@@ -66,7 +66,7 @@ describe('liftFN', function() {
     );
   });
 
-  it('retain order of arguments', function() {
+  it('retains order of arguments', function() {
     assert.isTrue(
       R.equals(
         RA.liftFN(3, add3)(Maybe.Some('a'), Maybe.Some('b'), Maybe.Some('c')),
@@ -85,7 +85,7 @@ describe('liftFN', function() {
     );
   });
 
-  it('is curried', function() {
+  it('should be curried', function() {
     const f4 = RA.liftFN(4);
     assert.strictEqual(typeof f4, 'function');
     assert.isTrue(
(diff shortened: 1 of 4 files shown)


js636f:
@char0n, I edited some descriptions in the test/list.js file, but I'm not sure about currying. It seems to me this is a very special function because it supports unlimited list of arguments and returns something useful even when called without any arguments. I don't know how to curry it properly. If it have to be curried anyway - give me a hint please.

test/liftFN.js
```diff
@@ -66,7 +66,7 @@ describe('liftFN', function() {
     );
   });
 
-  it('retain order of arguments', function() {
+  it('retains order of arguments', function() {
```
char0n: should retain order of arguments

test/list.js
```diff
@@ -3,19 +3,19 @@ import { assert } from 'chai';
 import * as RA from '../src';
 
 describe('list', function() {
-  it('tests creating list from items', function() {
+  it('creates list from items', function() {
```
char0n: For this whole file here, we have a convention to always start a test description with a `should` word.

`creates list from another list` -> `should create list from another list`

test/liftFN.js
```diff
@@ -26,7 +26,7 @@ describe('liftFN', function() {
     );
   });
 
-  it('throws error on variadic function is more arguments than arity', function() {
+  it('should throw an error on variadic function which have more arguments than its arity', function() {
```
char0n: can we pls break this into 
```js
context('given variadic function with more arguments than its arity', function() {
  specify('should trow an error', function() {
   ...
  });
});

PR #1294, opened 2020-01-29
Compact types

dimitropoulos:
before:

Screenshot_20200129_074136

after:

Screenshot_20200129_074109

When https://github.com/microsoft/TypeScript/pull/29317 lands, this will be able to be further improved to remove `false`, `0`, `""`, and `NaN`.  If you want to make an issue to check up on that when that happens please do.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1150,7 +1150,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact<T>(list: Array<T>): NonNullable<T>[];
+        compact<T>(list: T[]): Array<NonNullable<T>>;
 
         /**
          * Returns a new list containing the contents of the given list, followed by the given
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1150,7 +1150,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact(list: any[]): any[];
+        compact<T>(list: T[]): NonNullable<T>[];
 
         /**
          * Returns a new list containing the contents of the given list, followed by the given
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1150,7 +1150,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact<T>(list: T[]): NonNullable<T>[];
+        compact<T>(list: Array<T>): NonNullable<T>[];
 
         /**
          * Returns a new list containing the contents of the given list, followed by the given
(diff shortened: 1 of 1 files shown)


char0n:
Thank you for your contribution. Currently the build is failing on `ERROR: 1153:32  array-type  Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.`. 

Can you please align with that ? In order to verify this locally just run `npm run lint` which will list your changes made in typescript files.

dimitropoulos:
I was confused by this because `T[]` is found _all over_ that file.  By the way I happen to disagree with the rule but that's irrelevant.

I ran `npm run lint` before making any change and got:
```
$ npm run lint

> ramda-adjunct@2.25.0 lint /home/dimitri/src/github.com/char0n/ramda-adjunct
> better-npm-run lint

running better-npm-run in /home/dimitri/src/github.com/char0n/ramda-adjunct
Executing script: lint

to be executed: eslint ./ && tslint --project types/tsconfig.json -c types/tslint.json types/**/*.ts 
```

Could it be that something is off?  Test flake??

Is it asking for me to do `Array<NonNullable<T>>`?

char0n:
Right, so the linting works. I dug deeper and the problem is with the tests of the typings. If you run `npm run test:types` you'll see the error. IMHO if you amend the signature from
```ts
compact<T>(list: T[]): NonNullable<T>[];
```
to 
```ts
compact<T>(list: Array<T>): NonNullable<T>[];
```
then we should be fine. I don't have strong opinions about TypeScript as I'm not doing on daily basis, that's why ramda-adjunct adopts [dtslint](https://github.com/microsoft/dtslint) which is opinionated enough drives consistency of our typescript code.

types/index.d.ts
```diff
@@ -1150,7 +1150,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact(list: any[]): any[];
+        compact<T>(list: T[]): NonNullable<T>[];
```
```suggestion
        compact<T>(list: Array<T>): NonNullable<T>[];
```

dimitropoulos:
well.  cool.  I learned something!  Thanks, as always for the great project.  I'm always happy to contribute.

That said: sure does seem like something is wrong here?
Screenshot_20200130_085322

types/index.d.ts
```diff
@@ -1150,7 +1150,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact(list: any[]): any[];
+        compact<T>(list: Array<T>): NonNullable<T>[];
```
```suggestion
        compact<T>(list: T[]): Array<NonNullable<T>>;
```

char0n:
Yeah, that looks crazy, I had to checkout inside your branch to understand what is going on. I've suggested a change that finally works and tested it locally that it really works.

types/index.d.ts
dimitropoulos: cool, yeah, that's what I was trying to say in https://github.com/char0n/ramda-adjunct/pull/1294#issuecomment-579824632

sounds good.

char0n:
Took us some time to figure it out but we made it. Thank you!

PR #1320, opened 2020-02-20

js636f:
Batch 19

Files:
test/
- move.js
- nand.js
- neither.js
- nonePass.js
- noop.js

src/
- neither.js
- nonePass.js

### DIFF
--- a/src/neither.js
+++ b/src/neither.js
@@ -1,5 +1,4 @@
-import { compose, complement, either } from 'ramda';
-import curry2 from 'ramda/src/internal/_curry2';
+import { curry, compose, complement, either } from 'ramda';
 
 /* eslint-disable max-len */
 /**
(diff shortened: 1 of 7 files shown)


src/neither.js
```diff
@@ -1,4 +1,5 @@
 import { compose, complement, either } from 'ramda';
+import curry2 from 'ramda/src/internal/_curry2';
```
char0n: Can we use `curry` or `curryN` instead of internal `curry2`. We only use `curry1` from internals and probably get rid of it in the future and replace it with `curryN` so that we're dependant on internal api or ramda as few as possible.

src/neither.js
```diff
@@ -33,6 +34,6 @@ import { compose, complement, either } from 'ramda';
  * f(9); //=> true
  */
 /* eslint-enable max-len */
-const neither = compose(complement, either);
+const neither = curry2(compose(complement, either));
```
char0n: curry or curryN

src/neither.js
js636f: OK, I used curry instead of curry2

Issue #1242, opened 2019-12-25
sortByProps


char0n:
**Is your feature request related to a problem? Please describe.**

Sort a List by array of props (if first prop equivalent, sort by second, etc.)

**Describe the solution you'd like**

```js
const firstTruthy = ([head, ...tail]) => R.reduce(R.either, head, tail);
const makeComparator = (propName) => R.comparator((a, b) => R.gt(R.prop(propName, a), R.prop(propName, b)));
const sortByProps = (props, list) => R.sort(firstTruthy(R.map(makeComparator, props)), list)

sortByProps(["a","b","c"], [{a:1,b:2,c:3}, {a:10,b:10,c:10}, {a:10,b:6,c:0}, {a:1, b:2, c:1}, {a:100}])

//=> [{a:100}, {a:10,b:10,c:10}, {a:10,b:6,c:0}, {a:1,b:2,c:3}, {a:1, b:2, c:1}]
```

**Describe alternatives you've considered**

--

**Additional context**

This is a [ramda cookbook](https://github.com/ramda/ramda/wiki/Cookbook#sort-a-list-by-array-of-props-if-first-prop-equivalent-sort-by-second-etc) function.
Originally mentioned in https://github.com/char0n/ramda-adjunct/issues/945

Issue #1240, opened 2019-12-25
sortByPath


char0n:
**Is your feature request related to a problem? Please describe.**

Sorts the list according to the supplied path

**Describe the solution you'd like**

```js
const sortByFirstItem = R.sortByPath([0]);
const pairs = [[-1, 1], [-2, 2], [-3, 3]];
sortByFirstItem(pairs); //=> [[-3, 3], [-2, 2], [-1, 1]]

const sortByName = R.sortByPath(['name']);
const alice = {
  name: 'ALICE',
  age: 101
};
const bob = {
  name: 'Bob',
  age: -10
};
const clara = {
  name: 'clara',
  age: 314.159
};
const people = [clara, bob, alice];
sortByName(people); //=> [alice, bob, clara]
```

**Describe alternatives you've considered**

---

**Additional context**

Original issue: https://github.com/char0n/ramda-adjunct/issues/945

Mxchaeltrxn:
Hey I can try do this

Mxchaeltrxn:
@char0n This seems very similar to sortByProps except that this also works for lists of lists. What exactly is meant by 'path'?

char0n:
@Mxchaeltrxn I've assigned it to you. But first let's focus on https://github.com/char0n/ramda-adjunct/pull/1323. There is still lot of work to be done there. After we're finished there, you'll already know how submit another PR. Thanks.

char0n:
@Mxchaeltrxn there is nothing stopping you to start working on this. Go ahead and good luck!

Mxchaeltrxn:
@char0n I'm a little confused as to why this function partly does the same thing as sortByProps. It makes more sense to me to make it work for lists of lists only because the other functionality was just written

char0n:
@Mxchaeltrxn good point. I've modified the issue description to reflect your comment.

PR #1323, opened 2020-02-26
Add sort by props func

Mxchaeltrxn:
Addresses #1242.

I've written the function based on the ramda cookbook and wrote tests

### DIFF
--- a/src/sortByProps.js
+++ b/src/sortByProps.js
@@ -9,7 +9,7 @@ import { reduce, either, curry, prop, sort, comparator, map, lt } from 'ramda';
  * @category List
  * @sig [k] -> [{k: v}] -> [{k: v}]
  * @param {Array.<string>} props A list of properties in the list param to sort by
- * @param {Array.<string>} list A list of objects to be sorted
+ * @param {Array.<object>} list A list of objects to be sorted
  * @return {Array.<object>} A new list sorted by the properties in the props param
  * @example
  *
--- a/src/sortByProps.js
+++ b/src/sortByProps.js
@@ -1,0 +1,31 @@
+import { reduce, either, curry, prop, sort, comparator, map, lt } from 'ramda';
+
+/**
+ * Sort a list of objects by a list of props (if first prop value is equivalent, sort by second, etc).
+ *
+ * @func sortByProps
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.26.0|v2.26.0}
+ * @category List
+ * @sig [k] -> [{k: v}] -> [{k: v}]
+ * @param {Array.<string>} props A list of properties in the list param to sort by
+ * @param {Array.<string>} list A list of objects to be sorted
+ * @return {Array.<object>} A new list sorted by the properties in the props param
+ * @example
+ *
+ * sortByProps(['num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2} {num: 3}]
+ * sortByProps(['letter', 'num'], [{num: 3, letter: 'a'}, {num: 2, letter: 'a'} {num: 1, letter: 'z'}])
+ * //=> [ {num: 2, letter: 'a'}, {num: 3, letter: 'a'}, {num: 1, letter: 'z'}]
+ * sortByProps(['name', 'num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2}, {num: 3}]
+ */
+
+const sortByProps = curry((props, list) => {
+  const firstTruthy = ([head, ...tail]) => reduce(either, head, tail);
+  const makeComparator = propName =>
+    comparator((a, b) => lt(prop(propName, a), prop(propName, b)));
+  return sort(firstTruthy(map(makeComparator, props)), list);
+});
+
+export default sortByProps;
--- a/test/sortByProps.js
+++ b/test/sortByProps.js
@@ -1,0 +1,75 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe('sortByProps', function() {
+  const alice = {
+    name: 'ALICE',
+    age: 101,
+    lastName: 'Smith',
+  };
+  const bob = {
+    name: 'Bob',
+    age: -10,
+    lastName: 'Smith',
+  };
+  const clara = {
+    name: 'clara',
+    age: 314.159,
+    lastName: 'Kris',
+  };
+  const people = [clara, bob, alice];
+
+  context('given an array of objects', function() {
+    specify('should curry', function() {
+      assert.deepEqual(
+        RA.sortByProps(['name'])(people),
+        RA.sortByProps(['name'], people)
+      );
+    });
+
+    specify(
+      'should sort by property if property to sort by exists',
+      function() {
+        assert.deepEqual(RA.sortByProps(['name'])(people), [alice, bob, clara]);
+      }
+    );
+
+    specify(
+      'should return original list if property to sort by does not exist',
+      function() {
+        assert.deepEqual(RA.sortByProps(['p'])(people), people);
+      }
+    );
+
+    specify('should sort by multiple properties', function() {
+      assert.deepEqual(RA.sortByProps(['lastName', 'name'])(people), [
+        clara,
+        alice,
+        bob,
+      ]);
+    });
+
+    specify('should ignore properties that do not exist', function() {
+      assert.deepEqual(RA.sortByProps(['lastName', 'p', 'name'])(people), [
+        clara,
+        alice,
+        bob,
+      ]);
+    });
+  });
+
+  context('not given an array of objects', function() {
+    specify('', function() {
+      assert.deepEqual(RA.sortByProps(['name'])(['a', 'b', 'c']), [
+        'a',
+        'b',
+        'c',
+      ]);
+    });
+
+    specify('should return error', function() {
+      assert.deepEqual(RA.sortByProps(['name'])(true), []);
+    });
+  });
+});
--- a/test/sortByProps.js
+++ b/test/sortByProps.js
@@ -20,30 +20,26 @@ describe('sortByProps', function() {
   };
   const people = [clara, bob, alice];
 
-  context('given an array of objects', function() {
-    specify('should curry', function() {
+  context('given a list of objects', function() {
+    it('should be curried', function() {
       assert.deepEqual(
         RA.sortByProps(['name'])(people),
         RA.sortByProps(['name'], people)
       );
     });
 
-    specify(
-      'should sort by property if property to sort by exists',
-      function() {
-        assert.deepEqual(RA.sortByProps(['name'])(people), [alice, bob, clara]);
-      }
-    );
+    context('given property to sort by does not exist', function() {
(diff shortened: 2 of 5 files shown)


char0n:
Thank you for your contribution! I'll make a PR during tomorrow.

char0n:
Before I go into deep code review, please have a look at this [PR](https://github.com/char0n/ramda-adjunct/pull/1287/files). It's model PR and demonstrates what you need to add to the project for a successfull PR. Please try to align your contribution with this. Thank you!

Mxchaeltrxn:
@char0n Done. I'm not quite sure why circleci tests are failing though. They seem to be failing in the master branch too. Should this be an issue?

char0n:
> @char0n Done. I'm not quite sure why circleci tests are failing though. They seem to be failing in the master branch too. Should this be an issue?

Flakiness of the CirleCI. Don't worry I'll restart the failing tests on node 12.

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
+import { reduce, either, curry, prop, sort, comparator, map, lt } from 'ramda';
+
+import isNotArray from './isNotArray';
+
+/**
+ * Sort an list of objects by a list of props (if first prop value is equivalent, sort by second, etc).
```
char0n:
```suggestion
 * Sort a list of objects by a list of props (if first prop value is equivalent, sort by second, etc).
```

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @func sortByProps
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.26.0|v2.26.0}
+ * @category List
+ * @sig [{k: v}] -> [String] -> [{k: v}]
```
char0n: You signature looks like the function consumes data first, which is not the case. More proper signature would be:
```suggestion
 * @sig [k] -> [{k: v}] -> [{k: v}]
```

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+ * sortByProps(['num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2} {num: 3}]
+ * sortByProps(['letter', 'num'], [{num: 3, letter: 'a'}, {num: 2, letter: 'a'} {num: 1, letter: 'z'}])
+ * //=> [ {num: 2, letter: 'a'}, {num: 3, letter: 'a'}, {num: 1, letter: 'z'}]
+ * sortByProps(['name', 'num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2}, {num: 3}]
+ */
+
+const sortByProps = curry((list, props) => {
+  if (isNotArray(list) || isNotArray(props)) {
```
char0n: We expect here that the function is going to be total, so we're not checking the arguments. The function is supposed to consume two lists and return a list. That defines it's totality. Please drop this check.

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Sort an list of objects by a list of props (if first prop value is equivalent, sort by second, etc).
+ *
+ * @func sortByProps
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.26.0|v2.26.0}
+ * @category List
+ * @sig [{k: v}] -> [String] -> [{k: v}]
+ * @param {[{k: v}]} list A list of objects to be sorted
```
char0n:
```suggestion
 * @param Array.<string> props A list of properties to sort by
```

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+ * @param {[string]} props A list of properties in the list param to sort by
```
char0n:
```suggestion
 * @param Array.<Object> list A list of objects to sort
```

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+ * @return {[{k: v}]} A new list sorted by the properties in the props param
```
char0n:
```suggestion
 * @return Array.<Object> A new list sorted by the properties
```

src/sortByProps.js
char0n: These are [jsdoc](https://jsdoc.app/) params and return. They don't share the same rules as [@sig](https://github.com/ramda/ramda/wiki/Type-Signatures)

src/sortByProps.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * sortByProps(['num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2} {num: 3}]
+ * sortByProps(['letter', 'num'], [{num: 3, letter: 'a'}, {num: 2, letter: 'a'} {num: 1, letter: 'z'}])
+ * //=> [ {num: 2, letter: 'a'}, {num: 3, letter: 'a'}, {num: 1, letter: 'z'}]
+ * sortByProps(['name', 'num'], [{num: 3}, {num: 2}, {num: 1}])
+ * //=> [{num: 1}, {num: 2}, {num: 3}]
+ */
+
+const sortByProps = curry((list, props) => {
```
char0n: You need to switch the functions arguments. In Functional programming data is always the last argument
```suggestion
const sortByProps = curry((props, list) => {
```

char0n: Your examples are fine though

test/sortByProps.js
```diff
@@ -0,0 +1,57 @@
...
+  const clara = {
+    name: 'clara',
+    age: 314.159,
+    lastName: 'Kris',
+  };
+  const people = [clara, bob, alice];
+
+  context('given an array of objects', function() {
+    specify(
+      'property to sort by exists, should return sorted values',
```
char0n:
```suggestion
      'should sort by name property',
```

types/index.d.ts
```diff
@@ -2,1342 +2,1454 @@ declare var RA: RamdaAdjunct.Static;
 // TypeScript Version: 2.4
 
 declare namespace RamdaAdjunct {
```
char0n: You somehow reformatted the entire file and finding out what changed is not impossible. Please return back the formatting and make only necessary changes related to your PR.

types/test/sortByProps.ts
```diff
@@ -0,0 +1,23 @@
...
+  lastName: 'Smith',
+};
+const clara = {
+  name: 'clara',
+  age: 314.159,
+  lastName: 'Kris',
+};
+const people = [clara, bob, alice];
+
+RA.sortByProps(people)(['name']); // $ExpectType object[]
```
char0n: Parameters should be reversed and we're missing at least one test that is giving an error. Example: https://github.com/char0n/ramda-adjunct/blob/697fa142fde08a8d53ed2c3eda75ffe58ade217c/types/test/ceil.ts#L12

test/sortByProps.js
```diff
@@ -0,0 +1,57 @@
...
+
+    specify('ignore properties that do not exist', function() {
+      assert.deepEqual(RA.sortByProps(people)(['lastName', 'p', 'name']), [
+        clara,
+        alice,
+        bob,
+      ]);
+    });
+  });
+});
```
char0n: We're missing tests for currying. Example https://github.com/char0n/ramda-adjunct/blob/16b8831510a76c3f11c6480735710ad46d976983/test/divideNum.js#L14

test/sortByProps.js
```diff
@@ -0,0 +1,57 @@
...
+      function() {
+        assert.deepEqual(RA.sortByProps(people)(['lastName', 'name']), [
+          clara,
+          alice,
+          bob,
+        ]);
+      }
+    );
+
+    specify('ignore properties that do not exist', function() {
```
char0n:
```suggestion
    specify('should ignore properties that do not exist', function() {
```

char0n: The rule in tests is to always start with world `should`. That way tests are consistent and easy to read because you expect the senteces in standard format.

Mxchaeltrxn:
@char0n So I've implemented the changes you asked for. Let me know if anything else needs to be changed.

types/index.d.ts
```diff
@@ -1337,6 +1337,12 @@ declare namespace RamdaAdjunct {
          */
         padStart(targetLength: number, value: string): string;
         padStart(targetLength: number): (value: string) => string;
+
+        /**
+         * Sort list of objects by an array of properties
```
```suggestion
         * Sort a list of objects by a list of props (if first prop value is equivalent, sort by second, etc).

```

char0n: Let's unify description here with description in `sortByProps.js`

test/sortByProps.js
```diff
@@ -0,0 +1,75 @@
+import { assert } from 'chai';
```
char0n: This is my suggestion for the tests:
```js
import { assert } from 'chai';

import * as RA from '../src';

describe('sortByProps', function() {
  const alice = {
    name: 'ALICE',
    age: 101,
    lastName: 'Smith',
  };
  const bob = {
    name: 'Bob',
    age: -10,
    lastName: 'Smith',
  };
  const clara = {
    name: 'clara',
    age: 314.159,
    lastName: 'Kris',
  };
  const people = [clara, bob, alice];

  context('given list of objects', function() {
    specify('should sort by existing property', function() {
      assert.deepEqual(RA.sortByProps(['name'], people), [alice, bob, clara]);
    });

    context('given propertly to srot by does not exist', function() {
      specify('should return original list', function() {
        assert.deepEqual(RA.sortByProps(['p'], people), people);
      });
    });

    specify('should sort by multiple properties', function() {
      assert.deepEqual(RA.sortByProps(['lastName', 'name'], people), [
        clara,
        alice,
        bob,
      ]);
    });

    specify('should ignore properties that do not exist', function() {
      assert.deepEqual(RA.sortByProps(['lastName', 'p', 'name'], people), [
        clara,
        alice,
        bob,
      ]);
    });
  });

  context('given list of strings', function() {
    specify('should return unmodified list of strings', function() {
      assert.deepEqual(RA.sortByProps(['name'], ['a', 'b', 'c']), [
        'a',
        'b',
        'c',
      ]);
    });
  });

  context('given empty list', function() {
    specify('should return empty list', function() {
      assert.deepEqual(RA.sortByProps(['name'], []), []);
    });
  });

  it('should be curried', function() {
    assert.deepEqual(
      RA.sortByProps(['name'])(people),
      RA.sortByProps(['name'], people)
    );
  });
});
```

Please compensate into your tests.

Mxchaeltrxn:
@char0n Thanks for your detailed commentary! Things are making more sense to me now.

Also, just a gentle reminder to check out #1240 if this one is done.

types/index.d.ts
```diff
@@ -1337,6 +1337,13 @@ declare namespace RamdaAdjunct {
          */
         padStart(targetLength: number, value: string): string;
         padStart(targetLength: number): (value: string) => string;
+
+        /**
+         * Sort a list of objects by a list of props (if first prop value is equivalent, sort by second, etc).
+
```
```suggestion
```

types/index.d.ts
```diff
@@ -1337,6 +1337,13 @@ declare namespace RamdaAdjunct {
...
+         */
```
```suggestion
         */
```

Issue #1236, opened 2019-12-24
isIndexed


char0n:
**Is your feature request related to a problem? Please describe.**

To determine if something is an indexed data type. We can start assuming that value is indexed if it's either **String** or **Array** type.

**Describe the solution you'd like**

```js
isIndexed([1]); //=> true
isIndexed('test'); //=> true
```

**Describe alternatives you've considered**
-- 

**Additional context**

Originally conceived in https://github.com/char0n/ramda-adjunct/issues/66

damiangreen:
I'm having a crack at this one

char0n:
Go ahead and good luck!

PR #1337, opened 2020-03-09
Is indexed

damiangreen:
Add `isIndexed`. Addresses https://github.com/char0n/ramda-adjunct/issues/1236

### DIFF
--- a/src/isIndexed.js
+++ b/src/isIndexed.js
@@ -1,0 +1,25 @@
+import { either } from 'ramda';
+import curry1 from 'ramda/src/internal/_curry1';
+
+import isArray from './isArray';
+import isString from './isString';
+
+/**
+ * Determine if input value is an indexed data type.
+ *
+ * @func isIndexed
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.26.0|v2.26.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @example
+ *
+ * RA.isIndexed([1]) //=> true
+ * RA.isIndexed('test') //=> true
+ */
+
+const isIndexed = curry1(either(isString, isArray));
+
+export default isIndexed;
--- a/test/isIndexed.js
+++ b/test/isIndexed.js
@@ -1,0 +1,37 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import * as RA from '../src';
+
+describe('isIndexed', function() {
+  context('given a `String` value', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isIndexed('test'));
+      assert.isTrue(RA.isIndexed(Object('test')));
+      assert.isTrue(RA.isIndexed(''));
+    });
+  });
+
+  context('given array value', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isIndexed([]));
+      assert.isTrue(RA.isIndexed([1, 2]));
+    });
+  });
+
+  context('given a non `String` or `Array` value', function() {
+    specify('should return false', function() {
+      assert.isFalse(RA.isIndexed(true));
+      assert.isFalse(RA.isIndexed(new Date()));
+      assert.isFalse(RA.isIndexed(new Error()));
+      assert.isFalse(RA.isIndexed({ 0: 1, length: 1 }));
+      assert.isFalse(RA.isIndexed(1));
+    });
+  });
+
+  it('should support placeholder to specify "gaps"', function() {
+    const isIndexed = RA.isIndexed(R.__);
+
+    assert.isTrue(isIndexed([]));
+  });
+});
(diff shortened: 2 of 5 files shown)


src/isIndexed.js
```diff
@@ -0,0 +1,22 @@
+import { either } from 'ramda';
+
+import isArray from './isArray';
+import isString from './isString';
+/**
```
char0n:
```suggestion

/**
```

src/isIndexed.js
```diff
@@ -0,0 +1,22 @@
...
+ * Determine if input value is an indexed data type
```
char0n:
```suggestion
 * Determine if input value is an indexed data type.
```

test/isIndexed.js
```diff
@@ -0,0 +1,28 @@
...
+  context('given a `String` value', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isIndexed('test'));
+      assert.isTrue(RA.isIndexed(Object('test')));
+    });
+  });
+
+  context('given array value', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isIndexed([1, 2]));
```
char0n: Let's add empty array check
```suggestion
      assert.isTrue(RA.isIndexed([]));
      assert.isTrue(RA.isIndexed([1, 2]));
```

test/isIndexed.js
```diff
@@ -0,0 +1,28 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe('isIndexed', function() {
+  context('given a `String` value', function() {
+    specify('should return true', function() {
+      assert.isTrue(RA.isIndexed('test'));
+      assert.isTrue(RA.isIndexed(Object('test')));
```
char0n: Let's add empty string check
```suggestion
      assert.isTrue(RA.isIndexed(Object('test')));
      assert.isTrue(RA.isIndexed(''));
```

test/isIndexed.js
```diff
@@ -0,0 +1,28 @@
...
+  context('given a non `String` or `Array` value', function() {
+    specify('should return false', function() {
+      assert.isFalse(RA.isIndexed(true));
+      assert.isFalse(RA.isIndexed(new Date()));
+      assert.isFalse(RA.isIndexed(new Error()));
+      assert.isFalse(RA.isIndexed({ 0: 1, length: 1 }));
+      assert.isFalse(RA.isIndexed(1));
+    });
+  });
+});
```
char0n: We're missing test for currying in form on ramda placeholder. See how it should be done here: https://github.com/char0n/ramda-adjunct/blob/a75aa870faf3356e73c7de5fc3e7d5dbae1e071b/test/isFalse.js#L43

types/index.d.ts
```diff
@@ -1350,6 +1350,9 @@ declare namespace RamdaAdjunct {
          */
         sortByProps(props: string[], list: object[]): object[];
         sortByProps(props: string[]): (list: object[]) => object[];
+
+        /** Determine if input value is an indexed data type */
+        isIndexed(val: any): boolean;
```
char0n: I think typescript allows following syntax
```suggestion
        isIndexed(val: any): val is string | any[];
```

damiangreen:
thx for your feedback, i'll take a look later

src/isIndexed.js
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @func isIndexed
+ * @memberOf RA
+ * @category Type
```
char0n:
```suggestion
 * @since {@link https://char0n.github.io/ramda-adjunct/2.26.0|v2.26.0}
 * @category Type
```

test/isIndexed.js
damiangreen: Something like this?
```
    const isIndexed = RA.isIndexed(R.__);

    assert.isTrue(isIndexed([1, 2]));
  });
```
This doesn't work for me so i'll need to research this more

char0n: Take a look at this for example: https://github.com/char0n/ramda-adjunct/blob/86b7e1ca88b5f170ad977d930dc907a61734ad68/src/isNegative.js#L26

Your `inIndexed` at this moment is not curried. We have to wrap it into `curry1`. More info about ramda currying here: https://fr.umio.us/favoring-curry/

damiangreen: ah thanks!

damiangreen: I, my other PR i curried it, using `curry`, not `curry1`. I'm not sure  if there's any difference? e.g. `paths.js` uses `curry`

char0n: Excelusively for unary functions (functions with one argument) we historically used ramdas internal curry1 function which is faster then public `curry` or `curryN` functions. But it created some additional problems.

Check out this [issue](https://github.com/char0n/ramda-adjunct/issues/1340) for further explanation.

types/index.d.ts
```diff
@@ -1350,6 +1350,9 @@ declare namespace RamdaAdjunct {
          */
         sortByProps(props: string[], list: object[]): object[];
         sortByProps(props: string[]): (list: object[]) => object[];
+
+        /** Determine if input value is an indexed data type */
```
```suggestion
        /**
         * Determine if input value is an indexed data type.
         */
```

damiangreen:
since this is #1337 does that make me eLeet :)

Issue #1237, opened 2019-12-24
skipTake


char0n:
**Is your feature request related to a problem? Please describe.**

When given a number n and an array, returns an array containing every nth element

`R.addIndex(R.takeWhile)` with specialized predicate can be used to implement this function.

**Describe the solution you'd like**

```js
skipTake(2, [1,2,3,4]);
//=> [1, 3]

skipTake(3, _.range(20));
//=> [0, 3, 6, 9, 12, 15, 18]
```

**Describe alternatives you've considered**

I've considered [R.takeWhile](https://ramdajs.com/docs/#takeWhile), but doesn't exactly do what we want without specialized composition.

**Additional context**

damiangreen:
I've got a working implementation of this. I can create a PR once I've created tests

PR #1338, opened 2020-03-10
feat: skipTake

damiangreen:
Add `skipTake` function. Solves https://github.com/char0n/ramda-adjunct/issues/1237


char0n:
By a quick peak, I can tell we're missing tests for typings.

damiangreen:
yep, i had a crack at those, but  the types didnt work, i think because of the underlying type system in ramda

types/test/skipTake.ts
```diff
@@ -0,0 +1,4 @@
+import * as RA from 'ramda-adjunct';
+
+RA.skipTake(2, [1, 2, 3, 4]); // $ExpectType number[]
+RA.skipTake<number>(2)([1, 2, 3, 4]); // $ExpectType number[]
```
char0n: Let's add some error expectations. Example here: https://github.com/char0n/ramda-adjunct/blob/4d0e49a067439be21ddb525cc8a0d5090395421a/types/test/fnull.ts#L9-L8

damiangreen:
all changes made, let us know what you think

types/index.d.ts
```diff
@@ -2,1360 +2,1472 @@ declare var RA: RamdaAdjunct.Static;
 // TypeScript Version: 2.4
 
 declare namespace RamdaAdjunct {
```
char0n: This whole file have been changed. We only want changes regarding `skipTake`.

test/skipTake.js
```diff
@@ -0,0 +1,41 @@
...
+      assert.deepEqual(RA.skipTake(0, []), []);
+    });
+  });
+
+  it('should be curried', function() {
+    assert.deepEqual(RA.skipTake(2)([1, 2, 3, 4]), [1, 3]);
+    assert.deepEqual(RA.skipTake(2, [1, 2, 3, 4]), [1, 3]);
+  });
+
+  it('should support placeholder to specify "gaps"', function() {
```
char0n: We can drop this last test. `should be curried` is already testing support for placeholder implicitly.

test/skipTake.js
```diff
@@ -0,0 +1,41 @@
...
+      assert.deepEqual(RA.skipTake(3, []), []);
+    });
+  });
+
+  context('given negative count as input', function() {
+    specify('should return an empty array', function() {
+      assert.deepEqual(RA.skipTake(-1, []), []);
+    });
+  });
+
```
char0n: I'm still missing test for negative skip and non empty array. Not sure how the function behave in that scenario according to the tests.

damiangreen: negative skip? I'm not sure i follow that one

damiangreen: the first test test for a non empty array?

test/skipTake.js
damiangreen: pushed.

types/index.d.ts
damiangreen: fixed

test/skipTake.js
char0n: I'm talking about this usecase. I'm not sure what will be the result of this.
```js
RA.skipTake(-2, [1, 2, 3, 4])
```

damiangreen: very good point.. and right now it returns [1,3].
What do you believe should be the correct behaviour here? throw an exception or return an empty array?

char0n: I believe what we see is expected behavior. I'd just create a tests which shows that it doesn't really matter if the number is positive or negative integer.

Another thing that comes to my mind is how does the function reacts on providing `2.3` number instead of `2`?

char0n: Tests tests tests...they gives all the answers ;]

PR #1344, opened 2020-03-13

js636f:
Batch 23

Files:

test/
- replaceAll.js
- resolveP.js
- seq.js
- sliceFrom.js
- sliceTo.js


js636f:
Thank you!

Issue #1361, opened 2020-03-18
isPlainObj


char0n:
**Is your feature request related to a problem? Please describe.**
`isPlainObj/isPlainObject` contains reference of ramdas internal `_isObject` function. We need to remove reference to this internal function and replace it with appropriate code.

**Describe the solution you'd like**

Internal `_isObject` can be replaced by 
```js
Object.prototype.toString.call(x) === '[object Object]';
```

**Describe alternatives you've considered**

--

**Additional context**

--

char0n:
@willblake01 possible candidate for you.

PR #1365, opened 2020-03-19
refactor(src): remove internal ramda reference

willblake01:
replace internal ramda _isObject

Ref #1361

# Pull request template

Please use the following [PR](https://github.com/char0n/ramda-adjunct/pull/905/files) as an example
of pull request. This PR deals with adding a new feature called `allSettledP`.

If you have any additional questions please don't hesitate to contact any of the core committers.

### DIFF
--- a/src/isPlainObj.js
+++ b/src/isPlainObj.js
@@ -1,10 +1,11 @@
-import _isObject from 'ramda/src/internal/_isObject';
 import { pipe, both, equals, toString, pathSatisfies, curryN } from 'ramda';
 
 import isNull from './isNull';
 import isObjLike from './isObjLike';
 import isFunction from './isFunction';
 
+const isObject = val =>
+  Object.prototype.toString.call(val) === '[object Object]';
 const isObjectConstructor = pipe(toString, equals(toString(Object)));
 const hasObjectConstructor = pathSatisfies(
   both(isFunction, isObjectConstructor),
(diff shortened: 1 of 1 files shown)


willblake01:
@char0n Can you please review, thanks!

src/isPlainObj.js
```diff
@@ -1,10 +1,11 @@
-import _isObject from 'ramda/src/internal/_isObject';
 import { pipe, both, equals, toString, pathSatisfies, curryN } from 'ramda';
 
 import isNull from './isNull';
 import isObjLike from './isObjLike';
 import isFunction from './isFunction';
 
+const isObject = val =>
+  Object.prototype.toString.call(val) === '[object Object]';
```
char0n:
```suggestion
```

src/isPlainObj.js
```diff
@@ -1,10 +1,11 @@
-import _isObject from 'ramda/src/internal/_isObject';
 import { pipe, both, equals, toString, pathSatisfies, curryN } from 'ramda';
 
 import isNull from './isNull';
 import isObjLike from './isObjLike';
 import isFunction from './isFunction';
 
+const isObject = val =>
```
char0n: IMHO using following composition should do the trick too. Just to be consistent with other changes in other PRs. This suggestion should be completely equivavelnt.
```suggestion
const isObject = pipe(type, identical('Object'));
```

willblake01: Sure thing, I like that. I think I did this one first or I probably would have gone with that the first time.

Issue #1253, opened 2020-01-04
isBigInt


char0n:
**Is your feature request related to a problem? Please describe.**

Predicate for determining if a value is a big int.

**Describe the solution you'd like**
```js
isBigInt(10n); //=> true
isBigInt(10); //=> false
```

**Describe alternatives you've considered**

--

**Additional context**

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

danieldeer:
Hi there, I'm going to work on it

PR #1406, opened 2020-04-15
isBigInt

FelipeParreira:
[#1253](https://github.com/char0n/ramda-adjunct/issues/1253)


char0n:
Thank you, I'll have a look at this PR during tommorow/friday.

FelipeParreira:
It seems there is an issue with the test runners that try to run tests on the browser. I have tried locally and it I have the same problem. It seems the problem is mostly with Safari and the headless version of Chrome.

There is another issue related to the big int syntax and how to initialize it properly. I tried to work around it with a polyfill library (which avoids the issue of the syntax/parsing error with the linter), but it will not initialize a big int de facto, just a wrapper, whose `value` property is still a `number`. 

Not sure how to solve these issues.

test/internal/polyfills/BigInt.js
```diff
@@ -0,0 +1,5 @@
+import BigInt from 'big-integer';
```
char0n, line 1: Let's create a `test/shared/BigInt.js` file instead and drop this polyfill usage. (remove also from package.json/lock)


The file content will look like this:

```js
export default typeof BigInt === 'function' ? BigInt : undefined;
```

**Explanation**: I'd rather we test on real native values or not test at all. We always avoided to testing our predicate function on polyfills. IMHO it doesn't really make sense to test on behavior that is not native.

test/isBigInt.js
```diff
@@ -0,0 +1,77 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import * as RA from '../src';
+import MAX_SAFE_INTEGER from '../src/internal/polyfills/Number.MAX_SAFE_INTEGER';
+import MIN_SAFE_INTEGER from '../src/internal/polyfills/Number.MIN_SAFE_INTEGER';
+import BigInt from './internal/polyfills/BigInt';
```
char0n, line 7:
```suggestion
import BigInt from './shared/BigInt';
```

test/isBigInt.js
```diff
@@ -0,0 +1,77 @@
...
+
+      assert.isFalse(RA.isBigInt(null));
+      assert.isFalse(RA.isBigInt(undefined));
+    });
+  });
+
+  it('should support placeholder to specify "gaps"', function () {
+    const isBigInt = RA.isBigInt(R.__);
+
+    assert.isTrue(isBigInt(BigInt(10)));
```
char0n, line 75:
```suggestion
    assert.isFalse(isBigInt(1));
```

src/isBigInt.js
```diff
@@ -0,0 +1,23 @@
+import { curryN } from 'ramda';
+
+/**
+ * Checks if value is a Big Integer.
```
char0n, line 4:
```suggestion
 * Checks if value is a BigInt.
```

types/index.d.ts
```diff
@@ -325,6 +325,11 @@ declare namespace RamdaAdjunct {
          */
         isNotInteger(val: any): boolean;
 
+        /**
+         * Checks whether the passed value is a `bigint`.
```
char0n, line 329:
```suggestion
         * Checks if value is a BigInt.
```

types/index.d.ts
```diff
@@ -325,6 +325,11 @@ declare namespace RamdaAdjunct {
...
+         */
+        isBigInt(val: any): boolean;
```
char0n, line 331: We're missing tests for the typing. For more information how to create one please see e.g. https://github.com/char0n/ramda-adjunct/blob/bcb1f981fd46043fa1d1b3b95da00fac2d6f5b10/types/test/isSafeInteger.ts#L8

test/isBigInt.js
```diff
@@ -0,0 +1,77 @@
...
+
+import * as RA from '../src';
+import MAX_SAFE_INTEGER from '../src/internal/polyfills/Number.MAX_SAFE_INTEGER';
+import MIN_SAFE_INTEGER from '../src/internal/polyfills/Number.MIN_SAFE_INTEGER';
+import BigInt from './internal/polyfills/BigInt';
+import args from './shared/arguments';
+import Symbol from './shared/Symbol';
+
+describe('isBigInt', function () {
+  context('given a big integer value', function () {
```
char0n, line 12: And here comes the trick. We only run test if native `BigInt` is available. We only need to `wrap` first two `context`s. The third `it` can always run because we're asserting for `false`.
```js
if (BigInt !== undefined) {
  context('given a big integer value', function () {
    specify('should return true', function () {
     ...assertions...
    });
  });

  context('given a non-big-integer value', function () {
    specify('should return false', function () {
      ...assertions...
    });
  });
}

**Comment by char0n** [comment]
> It seems there is an issue with the test runners that try to run tests on the browser. I have tried locally and I have the same problem. It seems the problem is mostly with Safari and the headless version of Chrome.

> There is another issue related to the big int syntax and how to initialize it properly. I tried to work around it with a polyfill library (which avoids the issue of the syntax/parsing error with the linter), but it will not initialize a big int de facto (for earlier versions of Node), just a wrapper, whose value property is still a number.

> Not sure how to solve these issues.

I suspect the problems will disappear if you implement the changes I propose.

File: src/isBigInt.js [hunk 8]
```diff
@@ -0,0 +1,23 @@
...
+ * @example
+ *
+ * RA.isBigInt(5); // => false
+ * RA.isBigInt(Number.MAX_VALUE); // => false
+ * RA.isBigInt(-Infinity); // => false
+ * RA.isBigInt(10); // => false
+ * RA.isBigInt(10n); // => true
+ * RA.isBigInt(BitInt(9007199254740991)); // => true
+ */
+const isBigInt = curryN(1, (val) => typeof val === 'bigint');
```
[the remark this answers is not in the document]
FelipeParreira: I considered this when implementing, but the doc for `type` says it does not further differentiate between the different types of objects.

```
/**
 * Gives a single-word string description of the (native) type of a value, returning such answers as 'Object',
 * 'Number', 'Array', or 'Null'. Does not attempt to distinguish user Object types any further, reporting them
 * all as 'Object'.
 */
```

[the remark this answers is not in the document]
FelipeParreira: Oh, I think I got it. It works for native JS objects but not for objects defined by someone consuming the library, is that?

test/internal/polyfills/BigInt.js
FelipeParreira: Ok, I was trying to work around the fact the library was being tested on earlier versions of node. But I see your point.

src/isBigInt.js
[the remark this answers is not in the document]
char0n: > Oh, I think I got it. It works for native JS objects but not for objects defined by someone consuming the library, is that?

It explicitly states that "Does not attempt to distinguish **user** Object". The key for understanding is **user** here. Everything **native** will be reported as expected, and everything user defined as `Object`. Of course, in new ES6 there is [workaround](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) even for that.

test/isBigInt.js
```diff
@@ -0,0 +1,79 @@
...
+        assert.isTrue(RA.isBigInt(BigInt('0x1fffffffffffff')));
+        assert.isTrue(
+          RA.isBigInt(
+            BigInt('0b11111111111111111111111111111111111111111111111111111')
+          )
+        );
+      });
+    });
+
+    context('given a non-big-integer value', function () {
```
char0n, line 34: Put this function outside of the `if` statement. This block of code is not using `BigInt` symbol anywhere so it's safe to execute always.

test/isBigInt.js
```diff
@@ -0,0 +1,79 @@
...
+        if (Symbol !== 'undefined') {
+          assert.isFalse(RA.isBigInt(Symbol));
+        }
+
+        assert.isFalse(RA.isBigInt(null));
+        assert.isFalse(RA.isBigInt(undefined));
+      });
+    });
+
+    it('should support placeholder to specify "gaps"', function () {
```
char0n, line 73: Put this function outside of the `if` statement. This block of code is not using `BigInt` symbol anywhere so it's safe to execute always.

char0n:
Otherwise looks good. Two last CRCs to process and  we're good to go. Good job!

PR #1409, opened 2020-04-16
feat(type): add invokeArgs

js636f:
@char0n, I consider this as a draft PR. It seems to me everything works as expected, but I'm unsure about some things (for example, I don't know is it right to use "category Object", etc.). So, I'm waiting for your review and I'm ready to refine this PR.

### DIFF
--- a/src/invokeArgs.js
+++ b/src/invokeArgs.js
@@ -23,12 +23,12 @@ const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
   let fullPath = obj;
 
   try {
-    for (let i = 0; i < pathToMethod.length; i++) {
+    for (let i = 0; i < pathToMethod.length; i = i + 1) {
       fullPath = fullPath[pathToMethod[i]];
     }
 
     return fullPath.apply(obj, args);
-  } catch(e) {
+  } catch (e) {
     return undefined;
   }
 });
--- a/src/invokeArgs.js
+++ b/src/invokeArgs.js
@@ -1,7 +1,5 @@
-import { curryN } from 'ramda';
+import { curryN, path, apply } from 'ramda';
 
-import isNotArray from './isNotArray';
-import isNotObj from './isNotObj';
 import isNotFunction from './isNotFunction';
 
 /**
@@ -22,36 +20,11 @@ import isNotFunction from './isNotFunction';
  * RA.invokeArgs(['path', 'to', 'non-existent', 'method'], [-1], Math); //=> undefined
  */
 
-const checkArgs = (pathToMethod, args) => {
-  return isNotArray(pathToMethod) || isNotArray(args);
-};
-
 const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
-  if (checkArgs(pathToMethod, args)) return undefined;
+  const method = path(pathToMethod, obj);
+  if (isNotFunction(method)) return undefined;
 
-  /**
-   * If we need these test cases:
-   *  assert.strictEqual(RA.invokeArgs([], [1, 2], testFunc), 3);
-   *  assert.strictEqual(RA.invokeArgs(['methodWithoutArgs'], [], testObj), 1);
-   * then arrays 'pathToMethod' and 'args' can be empty, so we don't need isEmptyArray
-   */
-
-  let fullPath = obj;
-
-  for (let i = 0; i < pathToMethod.length; i += 1) {
-    if (isNotObj(fullPath)) return undefined;
-    fullPath = fullPath[pathToMethod[i]];
-  }
-
-  /**
-   * if we are here then we have a valid function/method to run
-   * We don't need rethrow, because if it throws here then this is a
-   * function/method implementation detaild which are outside of our control
-   */
-
-  if (isNotFunction(fullPath)) return undefined;
-
-  return fullPath.apply(obj, args);
+  return apply(method, args);
 });
 
 export default invokeArgs;
--- a/src/invokeArgs.js
+++ b/src/invokeArgs.js
@@ -2,7 +2,7 @@ import { curryN } from 'ramda';
 
 import isNotArray from './isNotArray';
 import isNotObj from './isNotObj';
-import isFunction from './isFunction';
+import isNotFunction from './isNotFunction';
 
 /**
  * Invokes the method at path of object with given arguments.
@@ -49,7 +49,9 @@ const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
    * function/method implementation detaild which are outside of our control
    */
 
-  if (isFunction(fullPath)) return fullPath.apply(obj, args);
+  if (isNotFunction(fullPath)) return undefined;
+
+  return fullPath.apply(obj, args);
 });
 
 export default invokeArgs;
--- a/src/invokeArgs.js
+++ b/src/invokeArgs.js
@@ -23,7 +23,7 @@ const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
   let fullPath = obj;
 
   try {
-    for (let i = 0; i < pathToMethod.length; i = i + 1) {
+    for (let i = 0; i < pathToMethod.length; i += 1) {
       fullPath = fullPath[pathToMethod[i]];
     }
 
--- a/src/invokeArgs.js
+++ b/src/invokeArgs.js
@@ -1,0 +1,36 @@
+import { curryN } from 'ramda';
+
+/**
+ * Invokes the method at path of object with given arguments.
+ * If method doesn't exists the function returns undefined.
+ *
+ * @func invokeArgs
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category Object
+ * @sig Array -> Array -> Object -> *
+ * @param {Array} pathToMethod Path to method
+ * @param {Array} args Method's arguments
+ * @param {Object} obj Object
+ * @return {*}
+ * @example
+ *
+ * RA.invokeArgs(['abs'], [-1], Math); //=> 1
+ * RA.invokeArgs(['nonexistentMethod'], [-1], Math); //=> undefined
+ */
+
+const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
+  let fullPath = obj;
+
+  try {
+    for (let i = 0; i < pathToMethod.length; i++) {
+      fullPath = fullPath[pathToMethod[i]];
+    }
+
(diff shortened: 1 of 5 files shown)


src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
+import { curryN } from 'ramda';
+
+/**
+ * Invokes the method at path of object with given arguments.
+ * If method doesn't exists the function returns undefined.
```
char0n, line 5: Let's drop this seconds sentence. It's an implementation detail, implied and expected.
```suggestion
```

src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @func invokeArgs
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
```
char0n, line 9:
```suggestion
 * @since {@link https://char0n.github.io/ramda-adjunct/2.27.0|v2.27.0}
```

src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
...
+ * @category Object
+ * @sig Array -> Array -> Object -> *
+ * @param {Array} pathToMethod Path to method
+ * @param {Array} args Method's arguments
+ * @param {Object} obj Object
```
char0n, line 14:
```suggestion
 * @param {Object} obj The object to query
```

src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Invokes the method at path of object with given arguments.
+ * If method doesn't exists the function returns undefined.
+ *
+ * @func invokeArgs
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.7.0|v2.7.0}
+ * @category Object
+ * @sig Array -> Array -> Object -> *
+ * @param {Array} pathToMethod Path to method
```
char0n, line 12:
```suggestion
 * @param {Array.<string|number>} path The path of the method to invoke
```

src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Array} args Method's arguments
```
char0n, line 13:
```suggestion
 * @param {Array} args The arguments to invoke the method with
```

src/invokeArgs.js
```diff
@@ -0,0 +1,36 @@
...
+ * @param {Object} obj Object
+ * @return {*}
+ * @example
+ *
+ * RA.invokeArgs(['abs'], [-1], Math); //=> 1
+ * RA.invokeArgs(['nonexistentMethod'], [-1], Math); //=> undefined
```
char0n, line 19:
```suggestion
 * RA.invokeArgs(['path', 'to', 'non-existent', 'method'], [-1], Math); //=> undefined
```

types/test/invokeArgs.ts
```diff
@@ -0,0 +1,4 @@
+import * as RA from 'ramda-adjunct';
+
+RA.invokeArgs(['abs'], [-1], Math); // $ExpectType any
+RA.invokeArgs(['nonexistentMethod'], [-1], Math); // $ExpectType any
```
char0n, line 4:
```suggestion
RA.invokeArgs(['nonexistentMethod'], [-1], Math); // $ExpectType any
RA.invokeArgs([0, 'a'], [-1], Math); // $ExpectType any
RA.invokeArgs([{}], [-1], Math); // // $ExpectError
RA.invokeArgs([0], {}, Math); // // $ExpectError
```

types/index.d.ts
```diff
@@ -42,6 +42,11 @@ declare namespace RamdaAdjunct {
     type DictPred<T> = (value: T, key: string) => boolean;
 
     interface Static {
+        /**
```
char0n, line 45: Please put it to the end of definitions not the beginnig.

types/index.d.ts
```diff
@@ -42,6 +42,11 @@ declare namespace RamdaAdjunct {
...
+         * Invokes the method at path of object with given arguments.
+         */
+        invokeArgs(pathToMethod: string[], args: Array<string | number>, obj: object): any;
```
char0n, line 48: Missing curried form of typings. For more info about how to create one check https://github.com/char0n/ramda-adjunct/blob/54256f34da171b08e0a30b4c198aa697693d1662/types/index.d.ts#L1347

src/invokeArgs.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Array} args The arguments to invoke the method with
+ * @param {Object} obj The object to query
+ * @return {*}
+ * @example
+ *
+ * RA.invokeArgs(['abs'], [-1], Math); //=> 1
+ * RA.invokeArgs(['path', 'to', 'non-existent', 'method'], [-1], Math); //=> undefined
+ */
+
+const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
```
char0n, line 21: I was thinking about how to simplify this code yesterday. The pattern of simplification that came to my mind is following: 
1.) pluck method from a path
2.) pluck a parent context
3.) if method is a function we bind the parent context and apply the arguments (if not we return undefined)
4.) if the method call itself throws, we should rethrow

Pseudocode (just to demonstrate the idea)
```js
const invokeArgs = curryN(3, (mpath, args, obj) => {
  const method = path(mpath, obj);
  const context = path(init(mpath), obj);
  
  if (isNotFunction(method)) return undefined;
  if (isEmptyArray(mpath)) return undefined;
  
  const boundMethod = bind(method, context);
  return apply(boundMethod, args);
});
```

src/invokeArgs.js
```diff
@@ -0,0 +1,57 @@
...
+};
+
+const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
+  if (checkArgs(pathToMethod, args)) return undefined;
+
+  /**
+   * If we need these test cases:
+   *  assert.strictEqual(RA.invokeArgs([], [1, 2], testFunc), 3);
+   *  assert.strictEqual(RA.invokeArgs(['methodWithoutArgs'], [], testObj), 1);
+   * then arrays 'pathToMethod' and 'args' can be empty, so we don't need isEmptyArray
```
char0n, line 36: If `path` is an empty array then we should do nothing. We have to take care of this corner case because  when we use `R.path` for the implementation, if provided with empty array as a first arugment, return original value provided as second argument.

```js
const obj = {};
const method = R.path([], obj);
assert.true(method === obj)

File: src/invokeArgs.js [hunk 12]
```diff
@@ -0,0 +1,57 @@
...
+ * RA.invokeArgs(['abs'], [-1], Math); //=> 1
+ * RA.invokeArgs(['path', 'to', 'non-existent', 'method'], [-1], Math); //=> undefined
+ */
+
+const checkArgs = (pathToMethod, args) => {
+  return isNotArray(pathToMethod) || isNotArray(args);
+};
+
+const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
+  if (checkArgs(pathToMethod, args)) return undefined;
```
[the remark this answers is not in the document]
js636f: OK

src/invokeArgs.js
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * if we are here then we have a valid function/method to run
+   * We don't need rethrow, because if it throws here then this is a
+   * function/method implementation detaild which are outside of our control
+   */
+
+  if (isNotFunction(fullPath)) return undefined;
+
+  return fullPath.apply(obj, args);
```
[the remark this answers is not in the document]
js636f: OK, I used R.apply here

src/invokeArgs.js
```diff
@@ -0,0 +1,57 @@
...
+  if (checkArgs(pathToMethod, args)) return undefined;
+
+  /**
+   * If we need these test cases:
+   *  assert.strictEqual(RA.invokeArgs([], [1, 2], testFunc), 3);
+   *  assert.strictEqual(RA.invokeArgs(['methodWithoutArgs'], [], testObj), 1);
+   * then arrays 'pathToMethod' and 'args' can be empty, so we don't need isEmptyArray
+   */
+
+  let fullPath = obj;
```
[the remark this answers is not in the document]
js636f: OK, I  used R.path here

src/invokeArgs.js
js636f: OK, I get it.

src/invokeArgs.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * RA.invokeArgs(['abs'], [-1], Math); //=> 1
+ * RA.invokeArgs(['path', 'to', 'non-existent', 'method'], [-1], Math); //=> undefined
+ */
+
+const invokeArgs = curryN(3, (pathToMethod, args, obj) => {
+  const method = path(pathToMethod, obj);
+  if (isNotFunction(method)) return undefined;
+
+  return apply(method, args);
```
char0n, line 27: Your method is not bounded, as the psedocode suggested it should be. Imagine following case:

```js
const obj = {
   prop: 1,
   method(val) {
     return this.prop + this.prop + val;
   }
}

invokeArgs(['method'], 1, obj); // => 3
```
Current implementation returns `NaN`.

test/invokeArgs.js
```diff
@@ -0,0 +1,52 @@
...
+  const testFunc = (a, b) => a + b;
+
+  it('should invoke the method at path of object with given arguments', function () {
+    assert.strictEqual(RA.invokeArgs(['abs'], [-1], Math), 1);
+    assert.strictEqual(RA.invokeArgs(['method'], [1, 2, 3], testObj), 6);
+    assert.strictEqual(
+      RA.invokeArgs(['innerObj', 'method'], [1, 2, 3], testObj),
+      6
+    );
+    assert.strictEqual(RA.invokeArgs([], [1, 2], testFunc), 3);
```
char0n, line 30: This test should assert for `undefined` as expected value. As pseudocode suggest, it's a corner case that we have to take care explicitly to be compatible with `lodash`, `underscore` or any other invoke library.

We also talked about it [here](https://github.com/char0n/ramda-adjunct/pull/1409#discussion_r411964700).

src/invokeArgs.js
js636f: OK

test/invokeArgs.js
js636f: OK, I think I finally got this part

js636f:
@char0n, are we ready for merging?

js636f:
@char0n, I'm happy to help. Thank you very much for the excellent support!

PR #1414, opened 2020-04-20
fix(iterable): Convert isIterable into a type guard function

### LINKED ISSUE #1412: [Regression] Changes to isIterable breaks type safety
**Describe the bug**
[This change](https://github.com/char0n/ramda-adjunct/commit/e7b35e58ba597ef2ff98fa022fadd467b2c5ab09#diff-88bec6beae84369c0376b56b3bb88fe1L52) causes compilation errors

**To Reproduce**
```
export function foo<T>(value: T | Iterable<T>): T {
  if (isIterable(value)) {
    return value[Symbol.iterator]().next()?.value;
  }

  return value;
}
```

**Expected behavior**
Should compile.

**Actual behavior**
Fails to compile since isIterable is not a type guard function. Replacing `Iterable<T>` with `T[]` and `isIterable(value)` with `isArray(value)` would make this sample compile, though would defeat the purpose of using `isIterable`.

kawazoe:
Fixes #1412

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -50,7 +50,7 @@ declare namespace RamdaAdjunct {
         /**
          * Checks whether the passed value is iterable.
          */
-        isIterable(val: any): boolean;
+        isIterable<T>(val: any): val is Iterable<T>;
 
         /**
          * Checks if input value is an empty `Array`.
--- a/types/test/isIterable.ts
+++ b/types/test/isIterable.ts
@@ -1,0 +1,100 @@
+import * as RA from 'ramda-adjunct';
+
+class Impl implements Iterable<number> {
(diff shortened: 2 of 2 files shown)


kawazoe:
@char0n I'm not too sure why the firefox build is broken, but the types test seems to be passing on the chrome build and cannot really have any impact on different browsers.

char0n:
@kawazoe flaky infra of CricleCI for browser tests. I'll run them again, no worries.

Looks like you change is fine with `dstlint` for now.

kawazoe:
Actually, I noticed that there isn't any typing test for isIterable. I'll add some before we merge this PR.

char0n:
Thank you!

Issue #1238, opened 2019-12-24
trampoline


char0n:
**Is your feature request related to a problem? Please describe.**

All that trampoline does is repeatedly call the return value of a function until it’s no longer a function

**Describe the solution you'd like**

```js
function evenOline(n) {
  if (n === 0)
    return true;
  else
    return partial1(oddOline, Math.abs(n) - 1);
}

function oddOline(n) {
  if (n === 0)
    return false;
  else
    return partial1(evenOline, Math.abs(n) - 1);
}

trampoline(oddOline, 3);
//=> true
trampoline(evenOline, 200000);
//=> true

trampoline(oddOline, 300000);
//=> false

trampoline(evenOline, 200000000);
// wait a few seconds
//=> true
```

**Describe alternatives you've considered**
--

**Additional context**

The code and exmplanation comes from book Functional Javascript by Michael Fogus. 
Possible implementation for the function:

```js
function trampoline(fun /*, args */) {
  var result = fun.apply(fun, _.rest(arguments));

  while (_.isFunction(result)) {
    result = result();
  }

  return result;
}```

PR #1425, opened 2020-04-27
trampoline

FelipeParreira:
[#1238](https://github.com/char0n/ramda-adjunct/issues/1238)

### DIFF
--- a/src/trampoline.js
+++ b/src/trampoline.js
@@ -2,7 +2,7 @@ import isFunction from './isFunction';
 
 /**
  * A function that repeatedly calls the return value of a function,
- *  until it is no longer a function.
+ * until it is no longer a function.
  *
  * @func trampoline
  * @memberOf RA
--- a/src/trampoline.js
+++ b/src/trampoline.js
@@ -30,12 +30,12 @@ import isFunction from './isFunction';
  *  return R.partial(evenOline, [Math.abs(n) - 1]);
  * };
  *
- * RA.trampoline(oddOline, 3); //=> true
- * RA.trampoline(evenOline, 200000); //=> true
- * RA.trampoline(oddOline, 300000); //=> false
- * RA.trampoline(evenOline, 200000000); //=> true
+ * RA.trampoline(oddOline, [3]); //=> true
+ * RA.trampoline(evenOline, [200000]); //=> true
+ * RA.trampoline(oddOline, [300000]); //=> false
+ * RA.trampoline(evenOline, [200000000]); //=> true
  */
-const trampoline = curryN(2, (func, ...args) => {
+const trampoline = curryN(2, (func, args) => {
   let result = func(...args);
 
   while (isFunction(result)) {
--- a/src/trampoline.js
+++ b/src/trampoline.js
@@ -8,7 +8,7 @@ import isFunction from './isFunction';
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.28.0|v2.28.0}
  * @category Function
- * @sig ... -> Any
+ * @sig (*... -> a) -> [*] -> a
  * @param  {Function} fn
  * @param  {Array} args The arguments to call fn with
  * @return {*} The result that is not a function
--- a/src/trampoline.js
+++ b/src/trampoline.js
@@ -8,7 +8,7 @@ import isFunction from './isFunction';
  * @memberOf RA
  * @category Function
  * @sig ... -> Any
- * @param  {function} func
+ * @param  {Function} fn
  * @param  {} ...args
  * @return {*} Exclude<any, Function>
  * @example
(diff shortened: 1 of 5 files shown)


src/trampoline.js
```diff
@@ -0,0 +1,46 @@
+import isFunction from './isFunction';
+
+/**
+ * A function that repeatedly calls the return value of a function,
+ *  until it is no longer a function.
```
char0n, line 5:
```suggestion
 * until it is no longer a function.
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @func trampoline
+ * @memberOf RA
+ * @category Function
+ * @sig ... -> Any
+ * @param  {function} func
```
char0n, line 11:
```suggestion
 * @param  {Function} fn
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param  {} ...args
```
char0n, line 12:
```suggestion
 * @param  {Array} args The arguments to call fn with
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ * @return {*} Exclude<any, Function>
```
char0n, line 13:
```suggestion
 * @return {*} The result that is not a function
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
+import isFunction from './isFunction';
+
+/**
+ * A function that repeatedly calls the return value of a function,
+ *  until it is no longer a function.
+ *
+ * @func trampoline
+ * @memberOf RA
```
char0n, line 8:
```suggestion
 * @memberOf RA
 * @since {@link https://char0n.github.io/ramda-adjunct/2.28.0|v2.28.0}
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ * @category Function
+ * @sig ... -> Any
```
char0n, line 10:
```suggestion
 * @sig (*... -> a) -> [*] -> a
```

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param  {function} func
+ * @param  {} ...args
+ * @return {*} Exclude<any, Function>
+ * @example
+ *
+ * function evenOline(n) {
```
char0n, line 16: Can we make this code example compatible with our linting/formatting rules and prefer arrow functions instead of function declaration?

src/trampoline.js
```diff
@@ -0,0 +1,46 @@
...
+ *     return R.partial(evenOline, [Math.abs(n) - 1]);
+ * }
+ *
+ * RA.trampoline(oddOline, 3); //=> true
+ * RA.trampoline(evenOline, 200000); //=> true
+ * RA.trampoline(oddOline, 300000); //=> false
+ * RA.trampoline(evenOline, 200000000); //=> true
+ */
+
+const trampoline = (func, ...args) => {
```
char0n, line 36: Every function in `RA` must be curried. So we have to transform current signature into following:
```js
const trampoline = curry((fn, args) =>
 ...body...
);
```
Please compensate this change in all tests, typings and typing tests.

src/trampoline.js
```diff
@@ -14,34 +16,33 @@ import isFunction from './isFunction';
...
+ * };
  *
  * RA.trampoline(oddOline, 3); //=> true
  * RA.trampoline(evenOline, 200000); //=> true
  * RA.trampoline(oddOline, 300000); //=> false
  * RA.trampoline(evenOline, 200000000); //=> true
  */
-
-const trampoline = (func, ...args) => {
+const trampoline = curryN(2, (func, ...args) => {
```
char0n, line 38: This notation is never used in Ramda or Ramda Adjunct as it is confusing. Function signature should be total and arity should be ported always correctly. In order to achieve these properties the signature have to look like this:
```suggestion
const trampoline = curryN(2, (func, args) => {
```
Usage:
```js
RA.trampoline(Math.max, [1, 2, 3]); // => 3

File: test/trampoline.js [hunk 10]
```diff
@@ -49,4 +52,29 @@ describe('trampoline', function () {
...
+
+    assert.isTrue(isArray([]));
+    assert.isFalse(isArray(() => {}));
+
+    const max = RA.trampoline(Math.max);
+
+    assert.equal(max(1, 2, 3), 3);
+  });
+
+  it('should support placeholder to specify "gaps"', function () {
```
char0n, line 67: Please drop this test. We only use placeholder tests for function with arity = 1. For other function we exclusively use `should support currying tests`.

test/trampoline.js
```diff
@@ -49,4 +52,29 @@ describe('trampoline', function () {
       }
     );
   });
+
+  it('should support currying', function () {
+    const isArray = RA.trampoline(Array.isArray);
```
char0n, line 57: I'd suggest simplification:
```js
it('should support currying', function () {
  assert.strictEqual(RA.trampoline(Math.max, [1, 2, 3]), 3);
  assert.strictEqual(RA.trampoline(Math.max)([1, 2, 3]), 3);
});

char0n:
Closing this due to unresponsiveness of author.

Issue #1230, opened 2019-12-24
rangeStep


char0n:
**Is your feature request related to a problem? Please describe.**

```js
const rangeStep = R.curry((start, stop, step) => R.map(
  n => start + step * n,
  R.range(0, (1 + (stop - start) / step) >>> 0)
));
```

**Describe the solution you'd like**
```js
rangeStep(1, 4, 1);   // [1, 2, 3, 4]
rangeStep(2, 8, 2);   // [2, 4, 6, 8]
rangeStep(0, 10, 3);  // [0, 3, 6, 9]
rangeStep(3, -3, -2); // [3, 1, -1, -3]
```

**Additional context**

This function was described in [ramda cookbook](https://github.com/ramda/ramda/wiki/Cookbook#create-an-incrementing-or-decrementing-range-of-numbers-with-a-step)

- [ ] imlement the function
- [ ] add ref to ramda cookbook

PR #1447, opened 2020-05-14
rangeStep

FelipeParreira:
[#1230](https://github.com/char0n/ramda-adjunct/issues/1230)


FelipeParreira:
I used `Math.floor` (instead of the bitwise operator) in order to avoid throwing errors when the number of elements in the list would be negative. In this case, it simply returns an empty list/array (just like the `range` function from Python).

char0n:
This PR was superseded by https://github.com/char0n/ramda-adjunct/pull/1708

Issue #206, opened 2017-11-09
isError


char0n:
**Is your feature request related to a problem? Please describe.**

Should detect JavaScript errors.

**Describe the solution you'd like**

```js
const isError = R.pipe(R.type, R.identical("Error"));

isError(new Error()); //=> true
```

**Describe alternatives you've considered**

--

**Additional context**
We need to test on all [major error classes](https://airbrake.io/blog/nodejs-error-handling/nodejs-error-class-hierarchy)

GingerPlusPlus:
`R.is(Error)`.

What are "error-like" objects?

char0n:
Error like object is an object that conforms to the following: 

```js
(typeof errorLike.message == 'string' && typeof errorLike.name == 'string' && !isPlainObject(errorLike));
```

It's been decided [here](https://github.com/char0n/ramda-adjunct/pull/346) that we're not going to deal with the `another realm` problem and we're going to conform to `R` on this and use `R.is` to check for type for various types.

PR #1449, opened 2020-05-16
isError

FelipeParreira:
[#206](https://github.com/char0n/ramda-adjunct/issues/206)


char0n:
I've removed assertions for AssertionError and SystemError as these are node.js specific and we're running our tests in browser alongside the node.js env. As long as any error class is a subclass of base `Error` native class, the function will always return true.

Issue #788, opened 2018-12-31
toNumber


char0n:

mellero:
Will try to implement this today

PR #1454, opened 2020-05-19
toNumber

FelipeParreira:
[#788](https://github.com/char0n/ramda-adjunct/issues/788)

### DIFF
--- a/src/canCallNumberFunctionOn.js
+++ b/src/canCallNumberFunctionOn.js
@@ -1,0 +1,12 @@
+import { both, complement, hasIn, either, __ } from 'ramda';
+
+import neither from './neither';
+import isSymbol from './isSymbol';
+import isObj from './isObj';
+
+const canCallNumberFunctionOn = neither(
+  isSymbol,
+  both(isObj, complement(either(hasIn('toString', __), hasIn('valueOf', __))))
+);
+
+export default canCallNumberFunctionOn;
--- a/src/toNumber.js
+++ b/src/toNumber.js
@@ -1,0 +1,25 @@
+import { ifElse, partial, identity } from 'ramda';
+
+import canCallNumberFunctionOn from './canCallNumberFunctionOn';
+
+/**
+ * Converts value to a number.
+ *
+ * @func toNumber
+ * @category List
+ * @sig * -> Number
+ * @param {*} val The value to convert
+ * @return {Number}
+ * @example
+ * toNumber('1'); //=> 1
+ * toNumber('abc'); //=> NaN
+ * toNumber(Symbol.for('test')); //=> NaN
+ * toNumber(Object.create(null)); //=> NaN
+ */
+const toNumber = ifElse(
+  canCallNumberFunctionOn,
+  Number,
+  partial(identity, [NaN])
+);
+
+export default toNumber;
--- a/test/toNumber.js
+++ b/test/toNumber.js
@@ -1,0 +1,93 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import BigInt from './shared/BigInt';
+import * as RA from '../src';
+
+describe('toNumber', function () {
+  context(
+    'given a [non NaN] number or a number-convertible value',
+    function () {
+      specify('should return a [non NaN] number', function () {
+        let result = RA.toNumber(12);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(new Number(12));
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(12e3);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(Infinity);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(12.9);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(-12);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber('12');
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(null);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber([]);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber([1]);
+        assert.isNumber(result);
+        assert.isNotNaN(result);
+
+        result = RA.toNumber(true);
+        assert.isNumber(result);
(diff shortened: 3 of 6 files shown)


src/canCallNumberFunctionOn.js
```diff
@@ -0,0 +1,12 @@
+import { both, complement, hasIn, either, __ } from 'ramda';
```
char0n, line 1: Let's drop this file. Files of `src/` directory should only form public APIs.

Instead let's create `src/internal/isCoercible.js` file and export the `isCoercible` function from there.

This function may become part of RA public API in future if there will be more usecases. We can implement it as follows:

```js
const hasToString = both(isObj, hasIn('toString'));
const hasValueOf = both(isObj, hasIn('valueOf'));
const isNotSymbol = complement(isSymbol);

const isCoercible = both(isNotSymbol, either(hasToString, hasValueOf));

export default isCoercible;
```

src/toNumber.js
```diff
@@ -0,0 +1,25 @@
...
+ * @sig * -> Number
+ * @param {*} val The value to convert
+ * @return {Number}
+ * @example
+ * toNumber('1'); //=> 1
+ * toNumber('abc'); //=> NaN
+ * toNumber(Symbol.for('test')); //=> NaN
+ * toNumber(Object.create(null)); //=> NaN
+ */
+const toNumber = ifElse(
```
char0n, line 19:
```js
const toNumber = ifElse(isCoercible, Number, always(NaN));
```

src/toNumber.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Converts value to a number.
+ *
+ * @func toNumber
+ * @category List
+ * @sig * -> Number
+ * @param {*} val The value to convert
+ * @return {Number}
+ * @example
+ * toNumber('1'); //=> 1
```
char0n, line 14:
```suggestion
 
 * toNumber('1'); //=> 1
```

types/index.d.ts
```diff
@@ -1303,6 +1303,12 @@ declare namespace RamdaAdjunct {
          */
         toArray<T>(iterable: Iterable<T> | T): any[];
 
+        /**
+         * Converts value to a number.
+         * It returns `NaN` whenever the global `Number` function would return `NaN` or throw an error.
```
char0n, line 1308:
```suggestion
         * It returns `NaN` whenever the global `Number` function would return `NaN` or throw an error.
```
Let's drop this sentence. People should expect that this function is safe and never throw any error.

test/toNumber.js
```diff
@@ -0,0 +1,93 @@
...
+import * as R from 'ramda';
+
+import BigInt from './shared/BigInt';
+import * as RA from '../src';
+
+describe('toNumber', function () {
+  context(
+    'given a [non NaN] number or a number-convertible value',
+    function () {
+      specify('should return a [non NaN] number', function () {
```
char0n, line 11: I'm not sure I agree with this system of assertions.

What would you say to this system ?

Before
```js
        result = RA.toNumber(12e3);
        assert.isNumber(result);
        assert.isNotNaN(result);

        result = RA.toNumber(Infinity);
        assert.isNumber(result);
        assert.isNotNaN(result);

        result = RA.toNumber(null);
        assert.isNumber(result);
        assert.isNotNaN(result);

        result = RA.toNumber(undefined);
        assert.isNaN(result);
```
After
```js
        assert.strictEqual(RA.toNumber(12e3), 12e3);
        assert.strictEqual(RA.toNumber(Infinity), Infinity);        
        assert.strictEqual(RA.toNumber(null), 0);
        assert.isNaN(RA.toNumber(undefined));
```

It's better to check to real values and explicitly see them tests. From your assertions it's not that apparent what the resulting value should be.

test/toNumber.js
```diff
@@ -0,0 +1,93 @@
+import { assert } from 'chai';
+import * as R from 'ramda';
+
+import BigInt from './shared/BigInt';
+import * as RA from '../src';
+
+describe('toNumber', function () {
+  context(
+    'given a [non NaN] number or a number-convertible value',
```
char0n, line 9:
```suggestion
    'given a coercible value',
```

test/toNumber.js
char0n, line 11:
```suggestion
      specify('should return a number', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,93 @@
...
+        if (BigInt !== undefined) {
+          result = RA.toNumber(BigInt(1234));
+          assert.isNumber(result);
+          assert.isNotNaN(result);
+        }
+      });
+    }
+  );
+
+  context('given NaN or a non-number-convertible value', function () {
```
char0n, line 69:
```suggestion
  context('given non-coercible value', function () {
```

char0n: And this block should only contain following two assertions:
```js
      assert.isNaN(RA.toNumber(Object.create(null)));
      assert.isNaN(RA.toNumber(Symbol.for('')));
```

test/toNumber.js
```diff
@@ -0,0 +1,93 @@
...
+      assert.isNaN(RA.toNumber(new Set()));
+      assert.isNaN(RA.toNumber(new Map()));
+      assert.isNaN(RA.toNumber(/my regex/gi));
+    });
+  });
+
+  it('should support placeholder to specify "gaps"', function () {
+    const toNumber = RA.toNumber(R.__);
+
+    assert.isNumber(toNumber(1));
```
char0n, line 88: Replace these 3 assertions with simple
```js
assert.strictEqual(toNumber(1), 1);

**Review by char0n — CHANGES_REQUESTED** [review]
Hi @FelipeParreira,

Sorry for waiting for CR (code review) so long. Please have a look at my comments. Thank you for your contribution!

**Comment by char0n** [comment]
@FelipeParreira can you please rebase this PR on top of current master ?

File: src/toNumber.js [hunk 9]
```diff
@@ -0,0 +1,25 @@
+import { ifElse, partial, identity } from 'ramda';
+
+import canCallNumberFunctionOn from './canCallNumberFunctionOn';
+
+/**
+ * Converts value to a number.
+ *
+ * @func toNumber
+ * @category List
```
char0n, line 9:
```suggestion
 * @category Math
```

src/toNumber.js
```diff
@@ -0,0 +1,25 @@
...
+ * @sig * -> Number
```
char0n, line 10:
```suggestion
 * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
 * @sig * -> Number
```

char0n:
@FelipeParreira would you be able to finish this during next 3 days ? THanks for letting me know.

char0n:
Closing due to inactivity as it blocks other at this point.

PR #1467, opened 2020-05-27
refactor: better typescript support

rafaelss95:
Firstly, I want to say thanks for this amazing project.

I just started to use it in a new project that uses TS and as I found some unexpected behaviors in type definitions, I'm sending this PR in hope to fix them :). 

- I had to update the types to 3.5 due to the utility types I used (`Exclude`, `Extract`, `Omit`, etc.), but I guess it's fine since this version has landed a year ago. In any case, let me know if it's the right thing to do.
- While I noticed that there are many type definitions that doesn't have test files, I preferred to not create them in this PR or it'd be a pain to review. If you agree, I could send smaller PRs later.
- `tslint.json`'s changes are basically formatting changes made by `prettier`.
- This PR includes support for readonly array and _tries_ to infer the types more concisely.

<hr>

Let me know if this PR is too big to review so I can split this as you like.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,5 +1,5 @@
-declare var RA: RamdaAdjunct.Static;
-// TypeScript Version: 2.4
+declare const RA: RamdaAdjunct.Static;
+// TypeScript Version: 3.5
 
 declare namespace RamdaAdjunct {
     interface Functor<T> {
@@ -20,12 +20,12 @@ declare namespace RamdaAdjunct {
     }
 
     interface Catamorphism<T> {
-        cata<T1>(leftFn: (v: T1) => T, rightFn: (v: T1) => T): T;
+        cata<U>(leftFn: (v: U) => T, rightFn: (v: U) => T): T;
     }
 
     enum SettledPromiseStatus {
-        Fulfilled = "fulfilled",
-        Rejected = "rejected",
+        Fulfilled = 'fulfilled',
+        Rejected = 'rejected',
     }
 
     interface SettledPromise<T> {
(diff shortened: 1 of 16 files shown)


char0n:
Hi @rafaelss95,

Thank you for this PR. My TS knowledge is mediocre at the best, so any improvement to our typings is very welcome.


> I had to update the types to 3.5 due to the utility types I used (Exclude, Extract, Omit, etc.), but I guess it's fine since this version has landed a year ago. In any case, let me know if it's the right thing to do.

Shouldn't be a problem

> While I noticed that there are many type definitions that doesn't have test files, I preferred to not create them in this PR or it'd be a pain to review. If you agree, I could send smaller PRs later.

Yes, I've started this effort some time ago. Every new PR has to be accompanied with typings tests now, but we have to create tests for all the functions that have been created before this effort started. If you're up to the task, go ahead and send smaller PRs adding typing tests. 

> tslint.json's changes are basically formatting changes made by prettier.

Acknowledged

> This PR includes support for readonly array and tries to infer the types more concisely.

Looks good!

types/test/toArray.ts
```diff
@@ -1,7 +1,10 @@
...
-RA.toArray(null); // $ExpectType any[]
-RA.toArray(undefined); // $ExpectType any[]
-RA.toArray(1); // $ExpectType any[]
-RA.toArray({}); // $ExpectType any[]
+RA.toArray('1'); // $ExpectType string[]
+RA.toArray([1, 2, 3]); // $ExpectType number[]
+RA.toArray({ a: 1, b: 2 }); // $ExpectType number[]
+RA.toArray({}); // $ExpectType never[]
+
+RA.toArray(null); // $ExpectError
```
char0n, line 8: Any expecting error here? I'd say null is valid input, because the function returns empty array.

types/test/sortByProps.ts
```diff
@@ -15,12 +15,19 @@ const clara = {
...
 
-RA.sortByProps(['name'])(people); // $ExpectType object[]
-RA.sortByProps(['p'])(people); // $ExpectType object[]
-RA.sortByProps(['lastName', 'name'])(people); // $ExpectType object[]
-RA.sortByProps(['lastName', 'p', 'name'])(people); // $ExpectType object[]
+// tslint:disable:max-line-length
+RA.sortByProps(mutableProps)(immutablePeople); // $ExpectType ({ name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; })[]
+RA.sortByProps(['p'])(immutablePeople); // $ExpectType ({ name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; })[]
+RA.sortByProps(immutableProps)(immutablePeople); // $ExpectType ({ name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; } | { name: string; age: number; lastName: string; })[]
+RA.sortByProps(immutableProps)(mutablePeople); // $ExpectType { name: string; age: number; lastName: string; }[]
```
char0n, line 27: I don't get this at all, can you pls explain the changes?

types/index.d.ts
```diff
@@ -204,8 +228,8 @@ declare namespace RamdaAdjunct {
          * Checks if value is not object-like.
          * A value is object-like if it's not null and has a typeof result of "object".
          */
-        isNotObjLike(val: any): boolean;
-        isNotObjectLike(val: any): boolean; // alias
+        isNotObjLike<T>(val: T): Exclude<T, object>;
+        isNotObjectLike<T>(val: T): Exclude<T, object>; // alias
```
char0n, line 232: You removed explicit information what this symbol is aliasing. Any particular reason for that ?

types/test/toArray.ts
rafaelss95: Hmm, I'd say if we go for this direction, we should change the method signature, because [you can't pass `null` where you expect an `Iterable`](https://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXwxwEEYYoBPAHgBUA+ACiwxDICMIQAueEs8gGSwBrEDVrwAPvACSzNhzEBKbtQDaAXQDcAKG2FeFeqmQQIizUA).

types/test/sortByProps.ts
rafaelss95: Of course, here we go:

Currently, `sortByProps` accepts any `string` inside array parameter, so even if your array is composed by object with properties say x and y, you could pass `['z']` and returns an `object[]`, which is not well typed, because you can't get autocompletion with this even if your input is typed. Now this method strict it to receive only valid properties and return the same type as received. Hope this clarifies :)

types/index.d.ts
rafaelss95: Excuse me, I don't get the question. Could you clarify please?

char0n: **before:**
```js
// alias of isNotObjLike
```

**now:**
```js
// alias
```

rafaelss95: No, I didn't change this. [The comment is as is before](https://github.com/char0n/ramda-adjunct/blob/master/types/index.d.ts#L208).

char0n: No worries, never-mind my comment here. You did not change the comment.

types/test/toArray.ts
char0n: Right, let's not go in this direction. Disregard my comment.

types/test/sortByProps.ts
char0n: Yeah I get it.

types/test/toArray.ts
```diff
@@ -1,7 +1,10 @@
...
 
-RA.toArray([1, 2, 3]); // $ExpectType any[]
-RA.toArray(null); // $ExpectType any[]
-RA.toArray(undefined); // $ExpectType any[]
-RA.toArray(1); // $ExpectType any[]
-RA.toArray({}); // $ExpectType any[]
+RA.toArray('1'); // $ExpectType string[]
+RA.toArray([1, 2, 3]); // $ExpectType number[]
+RA.toArray({ a: 1, b: 2 }); // $ExpectType number[]
+RA.toArray({}); // $ExpectType never[]
```
char0n, line 6: In my understanding `never` is a type that we use for functions that never returns, for example functions that always throws. I also understand that `never` is subtype of every other type and is assignable to every other type. But still why we use `never[]` here eludes me ;]

rafaelss95: Sorry for the delayed response. Btw, good observation! The reason for this is that since there are no keys/values in object, TS can't infer which type the `array` is and therefore assumes `never`.

Note that it's the same behavior of [**`ramda#keys`**](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ramda/index.d.ts#L895-L896), which also returns `never` for empty objects.

<hr>

When I looked at this test the first time I was wondering what was the real benefit of having this... I mean what do we expect when passing an empty object? An empty array? So why not just write `[]` instead of `toArray({})`? Anyway, at that time I preferred to keep this and just change the `ExpectType` to conform the stronger typing, but **now** I see it could be easily removed, don't you agree?

<hr>

The playgrounds for testing:

[**With empty object**](https://www.typescriptlang.org/play/#code/C4TwDgpgBAaghgGwK4QPIDMA8AVAfFAXimwG0BrCEAe3WIF0BuAKCYBMIBjBOAJ2nSQA7DsACWVQVGBUAgjx5wQOKBAAewCINYBnKFQBGAK07BcAClEaF+hBABcxAJQP4yNFjwlGLFtLkKQMwBvAF9HIA)
[**With filled object**](https://www.typescriptlang.org/play/#code/C4TwDgpgBAaghgGwK4QPIDMA8AVAfFAXimwG0BrCEAe3WIF0BuAKCYBMIBjBOAJ2nSQA7DsACWVQVGBUAgjx5wQOKBAAewCINYBnKFQBGAK07BcAClEaF+hBABcxAJQP4yNFjwlGLHy2lyFEDMAbzgHAEYAGih9BwByACY4gF9HIA)

rafaelss95:
@char0n I've rebased to solve the conflicts. Please let me know if there's anything pending to merge this.

Maushundb:
Can we move forward with this? I'd love to add a few type fixes on top of this refactor in the future. @rafaelpadovezi @char0n

rafaelpadovezi:
Hey @Maushundb , it looks like you mentioned the wrong Rafael. @rafaelss95 is the correct one

Maushundb:
@rafaelss95 whoops sorry

rafaelss95:
Hey @Maushundb, unfortunately I don't think this PR will be merged anytime soon, as I can't see any interest from the maintainer 😞.

char0n:
Sorry for my forced absence. Let's restart effort in this PR. Please rebase this PR on top of current master, and we'll go from there.

Issue #1602, opened 2020-09-21
Type of compact does not match the constraint it applies


BenoitHiller:
**Describe the bug**
The `compact` function has the following type:

```typescript
compact<T>(list: T[]): Array<NonNullable<T>>;
```

However it also filters out other falsy values which isn't reflected by the type constraint, namely: `false`, `''`, `0`, and `NaN`.

The type should be:

```typescript
compact<T>(list: T[]): Array<Exclude<T, false | null | undefined | '' | 0>>;
```

**To Reproduce**

1. Write a compact statement over a list that contains boolean values
2. Assert that the output does not contain false
3. See a type error

Here is a reproduction of the issue: https://codesandbox.io/s/elastic-banzai-50l4r?file=/index.ts

**Expected behavior**
There shouldn't be a type error.

**Screenshots**

**Desktop:**
 - The issue exists on everything running the Typescript compiler that I can get my hands on.

BenoitHiller:
I've opened a PR as it is a small change.

PR #1603, opened 2020-09-21
Improve return type of compact

BenoitHiller:
#1602

I'm unsure whether this constitutes a breaking change, as the types do represent an external API that people can rely on so if users were depending on the types returning too much then they could have issues after. If that is the case nearly all type changes would be breaking though.


types/index.d.ts
```diff
@@ -1187,7 +1187,7 @@ declare namespace RamdaAdjunct {
          * Creates an array with all falsy values removed.
          * The values false, null, 0, "", undefined, and NaN are falsy.
          */
-        compact<T>(list: T[]): Array<NonNullable<T>>;
+        compact<T>(list: T[]): Array<Exclude<NonNullable<T>, false | '' | 0>>;
```
char0n, line 1190: Aren't we missing `NaN`. `NaN` is not part of `NonNullable`.

char0n (commented):
Thanks! We don't consider enhancement to our types as breaking changes, though It may appear for some consumers at so.

types/index.d.ts
BenoitHiller: Unfortunately I'm pretty sure that they type of `NaN` is just `number` and that there is no more specific way to match it.

Like right now in the typescript APIs it just has the following to define NaN:
```typescript
declare var NaN: number;
```

So it doesn't have its own type. There seems to be an open proposal to add it:
https://github.com/Microsoft/TypeScript/issues/28682

char0n: Cool thank for explanation.

char0n (approved):
Thanks!

PR #1612, opened 2020-10-01
Add renameKeyWith

### LINKED ISSUE #1228: renameKeyWith

Renames single key of an object.

```js
const obj = {A: 1};
RA.renameKeyWith(R.concat('a'), 'A', obj); //=> {aA: 1}
```

**Describe alternatives you've considered**

--

**Additional context**

This function should be implemented as specialization of [R.renameKeysWith](https://char0n.github.io/ramda-adjunct/2.23.0/RA.html#.renameKeysWith). 

Related to https://github.com/char0n/ramda-adjunct/issues/1227

- [ ] implement the function
- [ ] add ref to ramda cookbook

dbartholomae:
Closes #1228

Hi there! Some questions:

1. I'm not sure I got the signature in the documentation correct, could you please check it? Currently it is `(a -> b) -> c -> {a: *} -> {b: *}`
2. I did two versions, one is a bit simpler (the latest commit), but the other one (before the latest commit) has the advantage of being stable when changing to an existing key name. Which one do you prefer?
3. For typing I just went with `object` as it is used for `renameKeysWith`, but actually I would prefer to limit the input to a `Record` and then limit the key to be a key of said record. The more complicated version has the disadvantage that it relies on `mergeRight` which we would have to use from `ramda-adjunct` even though it is deprecated to keep compatibilty with older `ramda` versions that do not include `mergeRight` yet. What do you think?

PS: There is no mention that PhantomJS is needed to run tests locally. I could either add it as devDependency or to the documentation in a separate PR if you want.

### DIFF
--- a/src/renameKeyWith.js
+++ b/src/renameKeyWith.js
@@ -1,13 +1,13 @@
-import { curry, isNil, omit, pick } from 'ramda';
+import { curry, equals, when } from 'ramda';
 
 import renameKeysWith from './renameKeysWith';
-import mergeRight from './mergeRight';
 
 /**
  * Creates a new object with the own properties of the provided object, but the
  * key `key` renamed according to logic of renaming function.
  *
- * If the new key name already existed in the object, it will be overwritten.
+ * Keep in mind that in case the new key name already existed on the object,
+ * the behaviour is undefined and the result may vary between various JS engines!
  *
  * @func renameKeyWith
  * @memberOf RA
@@ -24,9 +24,7 @@ import mergeRight from './mergeRight';
  * RA.renameKeyWith(R.concat('a'), 'A', { A: 1 }) //=> { aA: 1 }
  */
 const renameKeyWith = curry((fn, key, obj) =>
-  isNil(obj)
-    ? {}
-    : mergeRight(renameKeysWith(fn, pick(key, obj)), omit(key, obj))
+  renameKeysWith(when(equals(key), fn), obj)
 );
 
 export default renameKeyWith;
--- a/src/renameKeyWith.js
+++ b/src/renameKeyWith.js
@@ -13,7 +13,7 @@ import renameKeysWith from './renameKeysWith';
  * @memberOf RA
  * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
  * @category Object
- * @sig (k -> k) -> k -> {k: v} -> {k: v
+ * @sig (k -> k) -> k -> {k: v} -> {k: v}
  * @param {Function} fn Function that renames the keys
  * @param {!string} key Key to rename
  * @param {!Object} obj Provided object
--- a/src/renameKeyWith.js
+++ b/src/renameKeyWith.js
@@ -1,0 +1,33 @@
+import { curry, isNil, omit, pick } from 'ramda';
+
+import renameKeysWith from './renameKeysWith';
+import mergeRight from './mergeRight';
+
+/**
+ * Creates a new object with the own properties of the provided object, but the
+ * key `key` renamed according to logic of renaming function.
+ *
+ * Keep in mind that in the case of keys conflict is behaviour undefined and
+ * the result may vary between various JS engines!
+ *
+ * @func renameKeyWith
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.5.0|v1.5.0}
+ * @category Object
+ * @sig (a -> b) -> c -> {a: *} -> {b: *}
+ * @param {Function} fn Function that renames the keys
+ * @param {!string} key Key to rename
+ * @param {!Object} obj Provided object
+ * @return {!Object} New object with renamed key
+ * @see {@link RA.renameKeysWith|renameKeysWith}
+ * @example
+ *
+ * RA.renameKeyWith(R.concat('a'), 'A', { A: 1 }) //=> { aA: 1 }
+ */
+const renameKeyWith = curry((fn, key, obj) =>
+  isNil(obj)
+    ? {}
+    : mergeRight(renameKeysWith(fn, pick(key, obj)), omit(key, obj))
+);
+
+export default renameKeyWith;
--- a/src/renameKeyWith.js
+++ b/src/renameKeyWith.js
@@ -11,9 +11,9 @@ import renameKeysWith from './renameKeysWith';
  *
  * @func renameKeyWith
  * @memberOf RA
- * @since {@link https://char0n.github.io/ramda-adjunct/1.5.0|v1.5.0}
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
  * @category Object
- * @sig (a -> b) -> c -> {a: *} -> {b: *}
+ * @sig (k -> k) -> k -> {k: v} -> {k: v
  * @param {Function} fn Function that renames the keys
  * @param {!string} key Key to rename
  * @param {!Object} obj Provided object
--- a/src/renameKeyWith.js
+++ b/src/renameKeyWith.js
@@ -7,8 +7,7 @@ import mergeRight from './mergeRight';
  * Creates a new object with the own properties of the provided object, but the
  * key `key` renamed according to logic of renaming function.
  *
- * Keep in mind that in the case of keys conflict is behaviour undefined and
- * the result may vary between various JS engines!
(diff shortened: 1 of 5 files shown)


dbartholomae:
What does "add ref to ramda cookbook" mean? Where should I put it? This I did not do yet.

types/test/renameKeyWith.ts
```diff
@@ -0,0 +1,6 @@
+import * as RA from 'ramda-adjunct';
+
+RA.renameKeyWith((str) => 'a' + str, 'A', { A: 1 }); // $ExpectType object
+RA.renameKeyWith((str) => 'a' + str)('A', { A: 1 }); // $ExpectType object
+RA.renameKeyWith((str) => 'a' + str)('A')({ A: 1 }); // $ExpectType object
+RA.renameKeyWith((str) => 'a' + str, 'A')({ A: 1 }); // $ExpectType object
```
char0n, line 6: Adding some `$ExpectError` tests would be nice

types/index.d.ts
```diff
@@ -2,1415 +2,1561 @@ declare var RA: RamdaAdjunct.Static;
...
+      pathToMethod: string[]
+    ): (args: Array<string | number>, obj: object) => any;
+
+    /**
+     * Converts double-precision 64-bit binary format IEEE 754 to signed 32 bit integer number.
+     */
+    toInteger32(n: number): number;
+    toInt32(n: number): number; // alias
+
+    /**
```
char0n, line 1554: This whole file has been changed. Probably some formatting happened. This file cannot be reviewed.

src/renameKeyWith.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Creates a new object with the own properties of the provided object, but the
+ * key `key` renamed according to logic of renaming function.
+ *
+ * Keep in mind that in case the new key name already existed on the object,
+ * the behaviour is undefined and the result may vary between various JS engines!
+ *
+ * @func renameKeyWith
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/1.5.0|v1.5.0}
```
char0n, line 14:
```suggestion
 * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
```

src/renameKeyWith.js
```diff
@@ -0,0 +1,30 @@
...
+ * @category Object
+ * @sig (a -> b) -> c -> {a: *} -> {b: *}
```
char0n, line 16: I think this is most appropriate
```suggestion
 * @sig (k -> k) -> k -> {k: v} -> {k: v}
```

test/renameKeyWith.js
```diff
@@ -0,0 +1,30 @@
...
+    });
+  });
+
+  it('should not change other key-value pairs', function () {
+    assert.deepEqual(RA.renameKeyWith(concat('a'), 'A', { A: 1, B: 2 }), {
+      aA: 1,
+      B: 2,
+    });
+  });
+
```
char0n, line 19: Let's add test for a case when we want to rename non-existing key

test/renameKeyWith.js
```diff
@@ -0,0 +1,30 @@
...
+  it('should return an empty object when renames keys on null and undefined', function () {
+    assert.deepEqual(RA.renameKeyWith(concat('a'), 'A', null), {});
+    assert.deepEqual(RA.renameKeyWith(concat('a'), 'A', undefined), {});
+  });
+
+  it('should be curried', function () {
+    assert.deepEqual(RA.renameKeyWith(concat('a'))('A')({ A: 1 }), {
```
char0n, line 26: Can we make this more explicit ?

Explicit different combination of calling signatures would be great. https://github.com/char0n/ramda-adjunct/blob/b4f2188b602fde9d0d9c221ce05280d46a4ac510/test/viewOr.js#L31

char0n (changes requested):
Nice work! Though we still need to make some changes before we can merge.

test/renameKeyWith.js
dbartholomae: Would it be fine to add all possible asserts to this test?

types/test/renameKeyWith.ts
dbartholomae: Will do

src/renameKeyWith.js
dbartholomae: Thanks, will do :)

char0n:
> I'm not sure I got the signature in the documentation correct, could you please check it? Currently it is (a -> b) -> c -> {a: *} -> {b: *}

I've provided a signature suggestion in a CR (code review)

> I did two versions, one is a bit simpler (the latest commit), but the other one (before the latest commit) has the advantage of being stable when changing to an existing key name. Which one do you prefer?

The current one as it's compatible with [renameKeysWith](https://char0n.github.io/ramda-adjunct/2.28.0/RA.html#.renameKeysWith) behavior.

> For typing I just went with object as it is used for renameKeysWith, but actually I would prefer to limit the input to a Record and then limit the key to be a key of said record. 

I'd go for a compatible typing with `renameKeysWith`. Then in separate PR I'd go for a more proper typing for both this functions as you suggest. 

> The more complicated version has the disadvantage that it relies on mergeRight which we would have to use from ramda-adjunct even though it is deprecated to keep compatibilty with older ramda versions that do not include mergeRight yet. What do you think?

This doesn't need comment as we prefer the simpler version.

char0n:
> What does "add ref to ramda cookbook" mean? Where should I put it? This I did not do yet.

There is a cookbook for ramda: https://github.com/ramda/ramda/wiki/Cookbook

When we implement a function that can be scoped to cookbook we add it there. 

Example: https://github.com/ramda/ramda/wiki/Cookbook#map-keys-of-an-object-rename-keys-by-a-function

This comment basically means to add `renameKeyWith` function into the cookbook in the same style as we add other ramda-adjunct function there.

test/renameKeyWith.js
char0n: Sure, check out the test for `viewOr` that I've attached. We want the same thing here.

types/index.d.ts
dbartholomae: Will commit a cleaned up version. You might want to also use prettier for the TypeScript files, though - I just saw that the repo is using prettier and only noticed now that not only is prettier not used for TypeScript, but there is even a difference in indentation settings.

src/renameKeyWith.js
dbartholomae: Will do

dbartholomae:
I've added it in the cookbook.

test/renameKeyWith.js
dbartholomae: Done

dbartholomae:
And made all changes

char0n:
> PS: There is no mention that PhantomJS is needed to run tests locally. I could either add it as devDependency or to the documentation in a separate PR if you want.

Yeah I'm aware of this and it's very unfortunate. CircleCI started to fail the browser tests from some point for FireFox. As I'm not able to ignore some browsers I did have to[ list them explicitly](https://github.com/char0n/ramda-adjunct/blob/56994024c1a4b20c9bbed207c6d39b12b50a3fcc/testem.json#L9). If we dn't list explicitly the testem framework will use any available browser (and that's what I want). Dunno how to get out of this except adding PhantomJS as dev dep.

src/renameKeyWith.js
```diff
@@ -0,0 +1,30 @@
...
+ * key `key` renamed according to logic of renaming function.
+ *
+ * Keep in mind that in case the new key name already existed on the object,
+ * the behaviour is undefined and the result may vary between various JS engines!
+ *
+ * @func renameKeyWith
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
+ * @category Object
+ * @sig (k -> k) -> k -> {k: v} -> {k: v
```
char0n, line 16:
```suggestion
 * @sig (k -> k) -> k -> {k: v} -> {k: v}
```

types/index.d.ts
char0n: Agree. In past the `eslint` did not support TypeScript. Now it does and we're fine just using `eslint` + `prettier` to get the job done. I'll create a follow up issue for this.

char0n (approved):
Nice job! We're good to go here.

Issue #978, opened 2019-07-13
Finish our typescript definition testing via dtslint


char0n:
Related PR where the effort was started here: https://github.com/char0n/ramda-adjunct/pull/977

dbartholomae:
I'll add a couple (and in doing so already found a typing error) soon.

Maushundb:
For reference, TS 3.9 has added `@ts-expect-error` [comments](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments) built into the library to test failed typed assertions. dtslint seem a bit more strict since you can assert the actual type, but worth noting.

PR #1613, opened 2020-10-01
Add missing type tests

dbartholomae:
Progress on #978

With this, all functions alphabetically from `allEqual` to `concatRight` are type tested. This also fixes wrong types for a curried version of `cata`


char0n (changes requested):
Thanks for contributing! Please check other typings tests. When writing typing tests we're usually interested also in `$ExpectError` expectations

dbartholomae:
Sure, I've added some expected errors :)

char0n (approved):
Looks good! Thanks.

PR #1615, opened 2020-10-04
whereAny

### LINKED ISSUE #889: whereAny

Takes a spec object and a test object; returns true if the test satisfies the spec. Each of the spec's own properties must be a predicate function. Each predicate is applied to the value of the corresponding property of the test object. where returns true if one of the predicates return true, false otherwise. This function is just like [R.where](https://ramdajs.com/docs/#where) but uses `boolean OR` instead or `boolean AND`

Possible implementation
```js
const whereAny = R.curry((spec, testObj) => {
  for (let [key, fn] of Object.entries(spec)) {
    if (fn(testObj[key])) {return true}
  }
  return false
})

const pred = searchValue => whereAny({
  a: R.includes(searchValue),
  b: R.equals(searchValue),
  x: R.gt(R.__, 3),
  y: R.lt(R.__, 3)
});
```

**Describe alternatives you've considered**

--

**Additional context**

Originally conceived on SO: https://stackoverflow.com/questions/55688667/ramda-whereany
`R.where` - https://ramdajs.com/docs/#where

kcouliba:
Closes #889

Hey there, I came up with that implementation for the whereAny function. But I have some doubts :

- I am not sure about the signature, maybe there is a guide to improve my understanding of it?

- I am running several assertions in some tests. I could break down each assertion into separated tests, but since it is testing the exact same behavior, I made it that way.

Thanks for the opportunity, I am open to any feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /.idea
+/.vscode
 /.tmp
 /.nyc_output
 /node_modules
--- a/src/index.js
+++ b/src/index.js
@@ -174,6 +174,7 @@ export { default as spreadProp } from './spreadProp';
 export { default as spreadPath } from './spreadPath';
 export { default as flattenProp } from './flattenProp';
 export { default as flattenPath } from './flattenPath';
+export { default as whereAny } from './whereAny';
 export { default as unzipObjWith } from './unzipObjWith';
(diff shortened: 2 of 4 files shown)


char0n:
Please see a closing comment in #889. Pick another issue to implement.

PR #1617, opened 2020-10-05
feat: adds isNumber function

mellero:
ref: #788 , based on #1454 and the work done by @FelipeParreira.


src/internal/isCoercible.js
```diff
@@ -0,0 +1,12 @@
+import { both, complement, hasIn, either } from 'ramda';
```
char0n, line 1: The logic in `isCoercible` is not correct. Here is a code fragment that behaves correctly. Please use this code for `isCoercible` and we'll go from here:

```js
'use strict';

const R = require('ramda');
const RA = require('./lib');

const hasToString = R.pathSatisfies(RA.isFunction, ['toString']);
const hasValueOf = R.pathSatisfies(RA.isFunction, ['valueOf']);
const isNotSymbol = R.complement(RA.isSymbol);

const isCoercible = R.allPass([isNotSymbol, hasToString, hasValueOf]);

console.assert(isCoercible(1));
console.assert(isCoercible('test'));
console.assert(isCoercible({}));
console.assert(isCoercible([]));
console.assert(isCoercible(Object.create(null)) === false);
console.assert(isCoercible(Symbol.for(1)) === false);
```

We explicitly asserting if there is a `toString()` or `valueOf()` method on a provided value. If the value is not an object, `pathSatisfies` automatically returns false.

char0n (changes requested):
Hi @mellero, thank you for your contribution!

Let's fix `isCoercible` and we'll go from there.

char0n:
Closing this due to unresponsiveness of author.

PR #1620, opened 2020-10-07
fix Identity missing in documentation

### LINKED ISSUE #1614: RA.Identity monad not showing in docs
**Describe the bug**

RA.Identity monad is not showing in our documentation

image

**To Reproduce**
1. Go to https://char0n.github.io/ramda-adjunct/2.28.0/
2. See there is not R.Identity

**Expected behavior**
RA.Identity should show in documentation. For example version this version of documentation (2.10.0) is still showing it: https://char0n.github.io/ramda-adjunct/2.10.0/

image

**Additional context**
We use [jsdoc](https://www.npmjs.com/package/jsdoc). We've updated jsdoc to v3.6.6 and maybe that is part of the problem. Needs to be investigated further.

wanjuntham:
Closes #1614

It is now working as expected.
image

As mentioned in [JSDoc `@private` page](https://jsdoc.app/tags-private.html),
> The `@private` tag marks a symbol as private, or not meant for general use. Private members are not shown in the generated output unless JSDoc is run with the `-p/--private` command-line option. In JSDoc 3.3.0 and later, you can also use the `-a/--access` command-line option to change this behavior.

### DIFF
--- a/src/fantasy-land/Identity.js
+++ b/src/fantasy-land/Identity.js
@@ -66,7 +66,6 @@ class Identity {
   /**
    * Private constructor. Use {@link RA.Identity.of|Identity.of} instead.
    *
-   * @private
    * @param {*} value
    * @return {RA.Identity}
    */


char0n (approved):
Cool! Thank you for the contribution

PR #1639, opened 2020-10-17
feat(877): matchall

ltruchot:
Contains:
- native curried matchAll 
- ponyfill curried matchAll
- tests
- typescript definition
- export

Decisions/Tradeoffs:
- native "matchAll" returns an iterable type: RegExpStringIterator. This is why most of the examples look like `[..." testtest ".matchAll (" te ")`, spreading the iterable into a regular Array. Creating this kind of iterator in polyfill involves symbols, generators, and other ES6 elements that should be filled by themselves, in a long and not perfect chain. So I decided to directly return the Array as a result, making some rarely used optimization of matchAll impossible.
- EcmaScript official documentation says that matchAll non global regex should throw an error (@see 21.1.3.12 in https://tc39.es/ecma262/#sec-string.prototype.matchall). Main browsers implements that behavior but node.js 12+ doesn't, so there is no test to find this error except for the ponyfill that enforce the correct error.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -212,6 +212,7 @@ export { default as toUinteger32 } from './toUinteger32';
 export { default as toUint32 } from './toUinteger32'; // alias of to toUinteger32
 // String
 export { default as replaceAll } from './replaceAll';
+export { default as matchAll } from './matchAll';
 export { default as escapeRegExp } from './escapeRegExp';
 export { default as trimStart } from './trimStart';
 export { default as trimLeft } from './trimStart'; // alias of trimStart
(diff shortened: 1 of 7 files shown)


ltruchot:
@char0n about the codeclimate error: not sure if we want to avoid this glue-code in the ponyfill since there is no "helper" or "shared" folder here, but I can do it if you prefer!

ltruchot:
@char0n Some news about this feature ?

char0n:
@ltruchot sorry this falled through my net. Are you willing to continue working on this PR?

ltruchot:
I was satisfied with my PR, but I'm pretty sure if I read it now I'll found improvment to do. Can you let me few day to check that ?

char0n:
> I was satisfied with my PR, but I'm pretty sure if I read it now I'll found improvment to do. Can you let me few day to check that ?

Sure, I'm sure it's a pristine work ;] I was more about if I do the code review (CR) and will have some suggestions or questions, if I should expect some answer, after this long time. But obviously I can ;] I'll get into CR.

Thank you for your work!

PR #1672, opened 2020-11-20
fix(type): type narrowing for isNotNil

Maushundb:
# Summary

This PR adds better typing for `isNotNil` so that it produces type narrowing when used in a conditional expression. Currently, the function decl just types `isNotNil` as a function that takes in anything, and returns a boolean. What it also does, but the current typing does not reflect,  is check if the type is not `null` or `undefined`, which, if done without RA like `a !== null && a!== undefined` would narrow the type to account for that. 

TS has special syntax for marking a function that produces type narrowing, which is what I used here. This will allow users of the lib to write code as follows without TS throwing an error:

```
// toy implementation
const isNotNilBefore = <T>(val: T | null | undefined) => {
  return val !== null && val !== undefined;
};
const isNotNil = <T>(val: T | null | undefined): val is T => {
  return val !== null && val !== undefined;
};

type Test = {
  a: number | undefined;
}
const myObj: Test = {a: 1}

if (isNotNilBefore(myObj.a)) {
  myObj.a + 1 // TS error! a is potentially undefined
}
if (isNotNil(myObj.a)) {
  myObj.a + 1 // Ok
}

```

See [this link](https://www.typescriptlang.org/play?target=7&jsx=0&ts=4.0.5#code/MYewdgzgLgBAlhAciKi4BsBCBTAZiAJ2xgF4YAeAFQD4AKANwEN0AuGSmAHxjAFd10XGLzAATPHDDZRASlLUYAbwBQMGESi8CYGE0EBCEmT4CYAMjO7mMQ2RHjck6QG5lAX1ehIsBMlQZSChoGZjYObhNBbnsJKVk2PXgIdnklVXVsTW0rAyMefkELHJs8mMc41w9lZSgATwAHYkpsaECVNUY2PgBbACNsAiEyp1FK5S9W7tqAeV6AKzCW2DJFTpgARjdquFwYWl8UNCw8QmxaKdm5gDpGGTl2mAv5m5gAag33ZR29g-90c5mz1u93ST2ujDeHy2QA) to run this example in the TS playground.

# Test Plan
Unit tests


types/test/isNotNil.ts
```diff
@@ -0,0 +1,22 @@
...
+  obj.a; // $ExpectType number
+}
+
+if (RA.isNotNil(obj.b)) {
+  obj.b; // $ExpectType number
+}
+
+if (RA.isNotNil(obj.c)) {
+  obj.c; // $ExpectType number
+}
```
char0n, line 22:
```suggestion
}

RA.isNotNil('string'); // $ExpectType boolean
RA.isNotNil(1); // $ExpectType boolean
RA.isNotNil([]); // $ExpectType boolean
RA.isNotNil({}); // $ExpectType boolean
```

char0n (approved):
Great. Thanks for contributing!

types/test/isNotNil.ts
```diff
@@ -0,0 +1,27 @@
+import * as RA from 'ramda-adjunct';
+
+interface Test {
+  a: number | undefined;
+  b: number | null;
+  /* tslint:disable-next-line:no-null-undefined-union null or undefined is the accurate type here */
+  c: number | null| undefined;
```
char0n, line 7:
```suggestion
  c: number | null | undefined;
```

types/test/isNotNil.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+if (RA.isNotNil(obj.c)) {
+  obj.c; // $ExpectType number
+}
+
+RA.isNotNil('string'); // $ExpectType boolean
+RA.isNotNil(1); // $ExpectType boolean
+RA.isNotNil([]); // $ExpectType boolean
+RA.isNotNil({}); // $ExpectType boolean
```
char0n, line 27:
```suggestion
RA.isNotNil({}); // $ExpectType boolean
RA.isNotNil(null); // $ExpectType boolean
RA.isNotNil(undefined); // $ExpectType boolean
```

Issue #1807, opened 2021-03-12
Add .mocharc.js and .husky to .npmignore


char0n:

gunwooko:
Hi! Can I work on this? This is my first time to contribute in open-source projects.

char0n:
@gunwooko sure, go ahead!

PR #1808, opened 2021-03-14
chore(.npmignore): Add .mocharc.js and .husky to .npmignore

gunwooko:


char0n (approved):
Thanks for contributing!

char0n:
Closes #1807

Issue #793, opened 2019-01-02
isSentinelValue


char0n:
**Is your feature request related to a problem? Please describe.**

-1 is commonly called a "sentinel value," which basically means a value that's given an arbitrary semantic meaning within the greater set of values of its same type (numbers). The C-language uses -1 sentinel values for many functions that return >= 0 values for "success" and -1 for "failure."

**Describe the solution you'd like**
Possible implementation
```js
const isSentinelValue = curry(val => ~val === 0);
```

Assertions
```js
isSentinelValue(-1); //=> true
isSentinelValue('-1'); // => true
isSentinelValue(new Number('-1')); // => false
isSentinelValue(2); //=> false
isSentinelValue('a'); //=> false
```

**Describe alternatives you've considered**

**Additional context**

[The curious case of ~](https://github.com/getify/You-Dont-Know-JS/blob/9959fc904d584bbf0b02cf41c192f74ff4238581/types-grammar/ch4.md#the-curious-case-of-the-)

skShekhar:
@char0n Can I take this up

char0n:
@skShekhar sure go ahead.

skShekhar:
@char0n Is `[-1]` a sentinel value?

char0n:
@skShekhar that's a good question.

Sentinel value can only be [signed 32 integer](https://char0n.github.io/ramda-adjunct/2.32.0/RA.html#.isInteger32). We cannot count in implicit JavaScript coersions.

So the correct implementation would be:

```js
const isSentinelValue = curry(val => isInteger32(val) && ~val === 0);
```

skShekhar:
@char0n Using `isInteger32` will fail for `new Number(-1)` which is one of the assertions in the description.

char0n:
@skShekhar right, I've updated to assertion to expect `false`. `Number(-1)` is technically an object and not signed 32 bit integer value.

skShekhar:
@char0n Okay. If we are not going to do any type coercions, that means `'-1'` is also out. The only thing left is `-1`. Do we even need a function just to check if the value is `-1`?

char0n:
> @char0n Okay. If we are not going to do any type coercions, that means '-1' is also out. 
Yes, it means exactly that.

> The only thing left is -1. Do we even need a function just to check if the value is -1?

Yes, we have a lot functions asserting on single value. The function becomes handy in more complex functional compositions like validation and others.

PR #1848, opened 2021-04-12
Add isSentinelValue

skShekhar:
This PR adds a predicate to determine if a value is a sentinel value. #793

### DIFF
--- a/src/isSentinelValue.js
+++ b/src/isSentinelValue.js
@@ -3,7 +3,7 @@ import { curryN } from 'ramda';
 import isInteger32 from './isInteger32';
 
 /**
- * Checks whether the passed value is a sentinel value.
+ * Checks whether the passed value is {@link https://github.com/getify/You-Dont-Know-JS/blob/9959fc904d584bbf0b02cf41c192f74ff4238581/types-grammar/ch4.md#the-curious-case-of-the-|a sentinel value}.
  *
  * @func isSentinelValue
  * @memberOf RA
@@ -12,7 +12,6 @@ import isInteger32 from './isInteger32';
  * @sig * -> Boolean
  * @param {*} val The value to test
  * @return {boolean}
- * @see {@link RA.isSentinelValue|isSentinelValue}
  * @example
  *
  * RA.isSentinelValue(-1); //=> true
@@ -21,6 +20,7 @@ import isInteger32 from './isInteger32';
  * RA.isSentinelValue(1); //=> false
  * RA.isSentinelValue([-1]); //=> false
  */
-const isSentinelValue = curryN(1, (val) => isInteger32(val) && val === -1);
+// eslint-disable-next-line no-bitwise
+const isSentinelValue = curryN(1, (val) => isInteger32(val) && ~val === 0);
 
 export default isSentinelValue;
--- a/src/isSentinelValue.js
+++ b/src/isSentinelValue.js
@@ -1,0 +1,26 @@
+import { curryN } from 'ramda';
+
+import isInteger32 from './isInteger32';
+
+/**
+ * Checks whether the passed value is a sentinel value.
+ *
+ * @func isSentinelValue
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.33.0|v2.33.0}
+ * @category Type
+ * @sig * -> Boolean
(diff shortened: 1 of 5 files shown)


skShekhar:
@char0n Can you review this PR

char0n:
@skShekhar thanks! I'll provide a review ASAP.

src/isSentinelValue.js
```diff
@@ -0,0 +1,26 @@
...
+ * Checks whether the passed value is a sentinel value.
+ *
+ * @func isSentinelValue
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.33.0|v2.33.0}
+ * @category Type
+ * @sig * -> Boolean
+ * @param {*} val The value to test
+ * @return {boolean}
+ * @see {@link RA.isSentinelValue|isSentinelValue}
```
char0n, line 15:
```suggestion
 * @see {@link RA.isSentinelValue|isSentinelValue}
```
No need to mention the reference to function to itself.

src/isSentinelValue.js
```diff
@@ -0,0 +1,26 @@
...
+ * @example
+ *
+ * RA.isSentinelValue(-1); //=> true
+ *
+ * RA.isSentinelValue('-1'); //=> false
+ * RA.isSentinelValue(1); //=> false
+ * RA.isSentinelValue([-1]); //=> false
+ */
+const isSentinelValue = curryN(1, (val) => isInteger32(val) && val === -1);
```
char0n, line 24: Can we change this to idiomatic sentinel value check? The `isInteger` predicate already guarantees that value is signed 32 bit integer before it moves to `~val === 0`. If eslint is giving you issues, use `eslint-disable-line no-bitwise`.

```suggestion
const isSentinelValue = curryN(1, (val) => isInteger32(val) && ~val === 0);
```

src/isSentinelValue.js
```diff
@@ -0,0 +1,26 @@
+import { curryN } from 'ramda';
+
+import isInteger32 from './isInteger32';
+
+/**
+ * Checks whether the passed value is a sentinel value.
```
char0n, line 6:
```suggestion
 * Checks whether the passed value is {@link https://github.com/getify/You-Dont-Know-JS/blob/9959fc904d584bbf0b02cf41c192f74ff4238581/types-grammar/ch4.md#the-curious-case-of-the-|a sentinel value}.
```

types/index.d.ts
```diff
@@ -455,6 +455,11 @@ declare namespace RamdaAdjunct {
          */
         isSparseArray(val: any): boolean;
 
+        /**
+         * Checks whether the passed value is a sentinel value.
```
char0n, line 459:
```suggestion
         * Checks whether the passed value is {@link https://github.com/getify/You-Dont-Know-JS/blob/9959fc904d584bbf0b02cf41c192f74ff4238581/types-grammar/ch4.md#the-curious-case-of-the-|a sentinel value}.

char0n (changes requested):
Great job! Can you please look into my CR suggestion before we move forward?

src/isSentinelValue.js
skShekhar: Is there any reason why you prefer `~val === 0` over `val === -1`. Isn't sentinel value directly defined as -1?

char0n: Technically both are equivalent. The reason I prefer former (`~val === 0`) is that this operation actually performs a complement of the value and then we assert if the complement equals `+0` or `-0` and `0` respectively. This actually describe what the algorithm for computing sentinel value actually is rather then just asserting for `-1`.

I guess that's my reasoning about it.

src/isSentinelValue.js
skShekhar: Removed the reference

src/isSentinelValue.js
skShekhar: Addressed

src/isSentinelValue.js
skShekhar: Addressed

types/index.d.ts
skShekhar: Addressed

char0n (approved):

PR #2103, opened 2021-11-26
feat(toNumber): implement function

matheushahnn:
Ref #788

Followed this conversation https://github.com/char0n/ramda-adjunct/pull/1454 to create this PR


src/internal/isCoercible.js
```diff
@@ -0,0 +1,10 @@
+import { both, complement, either, hasIn, __ } from 'ramda';
+
+import { isObj, isSymbol, neither } from '..';
+
+const isCoercible = neither(
+  isSymbol,
+  both(isObj, complement(either(hasIn('toString', __), hasIn('valueOf', __))))
```
nolleto, line 7: Here you don't need `__`.

```js
both(isObj, complement(either(hasIn('toString'), hasIn('valueOf'))))
```

Usually, you will need it when you need to fill one parameter in the middle of a ramda function.

```js
// RamdaJs example
const greet = R.replace('{name}', R.__, 'Hello, {name}!');
greet('Alice'); //=> 'Hello, Alice!'
```

You can call
```js
const greet = R.replace('{name}', 'Alice');
greet('Hello, {name}!'); //=> 'Hello, Alice!'
```

Because, since the last parameter was not given, the `greet` function will wait for it.
Back to the function `const greet = R.replace('{name}', R.__, 'Hello, {name}!');`

Here, the `greet` function it's like:
```js
const greet = name => R.replace('{name}', name, 'Hello, {name}!');
```

So, with `__` you can choose with the parameter you want to pass later 😄

test/toNumber.js
```diff
@@ -0,0 +1,24 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe('ToNumber', function () {
+  context('given a number', function () {
```
nolleto, line 6: Maybe, you could create a `context` for all JS types 👀 

[number, boolean, string, ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures)

src/internal/isCoercible.js
matheushahnn: Nice catch, thank you

test/toNumber.js
matheushahnn: Sounds good!
I added the tests for all JS types

nolleto (approved):

src/toNumber.js
```diff
@@ -0,0 +1,22 @@
...
+ * Converts value to a number.
+ *
+ * @func toNumber
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
+ * @param {*} val The value to convert
+ * @return {Number}
+ * @example
+ *
+ * toNumber('1'); //=> 1
```
char0n, line 15: I'm suggesting following examples here. The examples derives from [lodash.toNumber examples](https://lodash.com/docs/4.17.15#toNumber)

````js
RA.toNumber(3.2); //=> 3.2
RA.toNumber(Number.MIN_VALUE); //=> 5e-324
RA.toNumber(Infinity); //=> Infinity
RA.toNumber('3.2'); //=> 3.2
RA.toNumber(Symbol('3.2')); //=> NaN

File: test/toNumber.js [hunk 4]
```diff
@@ -0,0 +1,62 @@
...
+      assert.isNaN(RA.toNumber(undefined));
+    });
+  });
+  context('given a string', function () {
+    specify('should return NaN', function () {
+      assert.isNaN(RA.toNumber('abc'));
+    });
+  });
+  context('given a symbol', function () {
+    specify('should return a NaN', function () {
```
char0n, line 53:
```suggestion
    specify('should return NaN', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+      assert.isNaN(RA.toNumber(Symbol.for('test')));
+    });
+  });
+  context('given an object', function () {
+    specify('should return a NaN', function () {
```
char0n, line 58:
```suggestion
    specify('should return NaN', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
+import { assert } from 'chai';
+
+import BigInt from './shared/BigInt';
+import * as RA from '../src';
+
+describe('ToNumber', function () {
+  context('given a number', function () {
```
char0n, line 7: If there are multiple `context` blocks in the same level can we please insert single empty line between them?

```js
context('desc1', function () {
});
context('desc2', function() { 
});
```
vs 

```js
context('desc1', function () {
});

context('desc2', function() { 
});
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+    specify('should return 0', function () {
+      assert.strictEqual(RA.toNumber(null), 0);
+    });
+  });
+  context('given a undefined', function () {
+    specify('should return NaN', function () {
+      assert.isNaN(RA.toNumber(undefined));
+    });
+  });
+  context('given a string', function () {
```
char0n, line 47:
```suggestion
  context('given a string that cannot be parsed to a number', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+        assert.strictEqual(RA.toNumber(true), 1);
+      });
+    });
+  });
+  context('given null', function () {
+    specify('should return 0', function () {
+      assert.strictEqual(RA.toNumber(null), 0);
+    });
+  });
+  context('given a undefined', function () {
```
char0n, line 42:
```suggestion
  context('given undefined', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+  });
+  context('given a bigint', function () {
+    specify('should return the number', function () {
+      assert.strictEqual(
+        RA.toNumber(BigInt(9007199254740991)),
+        9007199254740991
+      );
+    });
+  });
+  context('given a boolean', function () {
```
char0n, line 25:
```suggestion
  context('given a boolean value', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+    specify('should return the number', function () {
+      assert.strictEqual(RA.toNumber('123'), 123);
+      assert.strictEqual(RA.toNumber(123), 123);
+      assert.strictEqual(RA.toNumber(12e3), 12e3);
+      assert.strictEqual(RA.toNumber(Infinity), Infinity);
+      assert.strictEqual(RA.toNumber(+Infinity), +Infinity);
+      assert.strictEqual(RA.toNumber(-Infinity), -Infinity);
+    });
+  });
+  context('given a bigint', function () {
```
char0n, line 17:
```suggestion
  context('given a bigint value', function () {
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+  context('given a symbol', function () {
+    specify('should return a NaN', function () {
+      assert.isNaN(RA.toNumber(Symbol.for('test')));
+    });
+  });
+  context('given an object', function () {
+    specify('should return a NaN', function () {
+      assert.isNaN(RA.toNumber(Object.create(null)));
+    });
+  });
```
char0n, line 61: We're missing currying test:

```js
  it('should support placeholder to specify "gaps"', function () {
    const toNumber = RA.toNumber(R.__);

    assert.stringEqual(toNumber(3.2), 3.2);
  });
```

Full context here: https://github.com/char0n/ramda-adjunct/blob/b95fa04adaed288259947dbac5f0225e0ae138f4/test/isArray.js#L24

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+  context('given a symbol', function () {
+    specify('should return a NaN', function () {
+      assert.isNaN(RA.toNumber(Symbol.for('test')));
+    });
+  });
+  context('given an object', function () {
+    specify('should return a NaN', function () {
+      assert.isNaN(RA.toNumber(Object.create(null)));
```
char0n, line 59: This test is testing special kind of object with no prototype. The description of the test should probably reflect that.

We're also missing tests regarding `toString` and `valueOf` which makes objects coercible and thus usable with toNumber. Here is the list of cases which are warranted here. What do you think?

```js
const obj0 = {};

const obj1 = {
  toString() {
    return 1;
  },
};

const obj2 = {
  valueOf() {
    return 2;
  },
};

console.dir(Number(obj0));
console.dir(Number(obj1));
console.dir(Number(obj2));
```

test/toNumber.js
```diff
@@ -0,0 +1,62 @@
...
+      assert.strictEqual(RA.toNumber('123'), 123);
+      assert.strictEqual(RA.toNumber(123), 123);
+      assert.strictEqual(RA.toNumber(12e3), 12e3);
+      assert.strictEqual(RA.toNumber(Infinity), Infinity);
+      assert.strictEqual(RA.toNumber(+Infinity), +Infinity);
+      assert.strictEqual(RA.toNumber(-Infinity), -Infinity);
+    });
+  });
+  context('given a bigint', function () {
+    specify('should return the number', function () {
```
char0n, line 18:
```suggestion
    specify('should return a number', function () {
```

types/index.d.ts
```diff
@@ -1538,6 +1538,12 @@ declare namespace RamdaAdjunct {
         toUinteger32(val: number): number;
         toUint32(val: number): number; // alias
 
+        /**
+         * Converts value to a number.
+         * It returns `NaN` whenever the global `Number` function would return `NaN`.
```
char0n, line 1543: No need for this additional sentence. This sentence only reveal additional property of our `toNumber` function. More import is IMHO the fact that the function is safe and never throw errors, apart from `Number` function from javascript.
```suggestion
```

char0n (changes requested):
Thanks for contributing and great job here! Before we can merge, please have a look at my code review comments.

src/internal/isCoercible.js
```diff
@@ -0,0 +1,10 @@
+import { both, complement, either, hasIn } from 'ramda';
+
+import { isObj, isSymbol, neither } from '..';
+
+const isCoercible = neither(
+  isSymbol,
+  both(isObj, complement(either(hasIn('toString'), hasIn('valueOf'))))
```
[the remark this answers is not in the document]
matheushahnn: I agree with you, I changed to be more readable

char0n (approved):
Thanks for all changes. I'll hooks into this PR to make final minor changes before the merge.

Issue #654, opened 2018-08-02
feat: presence or nullIfEmpty


srghma:
like presence function in ruby

```
import * as R from 'ramda'

function presence(x) {
  return R.isEmpty(x) ? null : x
}

export default presence

```

char0n:
yeah why not
```js
const presence = when(isEmpty, stubNull)
```

tauantcamargo:
may i can get this one?

char0n:
@tauantcamargo sure

tauantcamargo:
so a few days ago i created this pull request https://github.com/char0n/ramda-adjunct/pull/2123 😃 @char0n

char0n:
@tauantcamargo doing the code review just now.

tauantcamargo:
@char0n don't worry. :D

PR #1463, opened 2020-05-23
feat(toInteger32): add toInteger32

OliverRadini:
#783

PR #2123, opened 2021-12-17
feat(presence): implemented presence function

tauantcamargo:
Implemented `presence` function.

Issue: #654

### DIFF
--- a/src/presence.js
+++ b/src/presence.js
@@ -1,0 +1,35 @@
+import * as R from 'ramda';
+
+/**
+ * Check if value is empty or undefined then returns null, otherwise, returns itself.
+ *
+ * @func presence
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
+ * @category Object
+ * @sig k -> null | k
+ * @param {k} val The value to test
+ * @return {null | k}
+ * @example
+ *
+ * presence({ foo: 'foo' }) // { foo: 'foo' }
+ * presence({}) // null
+ * presence(false) // null
+ * presence(true) // true
+ * presence('') // null
+ * presence('foo') // foo
+ * presence([]) // null
+ * presence([1, 2, 3]) // [1, 2, 3]
+ * presence(undefined) // null
+ * presence(0) // 0
+ * presence(null) // null
+ */
+
+const presence = R.cond([
+  [R.equals(false), R.always(null)],
+  [R.isEmpty, R.always(null)],
+  [R.isNil, R.always(null)],
+  [R.T, R.identity],
+]);
+
+export default presence;
--- a/test/presence.js
+++ b/test/presence.js
@@ -1,0 +1,93 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe.only('presence', function () {
+  context('given an object', function () {
+    specify('should return the object', function () {
+      const foo = { foo: 'foo' };
+
+      assert.strictEqual(RA.presence(foo), foo);
+    });
+  });
+
+  context('given an empty object', function () {
+    specify('should return null', function () {
+      const foo = {};
+
+      assert.strictEqual(RA.presence(foo), null);
+    });
+  });
+
+  context('given a boolean with value false', function () {
+    specify('should return null', function () {
+      const foo = false;
+
+      assert.strictEqual(RA.presence(foo), null);
+    });
+  });
+
+  context('given a boolean with value true', function () {
+    specify('should return boolean true', function () {
+      const foo = true;
+
+      assert.strictEqual(RA.presence(foo), foo);
+    });
+  });
+
+  context('given an empty string', function () {
+    specify('should return null', function () {
+      const foo = '';
+
+      assert.strictEqual(RA.presence(foo), null);
+    });
+  });
+
+  context('given a string', function () {
+    specify('should return the correct string', function () {
+      const foo = 'foo';
+
+      assert.strictEqual(RA.presence(foo), foo);
+    });
+  });
+
+  context('given an empty array', function () {
+    specify('should return null', function () {
+      const foo = [];
+
+      assert.strictEqual(RA.presence(foo), null);
+    });
+  });
+
+  context('given an array with values', function () {
+    specify('should return the array', function () {
+      const foo = [1, 2, 3];
+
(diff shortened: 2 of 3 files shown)


src/presence.js
```diff
@@ -0,0 +1,35 @@
+import * as R from 'ramda';
+
+/**
+ * Check if value is empty or undefined then returns null, otherwise, returns itself.
+ *
+ * @func presence
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
+ * @category Object
```
char0n, line 9:
```suggestion
 * @category Logic
```

src/presence.js
```diff
@@ -0,0 +1,35 @@
...
+ * @sig k -> null | k
```
char0n, line 10: Can we name the type parameter `a` and change the order of the result? It's more conventional for this codebase

```suggestion
 * @sig a -> a | null
```

src/presence.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {k} val The value to test
```
char0n, line 11:
```suggestion
 * @param {*} val The value to test
```

src/presence.js
```diff
@@ -0,0 +1,35 @@
...
+ * @return {null | k}
```
char0n, line 12:
```suggestion
 * @return {*}
```

src/presence.js
```diff
@@ -0,0 +1,35 @@
...
+ * presence('') // null
+ * presence('foo') // foo
+ * presence([]) // null
+ * presence([1, 2, 3]) // [1, 2, 3]
+ * presence(undefined) // null
+ * presence(0) // 0
+ * presence(null) // null
+ */
+
+const presence = R.cond([
```
char0n, line 28: Right, the implementation seems almost exactly precise. There is one small thing that we missed here; any string containing only white-space characters should fail the `present` test.

```js
RA.presence(''); // => null
RA.presence('   '); // => null
RA.presence('\t\n'); // => null
```

This regexp should solve this case: `/^\s*$/g`

More information about expected behavior in this article: https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html

src/presence.js
```diff
@@ -0,0 +1,35 @@
...
+ * @func presence
+ * @memberOf RA
+ * @since {@link https://char0n.github.io/ramda-adjunct/2.29.0|v2.29.0}
+ * @category Object
+ * @sig k -> null | k
+ * @param {k} val The value to test
+ * @return {null | k}
+ * @example
+ *
+ * presence({ foo: 'foo' }) // { foo: 'foo' }
```
char0n, line 15:
```suggestion
 * RA.presence({ foo: 'foo' }) // => { foo: 'foo' }
```

Please prefix all the examples with `RA.` prefix. Please also notice that we have a convention of showing the expected result using `// => {val}` convention.

src/presence.js
char0n: Also we could add link to `R.isEmpty` below the `@sig` tag
```js
* @see {@link http://ramdajs.com/docs/#isEmpty|R.isEmpty}
```

test/presence.js
```diff
@@ -0,0 +1,93 @@
+import { assert } from 'chai';
+
+import * as RA from '../src';
+
+describe.only('presence', function () {
+  context('given an object', function () {
+    specify('should return the object', function () {
+      const foo = { foo: 'foo' };
```
char0n, line 8: Can we name the parameter `val`?
```suggestion
      const val = { foo: 'foo' };
```

char0n (changes requested):
Thanks for contributing,It's looking good so far!. The PR must usually come with 5 files modified. You can see an example in this [PR](https://github.com/char0n/ramda-adjunct/pull/1463/files). This PR is missing typings and typings tests. Also please have a look at my code review comments.

tauantcamargo:
@char0n i'll take a look to the requests, thank u by the way.

src/presence.js
tauantcamargo: so @char0n i appreciate your suggestion, what do u think if we create first a method  `isBlank` to ramda-adjunct , so in the logic we could use it and can be used to other functionalities, what do u think about it?

tauantcamargo: @char0n i created this PR if u guess is necessary exists an `isBlank` let me know => https://github.com/char0n/ramda-adjunct/pull/2141/files

char0n: > so @char0n i appreciate your suggestion, what do u think if we create first a method `isBlank` to ramda-adjunct , so in the logic we could use it and can be used to other functionalities, what do u think about it?

Why not. Usual Modus operandi is that you put the helper function like in this case `isBlank` into `src/internal` and after the original PR dealing with `presence` is finished, you're return back to `src/internal/isBlank` and expose it as additional feature.

char0n: > @char0n i created this PR if u guess is necessary exists an `isBlank` let me know => https://github.com/char0n/ramda-adjunct/pull/2141/files

Right, I'll get into reviewing the https://github.com/char0n/ramda-adjunct/pull/2141/files right away. This PR is blocked until isBlank will hit the master branch.

tauantcamargo: @char0n  appreciate that .. thanks

char0n: Added following commit related to `isBlank` - https://github.com/char0n/ramda-adjunct/commit/1810924bd6286af5a04e8966e80095dc87799007

I found out implementation of `isBlank` was not aligned with the  [the article](https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html) we used as the reference.  `isBlank` should also return `true` for `null` value.

char0n:
I'm not sure I understand changes that have been made to this PR during last 3 commits. `master` branch now has `isBlank` function which is part of the public API of ramda-adjunct. We just cannot move it to `internal` directory anymore.

Now we should also fully utilize `isBlank` implementation to it's full potential:

```js
const presence = when(isBlank, stubNull);
```

Notes to above implementation:
- present is opposite of blank
- when value is present return it, otherwise return null
- the above sentence can also be: when value is blank return null, othersiwe return the original value (and that's our implementation)

tauantcamargo:
@char0n  i got it .. i'll update it later ..

char0n:
Right, let me know if you need any assistence

PR #2157, opened 2022-01-17
fix(pickIndexes): use R.includes instead of R.contains

zhulduz:


char0n:
Superseded by https://github.com/char0n/ramda-adjunct/pull/2162

PR #3100, opened 2024-08-20
fix(types): fix types for ramda 0.29.0 breaking changes

novascreen:
Fixes the types for `propNotEq` and `pathNotEq` based on the breaking changes in ramda 0.29.0 which changed the order of the arguments.

https://github.com/char0n/ramda-adjunct/pull/2568 only updated the JSDoc comments ([here](https://github.com/char0n/ramda-adjunct/pull/2568/files#diff-afed5a813d3a23925b8ccb7dab2f762b56434c286a2ea796893435ee1c30428dR12-R15) and [here](https://github.com/char0n/ramda-adjunct/pull/2568/files#diff-216cf68f7a6e00d613c6a13a88322fcda412e39412c07910f275cb5aab19cffcR11-R13)), but not the types.


char0n (approved):
Thank you @novascreen!

